PR #37, opened 2011-11-30
Redis Resource Agent

coredump:
Provides a Resource Agent to manage master-slave sets of Redis servers. Depends on start-stop-daemon, unfortunately, since redis doesn't support dropping privileges by itself.


krig:
This request should be closed as it is superseded by the request in #212.

dmuhamedagic:
Closing.

PR #113, opened 2012-07-11
New RA: sshtunnel

aborrero:
This new resource agent adds functionality to automatically create ssh connections (tunnels)
between all configured pacemaker nodes and/or other ssh servers.
By using this RA, the admin could benefit from using transparent ssh, connection multiplexing and
other advantages of ssh prestablished connections.

I rewrite some code as fghass asked for.

The basics of this RA:
It get information about how many and the name of every node in the pacemaker cluster.
Then it creates a SSH connection to every node, except the local node executing the RA.

If you deploy this as a clone, you will get all nodes connected via SSH. If you previously configured your ssh clients and servers properly (key exchange, enable multiplexing, etc..) all tools and services running in the cluster that are based on SSH will benefit from this previously stablished connection (tunnel): those services maybe rsync, recurrent scp, remote commands, and others.
Also, it can connect to other external ssh servers (none of the cluster nodes) so it's easy to deploy vpn-like ssh tunnels and have it in HA.
Both options at the same time, or just one of two, (pacemaker nodes interconnect and external ssh servers connection) is possible as well.

There is no concrete problem to solve but (from my point of view) a nice, interesting and optional functionality.

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -82,6 +82,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_ServeRAID.7 \
                           ocf_heartbeat_SphinxSearchDaemon.7 \
                           ocf_heartbeat_Squid.7 \
+			  ocf_heartbeat_sstunnel.7 \
                           ocf_heartbeat_Stateful.7 \
                           ocf_heartbeat_SysInfo.7 \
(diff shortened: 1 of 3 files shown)


dmuhamedagic:
Any news here?

aborrero:
I'm no longer using this RA, you can safely drop the pull request.

regards.

PR #132, opened 2012-08-30
redis RA: bring up in slave mode

michael-dev:
This pull requests is based on the last redis pull request here and makes the RA adhere to the pacemaker spec, that multi-state resources need to be brought up in slave mode. This fixes a race-condition when pacemaker has not yet promoted a master and thus restarts all instances infinitely, as they come up as masters instead of slaves.

Further, it autodetects port and password required for redis-cli from the config file.


krig:
This request should be closed as it is superseded by the request in #212.

dmuhamedagic:
Closing.

PR #166, opened 2012-11-04
commits following field experience during the last few months

y-trudeau:
All my commits since 3.9.3.


dmuhamedagic:
@y-trudeau there will be a new release soon, the rc is planned for mid January. Do you see some fixes urgently needed for mysql?

PR #179, opened 2012-11-18
Asterisk monitoring via sipsak not working

gnuday:
Hi,

When the original resource agent script called sipsak, it was sending a sip packet with an invalid contact URI. As a result Asterisk was sending back a 404 error in reply to the SIP OPTIONS sent. I haven't been able to get sipsak to send valid contact details in this mode so instead have made it do a SIP REGISTER. The resource agent now accepts an additional parameter=monitor_password, this is the sip account password you register with. I've tested the resource agent successfully
against asterisk 1.6 and 1.8.18.

Cluster configuration example of use

primitive resASTERISK ocf:heartbeat:asterisk \
        params realtime="false" monitor_sipuri="sip:user1@192.168.1.4"  \ monitor_password="eo34d84nvhekr" \
        op monitor interval="60" \
        op start interval="0s" timeout="120s" \
        op stop interval="0s" timeout="120s" \
        meta migration-threshold="1"

Cheers,
John
email@johncahill.net

### DIFF
--- a/heartbeat/asterisk
+++ b/heartbeat/asterisk
@@ -5,7 +5,8 @@
 #
 # Description:  Manages an Asterisk PBX as an HA resource
 #
-# Authors:      Martin Gerhard Loschwitz
+# Authors:	John Cahill      
+#		Martin Gerhard Loschwitz
 #               Florian Haas
 #
 # Support:      linux-ha@lists.linux-ha.org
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
This doesn't automatically merge anymore, please rebase.

PR #212, opened 2013-03-07
add redis resource agent script

phemmer:
I tried some of the existing redis resource agent scripts in the pull requests here and ran into a plethora of issues so I ended up creating my own. This was written completely from scratch, not done by rewriting code from another resource agent, and as such the code is a lot simpler than the other redis scripts (less than half the size).

* It does not use `start-stop-daemon`, and thus is not dependent upon debian/ubuntu
* It properly sets the default master priority based on the `slave_priority` setting from the redis config.
* It preserves the currently running master when taking the resource from unmanaged to managed. This is important as I needed to convert a single redis server into a pacemaker master/slave cluster without shutting redis down.
* It properly starts each client up in slave mode, and wait for pacemaker to promote it to master.

### DIFF
--- a/heartbeat/redis
+++ b/heartbeat/redis
@@ -7,6 +7,8 @@
 : ${OCF_RESKEY_config:=/etc/redis/redis.conf}
 : ${OCF_RESKEY_user:=redis}
 : ${OCF_RESKEY_rundir:=/var/run/redis}
+: ${OCF_RESKEY_pidfile_name:=redis-server.pid}
+: ${OCF_RESKEY_socket_name:=redis.sock}
 : ${OCF_RESKEY_port:=6379}
 
 REDIS_SERVER="$OCF_RESKEY_bin"
@@ -14,8 +16,8 @@ REDIS_CLIENT="$OCF_RESKEY_client_bin"
 REDIS_CONFIG="$OCF_RESKEY_config"
 REDIS_USER="$OCF_RESKEY_user"
 REDIS_RUNDIR="$OCF_RESKEY_rundir"
-REDIS_PIDFILE="$OCF_RESKEY_rundir/redis-server.pid"
-REDIS_SOCKET="$OCF_RESKEY_rundir/redis.sock"
+REDIS_PIDFILE="$OCF_RESKEY_rundir/$OCF_RESKEY_pidfile_name"
+REDIS_SOCKET="$OCF_RESKEY_rundir/$OCF_RESKEY_socket_name"
 REDIS_REPLICATION_PORT="$OCF_RESKEY_port"
 
 function meta_data() {
@@ -73,6 +75,24 @@ Directory to store socket and pid file in
 </longdesc>
 <shortdesc lang="en">Redis var/run dir</shortdesc>
 <content type="string" default="${OCF_RESKEY_rundir}"/>
+</parameter>
+
+<parameter name="pidfile_name" unique="0" required="0">
+<longdesc lang="en">
+The filename to use for the pidfile. Will be created in the rundir. 
+Should only be a basename, not a full path.
+</longdesc>
+<shortdesc lang="en">Redis pidfile name</shortdesc>
+<content type="string" default="${OCF_RESKEY_pidfile_name}"/>
+</parameter>
+
+<parameter name="socket_name" unique="0" required="0">
+<longdesc lang="en">
+The filename to use for the socket. Will be crated in the rundir. 
+Should only be a basename, not a full path.
+</longdesc>
+<shortdesc lang="en">Redis socket name</shortdesc>
+<content type="string" default="${OCF_RESKEY_socket_name}"/>
 </parameter>
 
 <parameter name="port" unique="0" required="0">
@@ -184,7 +204,7 @@ function start() {
 
 	while true; do
 		# wait for redis to start
-		typeset -A info=
+		typeset -A info
 		while read line; do
 			[[ "$line" == "#"* ]] && continue
 			[[ "$line" != *":"* ]] && continue
@@ -226,7 +246,7 @@ function stop() {
 	pid="$(<"$REDIS_PIDFILE")"
 	kill -TERM "$pid"
 
-	for (( i=0; i<10; i++ )); do
+	while true; do
 		monitor
 		status=$?
 		if (( status == OCF_NOT_RUNNING )); then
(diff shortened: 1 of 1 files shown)


krig:
Given the amount of scrutiny and work that has gone into this resource agent, and testimonies that it works as intended, and after looking at the code and finding no issues with it, I would vote for merging this. Any further improvements are unlikely to happen as long as it is kept separately in a pull request, and since more people are likely to use it once merged, such issues would be found more quickly. @dmuhamedagic, do you agree?

dmuhamedagic:
> @dmuhamedagic, do you agree?

Yes, in the part that it will get more eyes when merged. However,
there are still some unsolved and potentially dangerous issues
which haven't been fixed (as the ad hoc 10 second stop timeout):
that's sure to come back with revenge sooner or later.

It would be great if various submitters would finally agree on
which submission to actually take for the RA basis. It's quite a
waste to have to review multiple agents which are supposed to do
the same thing. Apparently, every new contributor is ready
exercise their programming muscle and write a new agent.

At any rate, @krig, I'm very happy to let you deal with this from
now on :-)

dmuhamedagic:
Closed other pull requests, let's hope that this one will get merged :)

atorstling:
Hi! I've starting testing of this script and I've found a typo which triggers a rather nasty bug on older systems. On line 187 the hashtable is declared with an extra equals sign:  "typeset -A info=" instead of "typeset -A info". This gives the table an extra entry with the key "0", which in turn triggers a segfault. This bug (or a variant thereof) is mentioned here: https://bugs.launchpad.net/ubuntu/+source/bash/+bug/928900 and seems to be fixed in bash 4.1 patch level 9 according to http://www.rpmfind.net/linux/RPM/opensuse/11.4/x86_64/bash-4.1-20.25.1.x86_64.html (see the fourth entry in the changelog). Just remove that last equals sign and everything seems to work fine.

atorstling:
Another problem with this script is that the soon-to-be-master get configured with "SLAVEOF <own-hostname>" during startup, before anyone has been promoted. This works, but seemingly mostly by chance since Redis gets confused talking to itself and decides to abort:
```
[1905] 26 Jan 11:15:46.803 * SLAVE OF <own-hostname>:6379 enabled (user request)
[1905] 26 Jan 11:15:47.152 * Connecting to MASTER <owh-hostname>:6379
[1905] 26 Jan 11:15:47.160 * MASTER <-> SLAVE sync started
[1905] 26 Jan 11:15:47.161 * Non blocking connect for SYNC fired the event.
[1905] 26 Jan 11:15:47.161 * Master replied to PING, replication can continue...
[1905] 26 Jan 11:15:52.166 * (Non critical) Master does not understand REPLCONF listening-port: -Reading from master: Connection timed out
[1905] 26 Jan 11:15:52.166 * Partial resynchronization not possible (no cached master)
[1905] 26 Jan 11:15:57.171 # Unexpected reply to PSYNC from master: -Reading from master: Connection timed out
[1905] 26 Jan 11:15:57.171 * Retrying with SYNC...
[1905] 26 Jan 11:15:57.172 # Bad protocol from MASTER, the first byte is not '$' (we received '+OK'), are you sure the host and port are right?
```

 I have made a small modification which uses a placeholder hostname during this period of confusion. I haven't come up with a better way of handling this, since Redis doesn't seem to support being a free slave :). Anyway, the change is in demote():

```bash
        master_port="${REDIS_REPLICATION_PORT}"

        current_host=$(crm_node -n)
	if [ "$master_host" == "$current_host" ]
	then
 		ocf_log debug "slave which will soon become master, setting placeholder master"		
		master_host="no-such-master"
	fi
	redis_client slaveof "$master_host" "$master_port"					
```

Cheers!

atorstling:
Having looked at the 10 second delay I figured that it's actually not so bad, because after those 10 retries the script gives up and returns $OCF_ERR_GENERIC.  At least after one final attempt to check if Redis is really dead, which I don't really see the point of. The retry loop seems to be there in order to give the process some time to shut down. So the script is giving Redis 10 seconds to shut down and  issues an error if that doesn't succeed. So if your Redis is taking a long time to save on shutdown this can be a problem, but it won't be a silent error. Anyway, I believe that the stop function can be changed to:
```bash
function stop() {
	monitor
	status=$?

	if (( status == OCF_NOT_RUNNING )); then
		ocf_log info "stop: redis is already stopped"
		return $OCF_SUCCESS
	fi

	pid="$(<"$REDIS_PIDFILE")"
	kill -TERM "$pid"

	while true; do
		monitor
		status=$?
		if (( status == OCF_NOT_RUNNING )); then
			crm_master_reboot -D
			return $OCF_SUCCESS
		fi
		if (( status != OCF_RUNNING_MASTER )) && (( status != OCF_SUCCESS )) && (( status != OCF_ERR_GENERIC )); then # we allow OCF_ERR_GENERIC because monitor can generate an error if we probe redis in the middle of shutdown (the sucket won't be responding but the process is up)
			ocf_log err "stop: Unknown error while stopping"
			return $OCF_ERR_GENERIC
		fi
		sleep 1
	done
}
```

dmuhamedagic:
On Mon, Jan 26, 2015 at 06:02:10AM -0800, Alexander Torstling wrote:
> Having looked at the 10 second delay I figured that it's actually not so bad, because after those 10 retries the script gives up and returns $OCF_ERR_GENERIC.  
The reason I am insisting on this getting fixed is that a failed
stop operation leaves the resource in an unknown state which then
requires a fencing operation (node reboot) to get the cluster
back to a normal state. Hope that I don't need to clarify why
that could be inconvenient.

> Anyway, I believe that the stop function can be changed to:

Looks much better.

atorstling:
@dmuhamedagic: I would be glad if my changes helps getting this request merged. Can I contribute my modifications somehow? If this pull request gets merged as is I can open a new one with my changes, but I don't see how I could amend this one. Is that possible? Perhaps you can change it @phemmer? For the record I have done some testing with my changes, but not extensively, and I haven't run them in production yet.

phemmer:
I will try to review these changes in the next day or so. Shouldn't be hard, just need a few min, and quite busy lately.

atorstling:
@phemmer: I created a pull request with my modifications against your branch. I hope it worked as intended :)

phemmer:
Thanks for the fixes. They've been merged.

krig:
Since all of the known issues have been fixed, this looks good to me.

dmuhamedagic:
On Wed, Feb 11, 2015 at 06:49:26AM -0800, Kristoffer Gr�nlund wrote:
> Since all of the known issues have been fixed, this looks good to me.

OK. Thanks everybody for the coding and reviewing effort.

PR #220, opened 2013-04-05
Log level fix

rdassen:
Logging of the "mount" command line is done for informational/debugging purposes. It does not indicate an error condition and should thus not be logged at LOG_ERR level.


krig:
Could you squash these commits so the pull request doesn't include the merge commit?

PR #335, opened 2013-10-16
Check for compiler flags with -Werror

dirkmueller:
For example for -fstack-protector, it is accepted by gcc for
all targets, however those targets where it is actually not
supported, it will print a warning like

"-fstack-protector is not supported for this platform"

which raises an error when using -Werror later. So better
not add -fstack-protector then to the CC_EXTRA_FLAGS


dirkmueller:
not needed anymore

PR #399, opened 2014-03-28
Medium: mysql: Update mysql with more sensible defaults

krig:
The default values used by the mysql RA are outdated. This
patch updates the socket and binary options to more sensible
defaults.

### DIFF
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -3,17 +3,18 @@
 #######################################################################
 
 # Attempt to detect a default binary
-OCF_RESKEY_binary_default=$(which mysqld_safe 2> /dev/null)
+if [ -f /etc/redhat_release ]; then
+	OCF_RESKEY_binary_default="/usr/libexec/mysqld"
+else
+	OCF_RESKEY_binary_default="$(which mysqld 2> /dev/null)"
+fi
 if [ "$OCF_RESKEY_binary_default" = "" ]; then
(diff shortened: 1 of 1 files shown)


grooverdan:
I haven't seen a missing mysqld_safe before. Either way this patch doesn't change the behaviour in this respect.

The transition off mysqld_safe for rhel could be troubles some as mysqld_safe sets up:
* ulimits on files if open-file-limits (and core file size)
* syslog and error log redirection
* timezone
* numa interleave settings
* nice settings

Of these the the ulimits on files is important for databases with a large number of tables/connections - running out of the default file limit of 1K files is quite bad (gross understatement) for busy databases.

@krig adding --no-auto-restart or --nowatch (same outcome) to mysqld_safe will prevent it watching/restarting mysqld on its own which might be what you want if you need the RM to handle the restarts.

@dmuhamedagic  Detecting the socket can be done by parsing mysqld --verbose --help for socket the same way mysqld_safe does in parse_arguements

Removing defaulting to /usr/libexec/mysqld for rhel wouldn't seem safer because of the range of mysqld_safe options that can't require significant manipulation to be set elsewhere.

Adding --no-watch to the myqld_safe may seem  useful but you'll need to check that this exists on a range of mysql versions and alternates, mariadb, percona as the main ones.

krig:
@grooverdan Thank you for your comments! Your suggestions sound good, and I think this patch needs more development before merging. I will close this pull request for now.

grooverdan:
@dmuhamedagic also see my_print_defaults

PR #406, opened 2014-04-03
mysql: Pull out the correct variable when setting read-only mode

symonds:
MariaDB10 has 3 variables that include read_only. Pull out the correct variable instead of grep'ing for it.

### DIFF
--- a/heartbeat/mysql
+++ b/heartbeat/mysql
@@ -411,7 +411,7 @@ get_read_only() {
     local read_only_state
 
     read_only_state=`$MYSQL $MYSQL_OPTIONS_REPL \
-        -e "SHOW VARIABLES" | grep read_only | awk '{print $2}'`
+        -ss -e "SHOW VARIABLES LIKE 'read_only'" | awk '{print $2}'`
 
     if [ "$read_only_state" = "ON" ]; then
         return 0
--- a/heartbeat/mysql
+++ b/heartbeat/mysql
@@ -411,9 +411,9 @@ get_read_only() {
     local read_only_state
 
     read_only_state=`$MYSQL $MYSQL_OPTIONS_REPL \
-        -ss -e "SHOW VARIABLES LIKE 'read_only'" | awk '{print $2}'`
+        -ss -e "SHOW GLOBAL VARIABLES LIKE 'read_only'"
 
-    if [ "$read_only_state" = "ON" ]; then
+    if [ $read_only_state =~ '*ON' ]; then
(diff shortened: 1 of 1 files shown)


grooverdan:
for correctness it should also be SHOW GLOBAL VARIABLES LIKE 'read_only'

symonds:
@grooverdan Merged in the 2 suggestions, Thanks.

krig:
Currently, this PR is broken. I don't know if it is because of changes in the mysql agent, either way it would need to be fixed to be merged. @symonds, still interested in submitting this?

heartbeat/mysql
```diff
@@ -411,9 +411,9 @@ get_read_only() {
     local read_only_state
 
     read_only_state=`$MYSQL $MYSQL_OPTIONS_REPL \
-        -e "SHOW VARIABLES" | grep read_only | awk '{print $2}'`
+        -ss -e "SHOW GLOBAL VARIABLES LIKE 'read_only'"
 
-    if [ "$read_only_state" = "ON" ]; then
+    if [ $read_only_state =~ '*ON' ]; then
```
krig: According to http://shellcheck.net, this has two issues:
* SC2074 Can't use =~ in [ ]. Use [[..]] instead.
* Don't quote rhs of =~, it'll match literally rather than as a regex. [1]

[1]: https://github.com/koalaman/shellcheck/wiki/SC2076

dmuhamedagic: On Mon, Feb 22, 2016 at 06:00:05AM -0800, Kristoffer Grönlund wrote:
> >  
> > -    if [ "$read_only_state" = "ON" ]; then
> > +    if [ $read_only_state =~ '*ON' ]; then
> 
> According to http://shellcheck.net, this has two issues:
> * SC2074 Can't use =~ in [ ]. Use [[..]] instead.

I think that [[..]] is a bashism, this is a /bin/sh script.

krig: `=~` is not in POSIX test either

dmuhamedagic: On Mon, Feb 22, 2016 at 07:14:15AM -0800, Kristoffer Grönlund wrote:
> >  
> > -    if [ "$read_only_state" = "ON" ]; then
> > +    if [ $read_only_state =~ '*ON' ]; then
> 
> `=~` is not in POSIX test either

Right, so it shouldn't be used. Just wanted to warn about
introducing stuff which aren't available in /bin/sh.

PR #407, opened 2014-04-09
Stateful: Make master/slave scores configurable (bnc#867372)

krig:
This patch makes the slave/master scores set configurable, in order
to enable the simulation of resources that should promote the other
clone if the current master fails.


krig:
Since there has been no response here from the author, I'm going to close this.

PR #432, opened 2014-05-15
High: vm.sh: ensure XML domain spec - domain name match

jnpkrn:
Otherwise, it could happen that the domain specified via xmlfile
is successfully started, but any later action (migration, etc.),
may fail or accidentally manipulate with a different domain(!).

Signed-off-by: Jan Pokorný jpokorny@redhat.com

### DIFF
--- a/rgmanager/src/resources/vm.sh
+++ b/rgmanager/src/resources/vm.sh
@@ -352,7 +352,22 @@ build_virsh_cmdline()
 		cmdline="$cmdline -c $OCF_RESKEY_hypervisor_uri"
 	fi
 
-	cmdline="$cmdline $operation $OCF_RESKEY_name"
+	case "$operation" in
+	create)
+		# assert specified xmlfile is known to exist,
+		# sanity check it defines expected domain name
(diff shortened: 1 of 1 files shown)


guessi:
@jnpkrn,

I've noticed that there's a file permission change in this PR (100755 --> 100644)

should be 0755, right?

jnpkrn:
@guessi, you are right this was certainly unintended.  Quick check discovered another instance of such a side-effect change: PR #424.
On the other hand, this is a violation of unified approach (already violated by `SAP{Database,Instance}`, not a real issue per-se:
`make install` will run `install` utility that sets executable bits automatically and during the build process, we extract the metadata by running `bash $AGENT` anyway.

jnpkrn:
That being said, if the maintainer deems appropriate, will fix the perms, indeed.

guessi:
@jnpkrn 
it's just a reminder for pointing out there's unwanted permission changed in this commit :)

and agree, it's not a big issue if permissions were guaranteed by Makefile,

jnpkrn:
@guessi all in all, good spot, thanks!

PR #468, opened 2014-07-25
VirtualDomain: add option to sync vm config via csync2

oshikore:
When save_config_on_stop AND sync_config_on_save is enabled the config gets synced with csync2 -x vm.conf to all nodes configured in csync2. This will make sure that wherever the VM starts it will inherit the config changes.

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -22,12 +22,20 @@ OCF_RESKEY_autoset_utilization_cpu_default="true"
 OCF_RESKEY_autoset_utilization_hv_memory_default="true"
 OCF_RESKEY_migrateport_default=$(( 49152 + $(ocf_maybe_random) % 64 ))
 OCF_RESKEY_CRM_meta_timeout_default=90000
+OCF_RESKEY_save_config_on_stop_default=false
+OCF_RESKEY_sync_config_on_stop_default=false
 
 : ${OCF_RESKEY_force_stop=${OCF_RESKEY_force_stop_default}}
 : ${OCF_RESKEY_autoset_utilization_cpu=${OCF_RESKEY_autoset_utilization_cpu_default}}
 : ${OCF_RESKEY_autoset_utilization_hv_memory=${OCF_RESKEY_autoset_utilization_hv_memory_default}}
 : ${OCF_RESKEY_migrateport=${OCF_RESKEY_migrateport_default}}
 : ${OCF_RESKEY_CRM_meta_timeout=${OCF_RESKEY_CRM_meta_timeout_default}}
+: ${OCF_RESKEY_save_config_on_stop=${OCF_RESKEY_save_config_on_stop_default}}
+: ${OCF_RESKEY_sync_config_on_stop=${OCF_RESKEY_sync_config_on_stop_default}}
+
+if ocf_is_true ${OCF_RESKEY_sync_config_on_stop}; then
+	OCF_RESKEY_save_config_on_stop="true"
+fi
 #######################################################################
 
 ## I'd very much suggest to make this RA use bash,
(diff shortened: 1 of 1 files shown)


oshikore:
pff, I misplaced some of the commit notes.
I tried to address the features asked.
Please review, and note that I had no system to test it on. Again :-/

dmuhamedagic:
Looks OK for the most part, but the sync_on_stop option is superfluous. Can we have that removed completely and do csync2 whenever save_on_stop is true. Also, please rebase the pull request to a single patch. Thanks!

oshikore:
thanks for the feedback.

The rationale for keeping the save config on stop option would be to help people who use shared FS approach, but still want to make sure the config files are saved when they are changed by virsh edit/virtmanager, or use a completely different method for copying config files around (rsync, whatever)

dmuhamedagic:
On Mon, Jan 05, 2015 at 04:39:52AM -0800, oshikore wrote:
> thanks for the feedback.
> 
> The rationale for keeping the save config on stop option would be to help people who use shared FS approach, but still want to make sure the config files are saved when they are changed by virsh edit/virtmanager, or use a completely different method for copying config files around (rsync, whatever)

OK, that looks like a plausible reason.

Could you just rebase your branch so that we have just one
commit/patch.

I'll see if I can do some testing.

oshikore:
I did the rebase, finally.
Let me know if anything else is needed.

Thanks.

dmuhamedagic:
@oshikore Unfortunately, the patch cannot be merged automatically. I'm not sure why, but it looks like you took something else than master for rebase.

Further, there are two new options in save_config():

+ virsh $VIRSH_OPTIONS dumpxml --inactive --security-info ${DOMAIN_NAME} > ${CFGTMP}

What exactly do they mean and shouldn't that be in a different commit? In particular, I was not able to understand '--inactive' from the virsh help.

oshikore:
gah, I'll try to figure it out. This is the first time I ever used git, I mostly have no idea what I am doing...
As for the --inactive:

dumpxml domain [--inactive] [--security-info] [--update-cpu] [--migratable]
    Output the domain information as an XML dump to stdout, this format can be used by the create command. Additional options affecting the XML dump may be used. --inactive tells virsh to dump domain configuration that will be used on next start of the domain as opposed to the current domain configuration. Using --security-info will also include security sensitive information in the XML dump. --update-cpu updates domain CPU requirements according to host CPU . With --migratable one can request an XML that is suitable for migrations, i.e., compatible with older libvirt releases and possibly amended with internal run-time options. This option may automatically enable other options (--update-cpu, --security-info, ...) as necessary.

The point is, virsh edit edits not the running config but the inactive copy of it. It will take effect on next start. This will represent the latest edits even if the VM has not yet been rebooted since the last edit. Additionally it removes a few attributes that are changing all the time, such as the domain-id (sequential monotonically increasing counter). This makes it possible to compare our older config with the current one.

dmuhamedagic:
On Tue, Jan 20, 2015 at 04:39:06AM -0800, oshikore wrote:
> gah, I'll try to figure it out. This is the first time I ever used git, I mostly have no idea what I am doing...

Heh, I thought that you were doing just fine :)
Let me know if/where you have problems and we can try to figure
it out. I'm not a git expert either, but mostly manage to get by.

> As for the --inactive:
> 
> dumpxml domain [--inactive] [--security-info] [--update-cpu] [--migratable]
>     Output the domain information as an XML dump to stdout, this format can be used by the create command. Additional options affecting the XML dump may be used. --inactive tells virsh to dump domain configuration that will be used on next start of the domain as opposed to the current domain configuration. Using --security-info will also include security sensitive information in the XML dump. --update-cpu updates domain CPU requirements according to host CPU . With --migratable one can request an XML that is suitable for migrations, i.e., compatible with older libvirt releases and possibly amended with internal run-time options. This option may automatically enable other options (--update-cpu, --security-info, ...) as necessary.
> 
> The point is, virsh edit edits not the running config but the inactive copy of it.

virsh is my favourite nowadays.

> It will take effect on next start. This will represent the latest edits even if the VM has not yet been rebooted since the last edit. Additionally it removes a few attributes that are changing all the time, such as the domain-id (sequential monotonically increasing counter). This makes it possible to compare our older config with the current one.

So, --inactive should stay in?

oshikore:
It should. If we take it out, the RA will keep saving the configuration, each time the machine is stopped and started - because the domain's numerical id might change. I had a bit of trouble figuring that one out.

dmuhamedagic:
This pull request superceded.

PR #473, opened 2014-07-31
added parameter force_migration_uri

atemp:
When doing live migration add the possibility to enforce passing migrateuri to virsh.

Before, that happened only when migration_network_suffix was used. In a scenario where migration_network_suffix is "" but you still want migrateuri to be forwarded to virsh you can now enforce it. This is useful e.g. if you have multiple networks configured but still your suffix would be empty.


dmuhamedagic:
Superseeded by #558.

PR #509, opened 2014-11-03
Fix simple transposition of options for TESTURL

inetu-dsix:
The current script will use `status10url` for both level 10 and level 20 monitor operations, but the documentation indicates that `testurl` will be used for a level 20 monitor.

This patch restores the documented behavior.

### DIFF
--- a/heartbeat/nginx
+++ b/heartbeat/nginx
@@ -31,7 +31,7 @@
 #  OCF_RESKEY_status10regex
 #  OCF_RESKEY_status10url
 #  OCF_RESKEY_client
-#  OCF_RESKEY_testurl
+#  OCF_RESKEY_test20url
 #  OCF_RESKEY_test20regex
 #  OCF_RESKEY_test20conffile
 #  OCF_RESKEY_test20name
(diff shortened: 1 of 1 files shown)


inetu-dsix:
@dmuhamedagic There are two problems I am looking to correct here, let me know if you disagree with my assessment of the issues.

1. Documentation not matching function of the Level 20 check (Documentation says `testurl` but code uses `status10url`)
1. Current code will use `status10url` for Level 10 and Level 20 checks, but it seems that there should be an option to use different URLs

I see two possible solutions:

1. Update code to use `testurl` for the Level 20 check as documented.
1. Choose a different parameter name (perhaps `test20url`) and update both code and documentation

My patch implements the first option, but I am open to other solutions if they are preferred. Let me know your preference and I will gladly rework the patch as needed.

dmuhamedagic:
On Wed, Jan 07, 2015 at 07:59:36AM -0800, inetu-dsix wrote:
> @dmuhamedagic There are two problems I am looking to correct here, let me know if you disagree with my assessment of the issues.
> 
> 1. Documentation not matching function of the Level 20 check (Documentation says `testurl` but code uses `status10url`)
> 1. Current code will use `status10url` for Level 10 and Level 20 checks, but it seems that there should be an option to use different URLs

Right.

> I see two possible solutions:
> 
> 1. Update code to use `testurl` for the Level 20 check as documented.
> 1. Choose a different parameter name (perhaps `test20url`) and update both code and documentation
> 
> My patch implements the first option, but I am open to other solutions if they are preferred. Let me know your preference and I will gladly rework the patch as needed.

Using testurl for level 20 sounds confusing (as if the current
set of options weren't as confusing as possible). So, I'd vote
for the latter option:

- introduce status20url for level 20 monitor
- remove testurl completely

What do you say?

BTW, if you rebase the branch (preferably to just one patch for
this issue), you can "git push -f" to update the request.

inetu-dsix:
@dmuhamedagic Updated and rebased.

I decided on `test20url` instead of `status20url` since the other parameters are `test20regex`, `test20conffile`, and `test20name`

How does this look to you?

dmuhamedagic:
Good. Thanks!

Issue #526, opened 2014-12-11
iSCSILogicalUnit - Creating backstore/block fails with with lio-t and targetcli-fb


sq9mev:
```
iSCSILogicalUnit(p_lun_storage_xh1_0)[2745]:    2014/12/11_09:31:58 ERROR: Unexpected keyword parameter 'write_back'.
```

The problem originates in https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/iSCSILogicalUnit#L365

targetcli-fb interface changed in https://github.com/agrover/targetcli-fb/commit/198ea6cbcb6adea2af7da2d5104eb64edd338d7f probably due to changes in kernel:
https://lkml.org/lkml/2014/6/9/199

The simplies soolution is to remove  write_back=False in iSCSILogicalUnit#L365, but i'm not sure how does it cope with Datera targetcli (https://github.com/Datera/targetcli/blob/master/targetcli/ui_backstore.py#L366).

Any thoughts?

dmuhamedagic:
Fixed.

PR #532, opened 2014-12-18
#526, do not set write_back when creating backstore/block

sq9mev:
This works fine in my development enviroment.


dmuhamedagic:
Thanks!

PR #533, opened 2014-12-19
Low: use sh for :* removal and maths

grooverdan:
sh can remove :* and add

not a bashism


dmuhamedagic:
These changes are superfluous.

PR #540, opened 2015-01-06
Medium: ldirectord: Add Install section for systemd unit file

MichiroHibari:
ldirectord failed to systemctl enable, with following error:

```
# systemctl enable ldirectord
The unit files have no [Install] section. They are not meant to be enabled
using systemctl.
Possible reasons for having this kind of units are:
1) A unit may be statically enabled by being symlinked from another unit's
   .wants/ or .requires/ directory.
2) A unit's purpose may be to act as a helper for some other unit which has
   a requirement dependency on it.
3) A unit may be started when needed via activation (socket, path, timer,
   D-Bus, udev, scripted systemctl call, ...).
```

This commit enable to work "systemctl enable ldirectord".

### DIFF
--- a/ldirectord/systemd/ldirectord.service.in
+++ b/ldirectord/systemd/ldirectord.service.in
@@ -10,3 +10,6 @@ ExecReload=@sbindir@/ldirectord reload
 PIDFile=/var/run/ldirectord.ldirectord.pid
 Type=forking
 KillMode=none
+
+[Install]
+WantedBy=multi-user.target


dmuhamedagic:
@krig Kristoffer, you worked on this systemd file, can you please take a look.

krig:
Setting WantedBy= to multi-user.target will automatically start ldirectord at system startup. Generally, if ldirectord is managed by a cluster resource, it shouldn't be enabled in systemd to avoid systemd starting or stopping the resource by itself.

Is ldirectord usually not managed by the cluster?

dmuhamedagic:
I don't know, but it's certainly no good to automatically start it at boot. Isn't there something else that can be put into the Install section?

MichiroHibari:
Thanks for the comments.

> Is ldirectord usually not managed by the cluster?

Yes.
Often, ldirectord is used at no cluster environment.
Such as simple LVS load-balancer.

So far ldirectord can "chkconfig on ldirectord",
I think it's necessary "systemctl enable ldirectord".

> I don't know, but it's certainly no good to automatically start it at boot.

I see what you mean.
No good to automatically start at cluster environment.
But ldirectord not automatically start unless otherwise
execute "systemctl enable ldirectord".
Even if setting WantedBy= to multi-user.target,
default ldirectord unit file state is disabled ( not automatically start ).

krig:
Ah, I see, perhaps I have misinterpreted this behavior. In that case, I agree that this patch should be merged. Thank you!

dmuhamedagic:
This cannot be merged anymore. @MichiroHibari can you please rebase the pull request (you just need then to git push -f).

MichiroHibari:
Sorry for my English.

Rebase and pushed. Thank you!

PR #542, opened 2015-01-06
low: ldirectord: modify manual and sample for Radius check

MichiroHibari:
"password" option is not implemented.
"login" option is used for the attribute User-Name.

By the way, modify typo "::"->":"


dmuhamedagic:
Looks OK. Thanks!

PR #543, opened 2015-01-07
Use calculated default pidfile.

seabres:


seabres:
This was also my first intention, just to remove the PID file.
But what happens if parameter kamuser changes from one use to another ?
The directory remains, will not be created and therefore not writable by new kamuser. Same error as before.
This is not so unlikely as it sounds.
Most installations start as root to see it working anyhow and come quickly to the conclusion, it is not so good idea to run kamailio as root, it is better to run as ordinary user.
-> crm resource param reskam1 set kamuser kamailio
We could say, if you change parameters, you have to repair this by yourself. Event if the user delete the resource and recreate it, the error remain, it has to be created with a different resource name.
But isn´t it a good principle for a resource agent to be able to handle parameter changes without errors, especially when it is so easy to implement.
So, RA try to remove the complete directory, when it is created by itself.

dmuhamedagic:
On Fri, Jan 09, 2015 at 03:59:27AM -0800, Rainer Brestan wrote:
> Currently i am very skeptic we find a common way of handling PID files in RA.

What I meant was just the PID files directory, how particular
application interprets the content is another matter. Though I
think that there's some PID file standard which says that the PID
must be in the first line by itself. Or thereabouts :)

OK, I believe that we can merge this. Thanks for the good work!

dmuhamedagic:
Oh, I already did merge it :) Too many interruptions today.

PR #544, opened 2015-01-12
new RA sg_persist

lhwang:
sg_psersist is a resource agent that implements SPC-3 Persistent
Reservations (contribution from Evgeny Nifontov).

Customers with an active-passive storage setup using MD-raid1 need
to ensure, that the MD device is only activated on one system at a
time. While this is clearly possible with the existing Raid1 RA,
there is still the possibility that crazy system administrators
assemble such devices manually from the command line, leading to
a complete data-loss as a result...

There exists an implementation of a resource agent around the "sg_persist"
Utility from the sg3_utils package. The sg_persist RA is just based on
the sg3_utils tools. It works as a Master/Slave resource. It registers a
reservation key for devices when the resource starts and holds reservation
when the resource promotes to be a Master. There exist only one Master
in case of the reservation type is 1/3/5/6 and can have multi Masters
in case of the reservation type is 7/8.

Signed-off-by: Evgeny Nifontov
Signed-off-by: Liuhua Wang lwang@suse.com

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -132,6 +132,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_scsi2reservation.7 \
                           ocf_heartbeat_sfex.7 \
                           ocf_heartbeat_slapd.7 \
+                          ocf_heartbeat_sg_persist.7 \
                           ocf_heartbeat_symlink.7 \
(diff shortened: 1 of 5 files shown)


dmuhamedagic:
ocft tests passed. Merging. Still need to remove OCF_RESOURCE_INSTANCE from ocf_log calls.

PR #566, opened 2015-01-28
Minor updates to CTDB resource agent

jarrpa:
The main change here is the use of syslog by default. Productions systems SHOULD use syslog, as CTDB doesn't handle logging to a file directly very well.

The change for the socket file location is a suggestion based on the default CTDB behavior.

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -822,6 +822,7 @@ ocf_start_trace() {
 	fi
 	PS4='+ `date +"%T"`: ${FUNCNAME[0]:+${FUNCNAME[0]}:}${LINENO}: '
 	set -x
+	env=$( echo; printenv | sort )
 }
 ocf_stop_trace() {
 	set +x


dmuhamedagic:
@ddiss David, could you check this one please.

ddiss:
The whitespace and socket path changes look okay - upstream ctdb now uses the default CTDB_RUNDIR/ctdbd.socket path, so I think we should do the same in the RA master branch

I'd don't think the syslog default change is suitable as-is. ctdbd currently defaults to a file backed log, presumably due to the reason documented in the ctdbd.1 man page:
  Under heavy loads syslog(3) can block if the syslog
  daemon processes messages too slowly.  This can
  cause CTDB to block when logging.

jarrpa:
I'm not seeing that warning in my man pages... where are you seeing that?

ddiss:
ctdb/doc/ctdbd.1.xml in the Samba master branch...
https://git.samba.org/?p=samba.git;a=blob;f=ctdb/doc/ctdbd.1.xml;hb=refs/heads/master#l160

jarrpa:
Hmm. I'll look into that. Thanks!

jarrpa:
So it looks like CTDB does not in fact respond to SIGHUPs. It seems our options right now are:

1. Use file logging without log rotation, and possibly risk filling up all disk space
2. Use syslog but possibly block under heavy load

...is one better than the other?

There are patches in Samba master for CTDB that provide a non-blocking mechanism for syslog, however those have not been backported into Samba 4.2 yet (which includes CTDB now), nor has Samba 4.2. been officially released yet.

ddiss:
Thanks for the follow-up José.
My preference is still to take the whitespace cleanup and OCF_RESKEY_ctdb_socket changes only for now.

jarrpa:
Works for me! ...do I need create another pull request now with only the first two commits, or push a revert commit? Apologies, I don't have much GitHub experience. :)

ddiss:
Please submit a new pull request from a branch that only includes the two changes. @dmuhamedagic has final say on whether it gets merged or not.

PR #573, opened 2015-02-12
Fix virtual domain

yuusuke:
The following 2 points were corrected.
 * Fixed to correctly get the return code of the "virsh destroy" command
 * Force the "virsh destroy" of output in C


dmuhamedagic:
On Thu, Feb 12, 2015 at 02:12:34AM -0800, Yusuke Iida wrote:
> The following 2 points were corrected.
>  * Fixed to correctly get the return code of the "virsh destroy" command
>  * Force the "virsh destroy" of output in C

Very good. Thanks!

PR #577, opened 2015-02-19
[WIP] misc portblock fixes

aspiers:
This is a draft - I'm not a TCP/IP guru so someone else should verify my changes regarding TCP RST vs. ICMP port unreachable (although I did test them myself so I'm pretty sure I'm right).

Also I didn't yet test the `portno` type change or the example `crm configure` command.

### DIFF
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -39,7 +39,7 @@ usage()
 
 	$CMD is used to temporarily block ports using iptables.
 
-	It can be used to turn off a port before bringing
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
I'm also not a network expert, but the intention is clear. Furthermore, it is certainly possible in some situations that the Linux kernel generates a reset. Thanks for the update!

aspiers:
Updated with some suggested fixes.

PR #596, opened 2015-04-13
Remove LVM RA limit of not supporting volume group with no logical volum...

lhwang:
...e

It is not convinient now to a setup a LVM RA especially when using hawk
wizard. For example:
After configured a clvmd resource, we have to pause the configuration wizard;
Then create a vg and lv;
Then configue a LVM resource and start resources.

If remove the limit, we can:
Firstly create a clustered vg;
Then configure clvmd, LVM resoure and start resources;
At last create lv within the vg.

Also, when users configured a LVM resource with a vg with no lv,
the LVM resource will start failed and then stop clvmd resource too.
If users want to fix this they have to remove the LVM resource
and then create lv within the vg and then readd the LVM resource.

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -293,7 +293,7 @@ set_tags()
 }
 
 get_lv_number() {
-	echo "`vgdisplay $1|sed -n '/Cur LV/p'|awk '{print $3}'`"
+	vgdisplay $1|awk '/Cur LV/{print $3}'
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
We had to drop vgdisplay from status/monitor because of the timing issues. Can you figure another way to find out if the VG is without lvs?

PR #599, opened 2015-04-17
Remove LVM RA limit of not supporting volume group with no logical volum

lhwang:
It is not convinient now to a setup a LVM RA especially when using hawk
wizard. For example:
After configured a clvmd resource, we have to pause the configuration wizard;
Then start clvmd resource and create a vg with lv;
Then configue a LVM resource and start resources.

If remove the limit, we can:
Firstly create a clustered empty vg;
Then configure clvmd, LVM resoure togeter and start resources;
At last create lv within the vg.

Also, when users configured a LVM resource using a vg with no lv,
the LVM resource will start failed and then stop clvmd resource too.
If users want to fix this they have to remove the LVM resource
and then create lv within the vg and then readd the LVM resource.

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -293,6 +293,18 @@ set_tags()
 }
 
 #
+#	Return whether volume group is empty(0:empty; 1: contains lvs)
+#
+vg_is_empty() {
+	grep "logical_volumes" /etc/lvm/backup/$1
+	if [ $? -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
@lhwang Do we still need this one? I can vaguely recall some discussion on some ML regarding this.

dmuhamedagic:
Closing.

Issue #133, opened 2012-09-20
mysql RA: tmp files


ghost:
Hi,

I have patched the version to not use tmpfile on disk. It saves I/O and usage spaces. Here the code:
http://pastebin.com/zactvNXL

And also, log typo error with demote, promote.

Best regards,

PR #602, opened 2015-04-21
Update mysql: temporary files are not removed

daniesteban:
During the monitoring the temporary files generated in get_slave_info() are not removed, when variables are already defined the function does not assign any value to variable tmpfile.
This problem may cause the filling of the tmpfs partition.

Related issue: mysql RA: tmp files #133


daniesteban:
Forget it, that changes may cause worst mistakes during role change. But certainly the calls to get_slave_info do it from set_master() and check_slave() create check_slave.* files that after don't delete.

PR #611, opened 2015-05-09
redis: use masterauth password to connect

michael-dev:
This fixes redis monitoring and startup when masterauth is set.


dmuhamedagic:
This cannot be merged, please rebase first. Don't know anything about redis, but I suppose that this should be otherwise safe.

PR #614, opened 2015-05-19
Dev: Add Travis CI job which runs shellcheck on all scripts

krig:
This adds a Travis CI which invokes shellcheck on all scripts
in the repository. It doesn't currently signal failure, but the
build job output can be inspected for any issues that should
be fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: bash
+install:
+  - ./ci/install.sh
+script:
+  - ./ci/build.sh
+notifications:
+  email: false
+sudo: required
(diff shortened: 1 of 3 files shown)


dmuhamedagic:
On Tue, May 19, 2015 at 05:26:45AM -0700, Kristoffer Grönlund wrote:
> This adds a Travis CI which invokes shellcheck on all scripts
> in the repository. It doesn't currently signal failure, but the
> build job output can be inspected for any issues that should
> be fixed.

Sort of hard to argue with this one :) Let's see what will it
produce. It's very good that the build won't fail, as we have RAs
of various quality.

PR #616, opened 2015-05-29
Fixed https://bugzilla.redhat.com/show_bug.cgi?id=760251 for Redis 

elconas:
If prelink is enabled, glibc was updated and redis was not restarted (it can not be restartet actually due to this bug, so we must disable redis, update glibc and start redis again), failover / restart of the redis service failes due to open bug https://bugzilla.redhat.com/show_bug.cgi?id=760251 on RHEL / CentOS.

This affects ALL resource agents relying on "pidof" for determining process state of procted processes. So this may be a larger Problem for all clusters :)

### DIFF
--- a/heartbeat/redis
+++ b/heartbeat/redis
@@ -172,6 +172,21 @@ function last_known_master()
 	echo "$MASTER_HOST"
 }
 
+## Pidof does not work with prelink on centos
+##    see https://bugzilla.redhat.com/show_bug.cgi?id=760251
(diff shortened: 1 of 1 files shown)


oalbrigt:
This seems like it was fixed in sysvinit. Is it still an issue in any supported version of RHEL/CentOS or any other distro?

dmuhamedagic:
I think that, given that has already been fixed in the affected distribution, we can close this.

PR #619, opened 2015-06-11
Medium: pgsql: delete an old replication slot when creating the slot.

ytakeshita:

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -505,6 +505,25 @@ pgsql_methods() {
 EOF
 }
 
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Looks good otherwise. At least from the scripting perspective, I
suppose that you do take good care of the postgresql details :)

ytakeshita:
@dmuhamedagic 
Thanks for your comments.
I reflect your comments.
ref #634

ytakeshita:
@dmuhamedagic
Thanks for your comment.
I reflect your comment.
ref #646

PR #629, opened 2015-06-26
VirtualDomain: New resource attributes migration_speed and migration_downtime

psyhomb:
New VirtualDomain resource attributes

**migration_downtime**
Allow live migration of machines with higher memory pressure (milliseconds)

You can for example try to run stress test (to simulate higher memory pressure) inside guest os (cpu:2, ram:2GB):
```bash
stress --cpu 1 --vm 13 --vm-bytes 128M --vm-hang 1 --timeout 900s 
```
on the host side try to run live migration of that same vm with pcs (this cmd will probably timeout at the end and machine will be relocated (stop/start)):
```bash
pcs resource move testvm1 host2
```
then try the same cmd but with a **migration_downtime** attribute in place:
```bash
pcs resource update testvm1 migration_downtime=1000
pcs resource move testvm1 host2
```

**migration_speed**
Set the maximum migration bandwidth (in MiB/s) for a domain which is being migrated to another host
```bash
pcs resource update testvm1 migration_speed=100
```

Cheers

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -17,6 +17,8 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
 # Defaults
+OCF_RESKEY_migration_downtime_default=0
+OCF_RESKEY_migration_speed_default=0
 OCF_RESKEY_force_stop_default=0
 OCF_RESKEY_autoset_utilization_cpu_default="true"
 OCF_RESKEY_autoset_utilization_hv_memory_default="true"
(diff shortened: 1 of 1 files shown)


isavcic:
Great stuff, necessary additions to the RA.

digimer:
Flipping wonderful!

isavcic:
Sorry for pinging you directly @dmuhamedagic, but I believe that this PR should be reviewed sooner rather than later, because it addresses an important issue of a VM being stopped and started when a migration fails, and with current VirtualDomain RA it fails often in real production, where memory pressure is high so migration never finishes within the timeout period, no matter how high it's set. Unexpected VM restarts because of failed migration are not acceptable, IMHO.

On the other hand, there's also the issue that in some cases the migration consumes too much bandwidth and saturates the host network, but with this patch it's tunable. Also worth mentioning is that lowering the speed, without this patch, leads to a greater chance of migration failing due to memory pressure and migration not being able to keep up. 

Thanks for your time! Hvala!

dmuhamedagic:
On Tue, Oct 06, 2015 at 02:52:04AM -0700, Ivan Savčić wrote:
> Sorry for pinging you directly @dmuhamedagic, but I believe that this PR should be reviewed sooner rather than later, because it addresses an important issue of a VM being stopped and started when a migration fails, and with current VirtualDomain RA it fails often in real production, where memory pressure is high so migration never finishes within the timeout period, no matter how high it's set. Unexpected VM restarts because of failed migration are not acceptable, IMHO.
> 
> On the other hand, there's also the issue that in some cases the migration consumes too much bandwidth and saturates the host network, but with this patch it's tunable. Also worth mentioning is that lowering the speed, without this patch, leads to a greater chance of migration failing due to memory pressure and migration not being able to keep up. 
> 
> Thanks for your time! Hvala!

No need to apologize. I'll try to get to it hopefully this week.
I'm not a libvirt expert let alone a migration expert, perhaps
Florian can take a look too. @fghaas, what do you say?

dmuhamedagic:
Can you please rebase this so that we have a single patch. It'd be easier to follow the history.

psyhomb:
I have done a rebase

dmuhamedagic:
Great! Many thanks for the contribution. And patience! :)

psyhomb:
np, thank you

Issue #630, opened 2015-06-27
iSCSITarget: default portal conflict


pemer:
If you are useing the lio-t backend and the setting portal are defined != 0.0.0.0:3260 the targetd can not create a iscsi target, because the portal settings are in conflict.
At the moment the scripts send through targetcli following:

example:
```
targetcli /iscsi create iqn.2015-10.org.example.local
```

the targetd will create a default portal 0.0.0.0:3260. And this setting are in conflict, if in cluster a portal a defined.
The easiest way to fix the behaviour is, to set a global option that the targetd not creating a default  portal, if a user defined portal a configured.

example:
```
targetcli /iscsi set global auto_add_default_portal=false
```

pemer:
this changes will fix this behavior from targetd

sammcj:
+1 - this is quite annoying.

fghaas:
I've commented on the PR; please amend the commit as suggested and then force-push to your fork — happy to merge after that unless there are any objections. Thanks for the patch!

PR #631, opened 2015-06-27
iSCSITarget: fix portal bug, the default iscsi portal settings are no…

pemer:
please see bug: #630

…t loaded, if a target are createt

If a portal are defined, the target could not createt, because the default portal 0.0.0.0:3260 are loaded immediately, if the iISCSITarget skripte create a new target.
The global settings auto_add_default_portal=false must be configured, so that the targetd knows that no default portal are createt.
This settings are not conflicting, if there are sendet to targetd servals times.

### DIFF
--- a/heartbeat/iSCSITarget
+++ b/heartbeat/iSCSITarget
@@ -326,10 +326,13 @@ iSCSITarget_start() {
 		# automatically creates the corresponding target if it
 		# doesn't already exist.
 		for portal in ${OCF_RESKEY_portals}; do
-			ocf_run targetcli /iscsi create ${OCF_RESKEY_iqn} || exit $OCF_ERR_GENERIC
 			if [ $portal != ${OCF_RESKEY_portals_default} ] ; then
+				ocf_run targetcli /iscsi set global auto_add_default_portal=false || exit $OCF_ERR_GENERIC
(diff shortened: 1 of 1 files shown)


fghaas:
This patch looks good to me and it I've tested it to see that indeed behaves as expected. Thank you!

Can you please amend the commit to

(1) Contain a commit message in understandable English? As a suggestion, it could look and sound roughly like this:

```
iSCSITarget: properly create portals for lio-t implementation

Even when defining the "portals" parameter, targetcli creates a
default portal at 0.0.0.0:3260 if the auto_add_default_portal option
is set to true (default), causing the resource to fail with
OCF_ERR_GENERIC.

Suppress creation of the default portal by setting
auto_add_default_portal=false if the "portals" parameter is set,
and retain the original behavior if it is not.
```

(2) set the author email address in the commit to a proper recipient address; `titan.afdata.local` does not resolve in DNS.

Thank you!

pemer:
Dear Florian,

I hope you are now satisfied and you can merge the code.

regards Patrick

fghaas:
@pemer, thanks for updating the commit message — it's almost good to merge, but we still need a proper email address in the commit.

```
git commit --amend --author="Patrick Emer <your@preferred.email>"
git push -f
```

Thank you!

fghaas:
Superseded by #647.

PR #634, opened 2015-06-29
Fix initialize replication slot2

ytakeshita:

Reflects following comments #619


ytakeshita:
Sorry, I mistake the title of this pull request...
If you worry about this, I'll close this pull request, and create new one.

PR #639, opened 2015-07-03
Fix race condition with iSCSITarget with lio-t

### LINKED ISSUE #637: iSCSILogicalUnit lio-t implementation possible race condition
I've had great problems with the ocf provider for iSCSILogicalUnit when using lio-t where starting or migrating the resource fails until it's cleaned up.

What I've narrowed it down to is that the provider doesn't seem to wait long enough for `targetcli` to process and action the issued commands and may often result in configuration being only partially finished applying when the next command runs.

For example:

`Start iSCSITarget [Success] -> Start iSCSILogicalUnit [Failure!] -> Restart iSCSILogicalUnit [Success]`

I think this is because restarting the iSCSILogicalUnit resource with partial configuration is quicker than starting the resource.

- I've made and example video visualising the issue here: https://www.youtube.com/watch?v=_oSGYiokKU8&feature=hd=1
- Debug output with comments on the events: https://gist.github.com/sammcj/4e7e4d5c2075626337a9
- Cluster configuration: https://gist.github.com/sammcj/4e7e4d5c2075626337a9#comment-1484747
- CentOS 7
- Pacemaker 1.1.12 (23.el7)
- Resource-agents 3.9.5 (40.el7_1.4)

Impact: As far as I can tell this bug prevents iSCSILogicalUnit from being used with lio-t at all which is the only option with CentOS / RHEL 7.

sammcj:
Fixes #637

When using lio-t (targetcli) you should not set the allowed_initiators in the iSCSITarget and instead set it in iSCSILogicalUnit.

If allowed_initiators is set in iSCSITarget you will likely encounter a race condition where by the iSCSILogicalUnit must be cleaned-up before it will start as ACL generation would have been duplicated by iSCSITarget.

### DIFF
--- a/heartbeat/iSCSITarget
+++ b/heartbeat/iSCSITarget
@@ -111,6 +111,7 @@ Allowed initiators. A space-separated list of initiators allowed to
 connect to this target. Initiators may be listed in any syntax
 the target implementation allows. If this parameter is empty or
 not set, access to this target will be allowed from any initiator.
+When using lio-t you should set this in iSCSILogicalUnit - not here.
 </longdesc>
 <shortdesc lang="en">List of iSCSI initiators allowed to connect
 to this target</shortdesc>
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
No response. Closing.

PR #643, opened 2015-07-15
iSCSILogicalUnit improvements

### LINKED ISSUE #640: iSCSILogicalUnit incorrectly reports resources as stopped for lio-t implementation
In contrast to `lio`, with `lio-t` we're not checking for leftover configfs items in `/sys/kernel/target/core`. That means that a resource where an iSCSI LUN has been deconfigured, but its corresponding iblock has not, a resource is being reported as properly stopped even though it still has handles on underlying block devices open. This obviously wreaks havoc if the underlying block device is also a cluster-managed resource.

### LINKED ISSUE #641: iSCSILogicalUnit should return OCF_NOT_RUNNING if backing device does not exist
If `iSCSILogicalUnit` exposes a local LVM logical volume or even a hardware drive, then it is safe to assume that the device will "always" be there. However, if what we are exposing is itself a cluster resource, such as an RBD volume or a DRBD device, then the backing device will appear and disappear as needed.

It is safe to assume that if the backing file or block device does not exist, then it cannot be exported as an `iSCSILogicalUnit`. Even if it **has** been previously exported as an `iSCSILogicalUnit`, then it should be impossible for it to just disappear or be removed while open, as such an action would return `-EBUSY` while the device is open. If it disappears because the drive has physically been yanked, then that is a condition in which throwing up our hands and fencing the node is the proper action.

### LINKED ISSUE #642: iSCSILogicalUnit should remove lio-t backing store entries, not just iSCSI exports, on stop
During the `stop` operation, `targetcli delete` returns an error if the object we're about to delete already no longer exists. Since `stop` is a force operation and we must do anything we can to clean up the resource, don't exit on that condition but log a warning and proceed to removal of the backstore. If *that* fails, then we error out and potentially cause the node to be fenced.

fghaas:
This PR contains three improvements to `iSCSILogicalUnit`. One is generic, the other two are specific to the `lio-t` implementation.

- 1b3ac8f extends behavior previously present just for `lio` to `lio-t`, in that it checks for existence of the backing store ConfigFS entry in stop, as opposed to just that for the iSCSI LUN. Fixes #640.
- 7e051ab is being more persistent on stop. `targetcli` will fail when attempting to delete a LUN that has already been deleted; previously we would bail out then and there letting the backing store entries linger (and keeping underlying block devices busy). Fixes #642.
- aae6d1b is generic and checks the existence of the underlying block device on monitor. If the backing device is not there, we can safely assume that the LU is not exported. Fixes #641.

Issue #631 is related and is pending a merge.

@davidvossel, I believe you wrote the original `lio-t` implementation — would you mind taking a look?

### DIFF
--- a/heartbeat/iSCSILogicalUnit
+++ b/heartbeat/iSCSILogicalUnit
@@ -429,15 +429,21 @@ iSCSILogicalUnit_stop() {
 		fi
 		;;
 	lio-t)
-		ocf_run targetcli /iscsi/${OCF_RESKEY_target_iqn}/tpg1/luns delete ${OCF_RESKEY_lun} || exit $OCF_ERR_GENERIC
+		# "targetcli delete" will fail if the LUN is already
+		# gone. Log a warning and still push ahead.
+		ocf_run -warn targetcli /iscsi/${OCF_RESKEY_target_iqn}/tpg1/luns delete ${OCF_RESKEY_lun}
 		if [ -n "${OCF_RESKEY_allowed_initiators}" ]; then
 			for initiator in ${OCF_RESKEY_allowed_initiators}; do
 				if targetcli /iscsi/${OCF_RESKEY_target_iqn}/tpg1/acls/${initiator} status | grep "Mapped LUNs: 0" >/dev/null ; then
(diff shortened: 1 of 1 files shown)


fghaas:
Haven't seen any comments on this one, so I'm about to merge.

fghaas:
Merged after rebasing on current master.

PR #646, opened 2015-07-22
Medium: pgsql: delete an old replication slot when creating the slot.

ytakeshita:

Reflects following comments #619, #634


dmuhamedagic:
Thanks!

PR #649, opened 2015-07-28
ensure docker image name is valid before run

guessi:
container name may formatted as follow:
- image
- image:tag
- repository/image
- repository/image:tag

rule for container/repository name could be found at:
- https://github.com/docker/distribution/blob/v2.2.0/reference/regexp.go#L6-L16

rule for image tag:
- https://github.com/docker/distribution/blob/v2.2.0/reference/regexp.go#L24-L25

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -381,11 +381,34 @@ image_exists()
 	return 1
 }
 
+docker_image_name_validate()
+{
+	# ensure docker image name is valid before run, image name may be formatted as follow:
+	# - image
+	# - image:tag
+	# - repository/image
+	# - repository/image:tag
+	#
+	# rule for container/repository name could be found at:
+	# - https://github.com/docker/distribution/blob/v2.2.0/reference/regexp.go#L6-L16
+	#
+	# rule for image tag:
+	# - https://github.com/docker/distribution/blob/v2.2.0/reference/regexp.go#L24-L25
+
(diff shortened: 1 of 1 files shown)


guessi:
a simple script for this commit,
https://gist.github.com/d58ad8d775f6dba205f4.git

guessi:
a simple script for this commit,
https://gist.github.com/guessi/ba752cb412fc77685495

guessi:
@davidvossel , could you review this commit for me? thanks!!!

dmuhamedagic:
Not sure who's going to validate this monster regex :-/

guessi:
Oops... almost forgot I've commit this PR.

yes, it's an ugly regex check.

it's just a simple idea to do basic protection for preventing docker agent fall into unmanaged state. Just that simple.

docker is undergoing rapid change everyday, regex rule may be changed too, not sure if it could applied in future?

if we assume "image name" is always valid, close this PR is okay :)

dmuhamedagic:
On Wed, Jan 13, 2016 at 05:41:34AM -0800, guessi, mei wrote:
> docker is undergoing rapid change everyday, regex rule may be changed too, not sure if it could applied in future?

Right. If they had valid name rule spelled out somewhere, it
would have been easier to deal with it.

guessi:
if we assume people who use this agent SHOULD have basic knowledge to Docker, I'd close it, but still need proper handle the error from docker...

let's leave this PR open, listen to other's opinions :p

PR #650, opened 2015-07-28
Resource agent :Heartbeat :Raid1: fix RA cannot  manage include of start and stop Raid1 service.

zhilongliu:
Resource agent : Heartbeat : Raid1
redefine $mddev in function raid1_monitor_one and fix RA cannot  manage Raid1. The mddev comes directly from the configuration file (OCF_RESKEY_raidconf, typically /etc/mdadm.conf) from the ARRAY stanza. The mdadm.conf man page says: If the name does not  start  with  a  slash  ('/'),  it  is  treated as being in /dev/md/.  This patch adds the "/dev" prefix instead.

### DIFF
--- a/heartbeat/Raid1
+++ b/heartbeat/Raid1
@@ -360,6 +360,7 @@ raid1_monitor_one() {
 	# check if the md device exists first
 	# but not if we are in the stop operation
 	# device existence is important only for the running arrays
+	echo $mddev | grep -q -o "/dev/" || mddev="/dev/$mddev"
 	if [ "$__OCF_ACTION" != "stop" -a ! -b $mddev ]; then
 		ocf_log info "$mddev is not a block device"
 		return $OCF_NOT_RUNNING


dmuhamedagic:
On Mon, Jul 27, 2015 at 07:41:06PM -0700, zhilongliu wrote:
> Resource agent : Heartbeat : Raid1
> redefine $mddev in function raid1_monitor_one and fix RA cannot  manage Raid1. The mddev comes directly from the configuration file (OCF_RESKEY_raidconf, typically /etc/mdadm.conf) from the ARRAY stanza. The mdadm.conf man page says: If the name does not  start  with  a  slash  ('/'),  it  is  treated as being in /dev/md/.  This patch adds the "/dev" prefix instead.

Why does it add "/dev" and not "/dev/md"?

The fix should also take place in list_conf_arrays(), because
that's where we fetch arrays from mdadm.conf.

zhilongliu:
hi, Dejan;
   your suggestion is quite valuable, and I still have another question 
in code.

# vim ./resource-agents/heartbeat/Raid1
523 if [ "$MDDEV" = "auto" ]; then
524         RAIDDEVS=`list_conf_arrays`
525 else
526         RAIDDEVS="$MDDEV"
527 fi

    I didn't get the meaning of "$MDDEV" = "auto", and according to the 
testing, code goes the "RAIDDEVS="$MDDEV"" branch actually, thus here is 
my question:  what's the auto meaning and does this condition exist 
defect? I think it's the mainly part of this patch drafting.


Thanks very much,
-Zhilong

dmuhamedagic:
On Tue, Jul 28, 2015 at 07:08:42PM -0700, zhilongliu wrote:
> 523 if [ "$MDDEV" = "auto" ]; then
> 524         RAIDDEVS=`list_conf_arrays`
> 525 else
> 526         RAIDDEVS="$MDDEV"
> 527 fi
> 
>     I didn't get the meaning of "$MDDEV" = "auto", and according to the 
> testing, code goes the "RAIDDEVS="$MDDEV"" branch actually, thus here is 
> my question:  what's the auto meaning and does this condition exist 
> defect? I think it's the mainly part of this patch drafting.

From the metadata:

raiddev* (string): block device
    One or more block devices to use, space separated. Alternatively,
    set to "auto" to manage all devices specified in raidconf.

Hope that clears it up.

Thanks.

zhilongliu:
Really appreciate your patience and wonderful guidance, I still 
recommend to do something when this issue happens thereby improve RA 
maintainable.

1. Here I cut the piece of code in resource-agent:heartbeat:Raid1:
# vim ./resource-agents/heartbeat/Raid1
354 raid1_monitor_one() {
... ... ...
363         if [ "$__OCF_ACTION" != "stop" -a ! -b $mddev ]; then
364                 ocf_log info "$mddev is not a block device"       
--> I think this prompt isn't enough, and I added the follow line:
                        echo $mddev | grep -q -o "/dev/" || ocf_log info 
"$mddev lack of the absolute path, \
                        please check crm configure "raiddev" value."
365                 return $OCF_NOT_RUNNING
366         fi

    There is a reason for this behavior, I get used to "# mdadm -A md0  
or # mdadm -S md0 or balabala".
    Suppose "crm configure" encountered one defective value for raiddev, 
at least RA should give the relevant prompt so that locking issue quickly.

In addition:
    Here I cut the raiddev piece of meta_data in "Raid1", it only 
indicated that RA would invoke the raidconf when "raiddev=auto".
  # vim ./resource-agents/heartbeat/Raid1
meta_data() {
         cat <<END
... ... ...
<parameter name="raiddev" unique="0" required="1">
<longdesc lang="en">
One or more block devices to use, space separated. Alternatively,
set to "auto" to manage all devices specified in raidconf.
</longdesc>
<shortdesc lang="en">block device</shortdesc>
<content type="string" default="" />
</parameter>
... ... ...


Thanks very much and Best regards,
-Zhilong

dmuhamedagic:
I believe that we can close this one. Thanks!

Issue #632, opened 2015-06-29
pgsql failed to run in SELinux enforcing mode


hashnao:
Running PostgreSQL (Master/Slave) replicated cluster composed of Red Hat Enter Prise Linux Add-on and ocf:heartbeat:pgsql failed in SELinux enforcing mode.
I followed the instruction at <http://clusterlabs.org/wiki/PgSQL_Replicated_Cluster> to create the postgresql multi-state resource, and the instruction mentions that SELinux should be disabled (the reason is not stated). I confirmed that pgsql resource agent succeeds in running in SELinux only permissive or disabled mode. However, after analyzing the logging, I believe that the reason that the pgsql resource agent failed to run in SELinux enforcing mode is that it uses su command, which failed domain transitions.

hashnao:
I verified the pgsql resource agent in the following environment, and confirmed that the issue is 100% reproducible if it is run under SELinux enforcing mode because systemd-logind process failed to send a D-bus message to a cluster service show in the ausearch command result in the last part.

## OS version
 - OS: RHEL-7.1 (x86_64)
 - Kernel: 3.10.0-229.el7.x86_64
 - HA:
    corosync-2.3.4-4.el7_1.1.x86_64
    pacemaker-1.1.12-22.el7_1.2.x86_64,
    pcs-0.9.137-13.el7_1.2.x86_64
    resource-agents-3.9.5-40.el7
 - DB: postgresql-server-9.2.10-2.el7_1.x86_64

## The result in SELinux enforcing mode

### Create a multi-state resource (ocf:hearbeat:pgsql) and a resource group (ocf:hearbeat:IPaddr2)

* show SELinux mode
```
[root@db01 pcs]# getenforce
Enforcing
```

* configure property, resource defaults policy
```
[root@db01 pcs]# pcs property set no-quorum-policy="ignore"
[root@db01 pcs]# pcs resource defaults resource-stickiness="INFINITY"
[root@db01 pcs]# pcs resource defaults migration-threshold="1"
```

* create resource (vip)
```
[root@db01 pcs]# cib_file=cluster_pgsql.xml
[root@db01 pcs]# pcs -f ${cib_file} resource create vip-master IPaddr2 \
   ip="192.168.100.100" \
   nic="eth0" \
   cidr_netmask="24"

[root@db01 pcs]# pcs -f ${cib_file} resource create vip-rep IPaddr2 \
   ip="192.168.102.100" \
   nic="eth2" \
[...]

PR #652, opened 2015-07-30
Med: tomcat: Use runuser command instead of su to run selinux enforcing mode

hashnao:
It is recommended to use runuser command instead of su to run resources in SELinux enforcing mode because it would generate lots of SELinux errors as shown in the blog <http://danwalsh.livejournal.com/55588.html>. The issue #632 of ocf pgsql resource agent happens to ocf tomcat resource agent because it uses su command.


dmuhamedagic:
Thanks!

PR #654, opened 2015-08-05
send_arp: fix for infiniband, re-merge from upstream iputils arping

lge:
send_arp.linux.c was created to avoid a libnet dependency.
It was Based on arping from iputils as it was back then,
and adapted to the command line conventions established by
the libnet based send_arp by Matt Soffen,
as it was used by IPaddr and IPaddr2.

On infiniband, the link layer addresses won't fit into a sockaddr_ll,
causing a buffer overflow.
The broadcast address may be different from "all 0xff".

This (and some more) has been fixed meanwhile in iputils arping.

This is just a re-merge from iputils, keeping the few adaptations we
need for IPaddr2 command line conventions, and some static/inline/const
fixes to quiescen the compiler warnings.

### DIFF
--- a/tools/send_arp.linux.c
+++ b/tools/send_arp.linux.c
@@ -7,6 +7,23 @@
  *		2 of the License, or (at your option) any later version.
  *
  * Authors:	Alexey Kuznetsov, <kuznet@ms2.inr.ac.ru>
+ * 		YOSHIFUJI Hideaki <yoshfuji@linux-ipv6.org>
+ */
+
+/* Andrew Beekhof, Lars Ellenberg:
+ * Based on arping from iputils,
+ * adapted to the command line conventions established by the libnet based
+ * send_arp tool as used by the IPaddr and IPaddr2 resource agents.
+ * The libnet based send_arp, and its command line argument convention,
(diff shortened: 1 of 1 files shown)


lge:
Meanwhile, if your installed iputils arping is "recent enough" (after 2010 upstream, >= RHEL6, not yet RHEL5), a valid workaround would be:
`text -x /sbin/arping && ln -s arping /sbin/ipoibarping`
because IPaddr2 apparently special-cased infiniband devices for a long time, and prefers to use `ipoibarping`, if available, even with a suitable set of command line arguments.

Also note that IPaddr2 as is will "appear to work" on infiniband, it will just fail to properly send out the unsolicited ARPs, and clients may take a long time to update their neighbour tables.

lge:
We may consider changing send_arp.libnet.c calling conventions to match those of iputils arping instead, drop the `ipoibarping` special case from IPaddr2, change the calling convention in IPaddr2, and possibly add an optional preference on the platform `arping` over our own `send_arp`.
Then deprecate `send_arp` and at some point just depend on iputils being installed and working.

PR #657, opened 2015-08-10
Low: pgsql: fix grep failure when using pacemaker 1.1.12

higashik:
The keywords "FAILED" or "ORPHANED" might be displayed in previous position of the role.
So, I add that keywords to grep pattern.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -979,7 +979,7 @@ pgsql_replication_monitor() {
 
     # I can't get master node name from $OCF_RESKEY_CRM_meta_notify_master_uname on monitor,
     # so I will get master node name using crm_mon -n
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
On Sun, Aug 09, 2015 at 10:52:29PM -0700, Kazuhiko HIGASHI wrote:
> The keywords "FAILED" or "ORPHANED" might be displayed in previous position of the role.
> So, I add that keywords to grep pattern.

I wonder if it would be better to use wildcard instead:

"^${RESOURCE_NAME}[(:].*[):].*Master"

Program output is a moving target and being more inclusive helps
avoid regressions.

higashik:
Thank you for your comment.
I also think the wild card is more safety.
So, I created a new pull request below.
#662

Issue #656, opened 2015-08-07
The findif() function should be changed to calculate the broadcast address from the cidr netmask


krig:
- An interface without any address configured yet:
```
cn2:~ # ip a show dev vlan4
10: vlan4@bond0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP
    link/ether 52:54:00:9f:d3:d5 brd ff:ff:ff:ff:ff:ff
cn2:~ # ip r show dev vlan4
```

- An IPaddr2 resource with the following parameters:
```
OCF_RESKEY_nic="vlan4"
OCF_RESKEY_ip="192.168.4.1"
OCF_RESKEY_cidr_netmask="24"
(OCF_RESKEY_broadcast is NOT set)
```

findif returns an empty broadcast in this case:
```
cn2:~ # . /usr/lib/ocf/lib/heartbeat/findif.sh
cn2:~ # findif
vlan4 netmask 24 broadcast
```

The findif binary works as expected:
```
cn2:~ # /usr/lib64/heartbeat/findif
vlan4   netmask 255.255.255.0   broadcast 192.168.4.255
```

The issue is, that findif reads the broadcast from an existing ip address in this subnet, if there is none, it fails.
This is a regression from the old behaviour and does not work as documented in the IPaddr2 metadata:

```
<parameter name="broadcast">
<longdesc lang="en">
Broadcast address associated with the IP. If left empty, the script will
determine this from the netmask.
</longdesc>
<shortdesc lang="en">Broadcast address</shortdesc>
<content type="string" default=""/>
</parameter>
```

The findif() function should be changed to calculate the broadcast address from the cidr netmask.

Also:

in findif.sh:
```
set -- `ip -o -f $family addr show | grep $7`
```

this is broken, because `$7` is the IP address returned by `ip -o -f $family route list match $match $scope`

problem 1: An IPv4 address contains dots, grep uses regex...

problem 2: if IP address is 10.1.1.1, it will also match 110.1.1.1 as a substring. The grep needs more context.

possible fix:

```
ip -o -f $family addr show | fgrep "$family $7/"
```

kskmori:
> The issue is, that findif reads the broadcast from an existing ip address in this subnet, if there is none, it fails.

Could you elaborate what exactly "fails" means and the problem on your use case? 
[...]

PR #658, opened 2015-08-11
findif.sh: Use ipcalc tool to generate broadcast address from netmask (#656)

krig:


dmuhamedagic:
On Tue, Aug 11, 2015 at 02:56:20AM -0700, Kristoffer Grönlund wrote:
> 
> You can view, comment on, or merge this pull request online at:
> 
>   https://github.com/ClusterLabs/resource-agents/pull/658
> 
> -- Commit Summary --
> 
>   * findif-ipcalc: Add ipcalc tool to be used by findif.sh (#656)

Is there a way to fix this without introducing a new program?
@kskmori, Keisuke-san, you worked on a findif rewrite, any ideas?

krig:
I added the ipcalc tool for performing the broadcast address calculation based on the CIDR netmask. It's not super-complicated, but it's bit-twiddling and I am not confident in my ability to get it right in pure bash. I welcome alternative solutions!

krig:
Sorry, the broadcast address is calculated from the ip and netmask combined: So an address like 192.168.20.4/24 should give the mask 192.168.20.255.

krig:
I pushed an alternative implementation purely in bash, however it is only tested for very simple cases. Also it is adapted from code found here: http://www.linuxquestions.org/questions/programming-9/bash-cidr-calculator-646701/ so the license situation is at best murky. ;)

dmuhamedagic:
#! line is ignored in file which is sourced. Anyway, we cannot
change the shell, as not everybody uses bash.

krig:
Hmm, well, then that idea won't work either since the code needs bash features like exponentiation and arithmetic for loops.

kskmori:
> Is there a way to fix this without introducing a new program? @kskmori, Keisuke-san, you worked on a findif rewrite, any ideas?

I've commented on here:
https://github.com/ClusterLabs/resource-agents/issues/656#issuecomment-135323802

PR #660, opened 2015-08-12
Added "maxfiles" parameter

xtavras:
Added "maxfiles" parameter for increasing maximum number of open files (ulimit -n) for slapd process, useful for many LDAP connections.

### DIFF
--- a/heartbeat/slapd
+++ b/heartbeat/slapd
@@ -54,7 +54,7 @@
 : ${OCF_RESKEY_password=""}
 : ${OCF_RESKEY_parameters=""}
 : ${OCF_RESKEY_stop_escalate=15}
-: ${OCF_RESKEY_maxfiles=4096}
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
This introduces a regression by setting the default maxfiles
limit to 4096. I don't know how serious is that, but just to be
on the safe side I suppose that it would be better to remove the
default and set the limit _only_ if the maxfiles parameter is
set.

Also, the indentation is slightly different from the one already
used in the RA.

xtavras:
Hi @dmuhamedagic, slapd on Debian already has 4096 as hard limit, but  it's a good point and I agree with you, I've removed default value and fixed some syntax.

dmuhamedagic:
Yes, well, the ulimit could be set differently by other means.

Thanks for the contribution!

dmuhamedagic:
Fixed meta-data, rebased, pushed. Closing.

PR #662, opened 2015-08-20
Low: pgsql: fix grep failure when using pacemaker 1.1.12

higashik:
The keywords "FAILED" or "ORPHANED" might be displayed in previous position of the role.
And the program output is a moving target , any keywords might be displayed, so I add wild card to grep pattern.


kskmori:
The code looks good to me too, but the commit log message seems concatenated to a single very long line when looking from the git CLI commands.

@higashik Sorry for bothering you, but could you amend the commit log and insert a blank line to separate between the subject and the body? Just doing push --force is fine and I'll merge it.

ref.: https://www.kernel.org/pub/software/scm/git/docs/git-commit.html#_discussion

Thanks!

Issue #655, opened 2015-08-06
Is it mistake or some workaround in set_master_score?


m-ly4:
I found, that update of pgsql master attribute called for node $target (pgsql resource agent)
Is it mistake and $1 should be used or it is some sort of w/a?

```bash
set_master_score()
{
    local current_score

    current_score=`$CRM_ATTR_REBOOT -N "$1" -n "master-$3" -G -q 2>/dev/null`
    if [ -n "$current_score" -a "$current_score" != "$2" ]; then
        ocf_log info "Changing $3 master score on $1 : $current_score->$2."
>>>     exec_with_retry 0 $CRM_ATTR_REBOOT -N "$target" -n "master-$3" -v "$2"

    fi
    return 0
}
```

dmuhamedagic:
Looks like a bug to me too. @t-matsuo, Takatoshi-san, what is
your opinion?

higashik:
Hi, I am a colleague of @t-matsuo .
(Takatoshi-san was transferred to another company.)

I think this is Bug too.
But, the "target" variable has been defined in all of the calling function.
So, it was operating normally.

I would like to fix it.

higashik:
I create a new pull request #663 .
Would you please check it and merge?

t-matsuo:
It's a bug !!
Thanks for your report and fixing it.

dmuhamedagic:
Pull request merged.

PR #663, opened 2015-08-20
Low: pgsql: set_master_score use an undefined variable($target).

higashik:
This is the fix of https://github.com/ClusterLabs/resource-agents/issues/655 .
Would you please check it and merge?


dmuhamedagic:
Thanks!

PR #667, opened 2015-08-25
pgsql: Fix return code override in pgsql_real_start()

sobolev-es:
Commit d67c0ae (pgsql: Support replication slots) overrides return code of pgsql_real_monitor() in pgsql_real_start() function when replication_slot_name is defined.
So pgsql_real_start() never returns $OCF_RUNNING_MASTER and pgsql_promote() failed with error "Can't start PostgreSQL as primary on promote." on cluster startup.


dmuhamedagic:
Thanks!

PR #672, opened 2015-09-03
High: pgsql. Fix get_my_location sql regression

asp24:


dmuhamedagic:
Indeed this seems to have been introduced by accident in ac9cee5c0ce5cbc1f88304eea9f20bf470bd3b40

@kskmori, Keisuke-san, could you please check this one too. Thanks!

kskmori:
Yes, the fix correct.
Thank you for pointing out!

PR #674, opened 2015-09-03
Medium: pgsql. Fix exec_sql errors like "unknown variable select pg_ " on ubuntu

asp24:


ytakeshita:
I think "$1" is better than $1.
It appears to depend on dash rather than ubuntu.
without "" by using dash on RHEL, the error had occured. 
In the case of bash, the error regardless of the presence or absence of the double quotes does not occur.

dmuhamedagic:
Thanks!

PR #675, opened 2015-09-10
ldirectord.in: dns_check and fallbackcommand enhancements

MorphBonehunter:
- support for custom dns ports
- dns check results are now regex ready
- fallback cmd arguments added

### DIFF
--- a/ldirectord/ldirectord.in
+++ b/ldirectord/ldirectord.in
@@ -576,6 +576,8 @@ overridden by an optional per real-server based receive regexp.
 
 For a DNS check this should be any one the A record's addresses or
 any one of the PTR record's names.
+In case of dynamic DNS answers (different answers on the same question)
+a regex to match multiple addresses or PTR record names could also defined.
 
 For a MySQL check, the receive setting is not used.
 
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
@horms, could you take a look at this one please. Thanks!

MorphBonehunter:
We have rebased on clusterlabs master and append 2 commits with documentation and explanation of the functionality we have added.
With the rebase the comments from @horms got lost (first time i did such an thing on github, so maybe i did this wrong), sorry.

MorphBonehunter:
I "restore" the lost comments.
We "fix" only the DNS Check as we have an use case for this, we do not check other protocols so we can not make a statement how many other protocol checks don't honor the port configuration.

MorphBonehunter:
With the hint from @horms we check all the other "check_" subs and find all honor the port configuration.
With this task we discover that there is an sub for the port check which is unsed by the other checks, so we adjust our commit to unify the code.

horms:
@MorphBonehunter: Thanks. I'm happy with this series now.
@dmuhamedagic: please feel free to apply

dmuhamedagic:
Thanks!

PR #677, opened 2015-09-11
findif.sh: Add + and - broadcast arguments (bsc#940931) (#656)

krig:
Remove inaccurate statement about not passing a broadcast argument, and add description of `+` and `-` arguments for broadcast address.

These are passed to `ip address` to calculate the broadcast address based on the prefix.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -169,8 +169,10 @@ routing table.
 
 <parameter name="broadcast">
 <longdesc lang="en">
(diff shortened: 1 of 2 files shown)


krig:
ping @kskmori @dmuhamedagic

kskmori:
It looks great to me. Thank you for the improvements!

dmuhamedagic:
Sounds good to me too.

krig:
Thank you both!

PR #681, opened 2015-09-16
Low: s/ist/is typo fix

jnpkrn:


dmuhamedagic:
"ist" is "is" in German :)

jnpkrn:
Ich weiß :)

PR #683, opened 2015-10-01
Medium: SAPDatabase: Fix process detection if DIR_EXECUTABLE is set

krig:
(bsc#947243)

SAPDatabase uses pgrep and pkill to discover running SAP processes,
but if those processes were not started with the same absolute path
to the executable, pgrep/pkill won't match if the absolute path is
used when searching.

Since DIR_EXECUTABLE only ever changes the path and not the executable
name, it should be safe to use the executable name only and not the
absolute path when calling pgrep/pkill.

### DIFF
--- a/heartbeat/sapdb.sh
+++ b/heartbeat/sapdb.sh
@@ -44,11 +44,11 @@ background_check_saphostexec() {
 #                       misbehavior
 #
 cleanup_saphostexec() {
-  pkill -9 -f "$SAPHOSTEXEC"
-  pkill -9 -f "$SAPHOSTSRV"
-  oscolpid=`pgrep -f "$SAPHOSTOSCOL"`       # we check saposcol pid, because it
-                                            # might not run under control of
-					    # saphostexec
+  pkill -9 -f saphostexec
+  pkill -9 -f sapstartsrv
+  oscolpid=`pgrep -f saposcol`       # we check saposcol pid, because it
+                                     # might not run under control of
+                                     # saphostexec
 
   # cleanup saposcol shared memory, otherwise it will not start again
   if [ -n "$oscolpid" ];then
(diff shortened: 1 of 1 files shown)


krig:
ping @AlexanderKrauth

heartbeat/sapdb.sh
```diff
@@ -44,11 +44,11 @@ background_check_saphostexec() {
 #                       misbehavior
 #
 cleanup_saphostexec() {
-  pkill -9 -f "$SAPHOSTEXEC"
-  pkill -9 -f "$SAPHOSTSRV"
-  oscolpid=`pgrep -f "$SAPHOSTOSCOL"`       # we check saposcol pid, because it
-                                            # might not run under control of
-					    # saphostexec
+  pkill -9 -f saphostexec
+  pkill -9 -f sapstartsrv
```
AlexanderKrauth: Hi,
I got your point in general. But this change will cause new trouble.
The sapstartsrv executable may run multible times on a host. But not all instances must be related to this resource instance. E.g. they can also be related to a SAPInstance resource. Killing them all maybe causes damage to other resources.
Therefore it is necessary to identify the correct instance.
Do you have more details about the issue where it occured ?
IMO this can only happen, when the processes get startet manually on a console. Which should not be the case in a cluster environment.
Cheers
Alex

krig: Ah, I see.. I didn't realise that multiple instances could be active at once.

The process in particular that was the issue in this case was `saphostexec`. From looking at the resource agent, it looks to me like it expects the process to potentially be started by someone else. This was indeed the case that we encountered, only that the command line used to start it differed. As a result the SAPDatabase agent would stop the running instance and try to launch a new copy.

But, it is entirely possible that it is a question of misconfiguration / administrator error rather than something that really should be fixed.

krig:
Going to close this since it seems to be more of a configuration issue than a problem with the agent itself. Thanks!

PR #687, opened 2015-10-15
Fixed attempts to create iqn multiple times for every portal.

akeyjadu:
When multiple portals are provided using "portals=" switch
the old code will attempt to create IQN for every portal. This is not
correct. There should be just one IQN and multiple portals to it.
targetcli will return error if attempting to create IQN that already exists.


akeyjadu:
Yay!

PR #688, opened 2015-10-15
fix "error: setup_container: Resource mylxc: Unknown resource"

kirillrst:
Tested just on Ubuntu 14.04
Details: http://clusterlabs.org/pipermail/users/2015-April/000246.html

### DIFF
--- a/heartbeat/lxc
+++ b/heartbeat/lxc
@@ -30,7 +30,7 @@
 # Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
 
 # OCF instance parameters
-#       OCF_RESKEY_container
+#       OCF_RESKEY_containername
 #       OCF_RESKEY_config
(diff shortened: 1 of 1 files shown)


oalbrigt:
I dont get that error in Fedora 22, where it's still using OCF_RESKEY_container, so I guess it might've been fixed in a later version of resource-agents/pacemaker.

dmuhamedagic:
On the one hand, it seems like for some versions of pacemaker having a parameter "container" is a no go, right? Do we know which ones?
On the other hand, lxc was in at least one resource-agents release with the "container" parameter and normally it's forbidden to modify parameter names (otherwise we break existing configurations).
I think that we should close this and keep the "container" parameter if that works with new pacemaker.

krig:
Closing this based on reasoning in previous comment.

Issue #576, opened 2015-02-17
Please help improve RA meta data


lge:
See also my ML post,  2015-02-12 [Call for review of undocumented parameters in resource agent meta data](http://thread.gmane.org/gmane.linux.highavailability.user/40607/focus=7947) and the followup by Dejan.

> I just today noticed that apparently some resource agents
> accept and use parameters that are not documented in their meta data.
> 
> I now came up with a bash two-liner,
> which likely still produces a lot of noise,
> because it does not take into account that some agents
> "source" additional helper files.

> But here is the list:
> 
> --- used, but not described

This is bad and needs to be fixed.

> +++ described, but apparently not used.

Just drop?

- [ ] EvmsSCC           +OCF_RESKEY_ignore_deprecation
- [ ] Evmsd             +OCF_RESKEY_ignore_deprecation

	?? intentionally undocumented ??

No idea, but I doubt that anybody out there is using evms.

- [ ] IPaddr            +OCF_RESKEY_iflabel

According to the history, this was never used.

- [ ] IPaddr            -OCF_RESKEY_netmask
> 	Not sure.

This got renamed to cidr_netmask, in an effort to make it more
consistent with IPaddr2 :) The same as what you found below.

- [ ] IPaddr2           -OCF_RESKEY_netmask
> 
> 	intentional, backward compat, quoting the agent:
>         # Note: We had a version out there for a while which used
>         # netmask instead of cidr_netmask. Don't remove this aliasing code!

> Please help review these:
 
- [ ] IPsrcaddr         -OCF_RESKEY_ip
- [ ] IPsrcaddr         +OCF_RESKEY_cidr_netmask
- [ ] IPv6addr.c        -OCF_RESKEY_cidr_netmask
- [ ] IPv6addr.c        -OCF_RESKEY_ipv6addr
- [ ] IPv6addr.c        -OCF_RESKEY_nic
- [ ] LinuxSCSI         +OCF_RESKEY_ignore_deprecation
- [ ] Squid             -OCF_RESKEY_squid_confirm_trialcount
- [ ] Squid             -OCF_RESKEY_squid_opts
- [ ] Squid             -OCF_RESKEY_squid_suspend_trialcount
- [ ] SysInfo           -OCF_RESKEY_clone
- [ ] WAS6              -OCF_RESKEY_profileName

- [ ] apache            +OCF_RESKEY_use_ipv6

This is used in http-mon.sh, sourced by apache.

[...]

PR #691, opened 2015-10-23
Low: pgsql: properly show default values in meta-data

kskmori:
fixes two issues:
 - xlog_check_count: default value was not shown in the meta-data
 - ctl_opt: default value was not initialized properly
            (although it's been working perfectly as expected)

ref. https://github.com/ClusterLabs/resource-agents/issues/576#issuecomment-90826628


kskmori:
I believe that the fix is pretty obvious so I'm going to merge this.

PR #694, opened 2015-10-23
IPaddr2: Add IPv6 DAD collision detection

oalbrigt:


dmuhamedagic:
It seems like that there's a number of bashisms in the new code.
Note that IPaddr2 is /bin/sh, just like quite a few of other
ocf:heartbeat RA. There's the checkbashisms script which can
help.

krig:
There's a travis job that runs shellsheck already checked in, someone with admin rights just need to turn it on :) http://docs.travis-ci.com/user/getting-started/

kskmori:
This patch seems supersede the wait loop in run_send_ua() introduced by #181 .

I would appreciate if you could also make sure this and remove the duplicate loop to keep the code cleanly.
Thank you!

dmuhamedagic:
Keisuke-san, did you review the new code? Thanks!

kskmori:
Yes, the new code looks good. Thank you for that!

I would prefer the patches being squashed into a single commit before merging for simplifying  the commit history, but it does not matter.

dmuhamedagic:
Thanks everybody!

PR #697, opened 2015-11-11
Add NovaCompute RA from https://github.com/aspiers/osp-ha-deploy/commit/c46d87be3d99474a860eefbdae569d1af17e42f5

oalbrigt:


dmuhamedagic:
It is preferable that the RA is maintained where people have expertise of software which is managed by that RA.

Any particular reason we would want to move this agent to resource-agents?

fabbione:
@dmuhamedagic you are right, I think there was a miscommunication with @oalbrigt. this should have gone to the openstack-resource-agent project, but I know @aspiers was taking care of it.

PR #698, opened 2015-11-12
When stop processing is carried out when mysql fell while monitor does not detect an error, the cluster fails in a stop of mysql.

HideoYamauchi:
Hi All,

I send a patch to improve the problem of the stop that I reported by an email
 * http://clusterlabs.org/pipermail/users/2015-November/001773.html

The patch includes the next change.
 * Commonization of the function
 * Restraint of the output of the redundant deletion message of the PID file

Best Regards,
Hideo Yamauchi.


dmuhamedagic:
Thanks!

PR #701, opened 2015-11-18
IPaddr2: potential syntax error on if-then-else

guessi:
what we need here should be `A && (B || C)`

please refer to the result of following simple tests

before patch:
```
  true  && true  || true;  echo $?;  # return value: 0
  true  && true  || false; echo $?;  # return value: 0
  true  && false || true;  echo $?;  # return value: 0
  true  && false || false; echo $?;  # return value: 1
  false && true  || true;  echo $?;  # return value: 0
  false && true  || false; echo $?;  # return value: 1
  false && false || true;  echo $?;  # return value: 0
  false && false || false; echo $?;  # return value: 1
```
after patch:
```
  true  && (true  || true);  echo $? # return value: 0
  true  && (true  || false); echo $? # return value: 0
  true  && (false || true);  echo $? # return value: 0
  true  && (false || false); echo $? # return value: 1
  false && (true  || true);  echo $? # return value: 1
  false && (true  || false); echo $? # return value: 1
  false && (false || true);  echo $? # return value: 1
  false && (false || false); echo $? # return value: 1
```

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -800,7 +800,7 @@ ip_start() {
 		exit $OCF_SUCCESS
 	fi
 	
-	if [ -n "$IP_CIP" ] && [ $ip_status = "no" ] || [ $ip_status = "partial2" ]; then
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Looks good. Thanks!

PR #702, opened 2015-11-20
nfsserver: Add EXEC_MODE for systemd without nfs-lock.service

krig:
On SUSE systems, we do not have a nfs-lock.service service script,
but we use nfs-server.service to manage NFS using systemd.

This patch adds a new EXEC_MODE which uses nfs-server.service to manage
the NFS server, but calls rpc.statd to handle NFSv3 locking.


dmuhamedagic:
Looks good. Thanks!

PR #703, opened 2015-11-23
Mid : Generation and error check of the container name at the time of the globally-unique designation.

HideoYamauchi:
The patch converts a container name at the time of the globally-unique
designation into a form supported in docker.
When docker RA cannot generate the name at the time of globally-unique
designation automatically, the patch returns an error.

If there is the method that is better for handling of sed of the patch,
I want you to change it.

* Some problems still remain for the use with the clone of docker RA.
* This is a temporary correction, but this correction blocks some
problem movement.

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -417,7 +417,7 @@ if ocf_is_true "$OCF_RESKEY_CRM_meta_globally_unique"; then
 			exit $OCF_ERR_CONFIGURED
 		fi
 	fi
-	: ${OCF_RESKEY_name=`echo ${OCF_RESOURCE_INSTANCE} | sed -e 's/\(.*\):\(.*\)$/\1-\2/'`} 
+	: ${OCF_RESKEY_name=`echo ${OCF_RESOURCE_INSTANCE} | tr ':' '-'`} 
 else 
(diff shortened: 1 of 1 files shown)


HideoYamauchi:
Hi Dejan,

Okay!

I intend to adopt your opinion to a patch.
I wait for the opinion of other people a little more.

Many Thanks!
Hideo Yamauchi.

HideoYamauchi:
Hi Dejan, 

I switched to tr command.
How about with this patch?

Best Regards,
Hideo Yamauchi.

dmuhamedagic:
Hi Hideo-san,

On Thu, Dec 03, 2015 at 03:32:38PM -0800, Hideo Yamauchi wrote:
> Hi Dejan, 
> 
> I switched to tr command.
> How about with this patch?

Looks good to me. Could you please just rebase so that we have
only one patch in the pull request. Thanks!

HideoYamauchi:
Hi Dejan,

I closed this and sent a new request again.
　
*  https://github.com/ClusterLabs/resource-agents/pull/714

Many Thanks!
Hideo Yamauchi.

PR #705, opened 2015-11-24
nfsserver: Use rpc-statd.service for NFS locking in EXEC_MODE=3

krig:
The previous patch to fix this issue was insufficient. From what I can tell, we need to use the `rpc-statd.service` in the same way as `nfs-lock.service` is used in `EXEC_MODE=2`. `nfs-lock.service` is simply a renamed `rpc-statd.service` from NFS upstream.


dmuhamedagic:
OK. Hope that there won't be any more missing services.

PR #706, opened 2015-11-26
Nagios Resource Agent

oalbrigt:
Cleaned up version of: https://github.com/ClusterLabs/resource-agents/pull/636

Any suggestions for improvements are welcome.

### DIFF
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -94,6 +94,7 @@ ocf_SCRIPTS	     =  ClusterMon		\
 			ManageVE		\
 			mysql			\
 			mysql-proxy		\
+			nagios			\
 			named			\
 			nfsnotify		\
 			nfsserver		\
--- a/heartbeat/nagios
+++ b/heartbeat/nagios
@@ -1,0 +1,250 @@
+#!/bin/sh
+#
+#  License:      GNU General Public License (GPL)
+#  (c) 2015 T.J. Yang, O. Albrigtsen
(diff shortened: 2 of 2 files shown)


oalbrigt:
Thank you for your feedback. I've tried to fix all the issues you mentioned.

oalbrigt:
The "command" file can be generated by other applications with commands that Nagios will run periodically:
https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/extcommands.html

dmuhamedagic:
On Wed, Dec 02, 2015 at 06:16:01AM -0800, oalbrigt wrote:
> The "command" file can be generated by other applications with commands that Nagios will run periodically:
> https://assets.nagios.com/downloads/nagioscore/docs/nagioscore/3/en/extcommands.html

But how do we know that it's going to be there when we need it
and, in particular, when we remove it after every start?

oalbrigt:
I guess it's only used for commands that needs to be run while it's running like to trigger a reload or similar features, as it's removed during start in the init-script in RHEL7 that I used as a reference when going through the code.

dmuhamedagic:
Looks fine to me.
Did you also test using ocf-tester?
Care to rebase to a single patch?

oalbrigt:
It passed the ocf-tester when I commented out the ocf_is_root part, which made it fail due to the permissions test being done as nobody.

dmuhamedagic:
OK. Thanks!

PR #709, opened 2015-12-02
implement reload action (for normal processing node list changes)

kulaginm:
Implement reload action for pgsql resource agent.

Without this action this command
pcs resource update pgsql pgsql node_list="node1 node2 node3 ..."
will restart all postgres servers in cluster

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -7,6 +7,7 @@
 #               Florian Haas (florian@linbit.com) -- makeover
 #               Takatoshi MATSUO (matsuo.tak@gmail.com) -- support replication
 #               David Corlette (dcorlette@netiq.com) -- add support for non-standard library locations and non-standard port
+#               Mikhail A. Kulagin (m.kulagin@postgrespro.ru) -- add reload action
 #
 # Copyright:    2006-2012 Serge Dubrouski <sergeyfd@gmail.com>
 #                         and other Linux-HA contributors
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
@kskmori Could some of the pgsql experts take a look here. Thanks!

kskmori:
I'm afraid that the patch should be rejected, because it does always
reloading even if you changed the parameters that actually requires
restarting. (Ex. start_opt etc.)

If my understanding is correct, the current OCF spec. does not provide a
way to indicate which parameters are reload safe and which are not, and
Pacemaker will try to reload for any parameters with unique=0 if a
reload operation is implemented in an RA.
pgsql RA has a lot of parameters with unique=0 that requires restarting.


As a side note, the commit https://github.com/ClusterLabs/pacemaker/commit/2022d089 
is necessary (i.e. coming 1.1.14 or later) in order to work the
reload operation for pgsql, because it does not have any unique=1 parameters.
(I noticed that pgdata should be unique=1 at least, but it's an another issue)


If you only want to change node_list without restarting, then probably
we should reconsider how to obtain node_list so that we do not have to
change the parameter when the cluster membership has changed.

Currently node_list needs to be specified a fixed list manually because
we just could not find out a better way to do this other than that, so
the improvements are welcome.

Thank you.

heartbeat/pgsql
```diff
@@ -861,6 +865,61 @@ pgsql_stop() {
...
+                rc=1
+            else
+                ocf_log debug "empty recovery.conf"
+                rc=0
+            fi
+        fi
+        ocf_log debug "removing temporary file"
+        runasowner "rm -f ${tmp_file}"
+        if [ $rc -eq 0 ]; then
+            ocf_log debug "recovery.conf is unchanged -> reload is enough"
```
kskmori: I do not understand the intention here: reloading the config. when the config. is unchanged??
'reload_conf' is supposed to be used to reflect the changes in recovery.conf.

kskmori:
Closing as declined. Please let me know if I'm missing something. Or another enhancements are welcome. 
Thanks!

bubble75:
As a note: if you return ERR_GENERIC for the reload action, pacemaker will restart the resource. That gives much more flexibility, but, of course, reload failure will appear in CIB status.

PR #710, opened 2015-12-02
Filesystem: wait for glusterfs up

guessi:


dmuhamedagic:
Looks good. May I suggest to move the code to a separate
function. Perhaps create one generic wait function and put the
gluster specific code into another, maybe something like this:

wait_for_backend() {
	local backend_f timeout

	backend_f=$1
	timeout=${OCF_RESKEY_CRM_meta_timeout:-15000}
	while [ $timeout -gt 0 ]; do
		$backend_f && return 0
		timeout=$((timeout - 1))
		sleep 1
	done
	return 1
}
gluster_backend_test() {
	gluster volume status
}

if [ "$FSTYPE" = "glusterfs" ]; then
	if ! wait_for_backend gluster_backend_test; then
		ocf_exit_reason "Backend service [$FSTYPE] not ready"
		return $OCF_ERR_GENERIC
	fi
fi

guessi:
Umm...
your advice would be better, but there're several types of filesystem need
to take care of. Including exceptions, of course.

how about to wrap "mount" in a time-limited retry loop?

for ... {
   try_mount $FSTYPE
   count++
}

how do you think?

dmuhamedagic:
On Fri, Dec 04, 2015 at 06:54:29AM -0800, guessi, mei wrote:
> Umm...
> your advice would be better, but there're several types of filesystem need
> to take care of. Including exceptions, of course.
> 
> how about to wrap "mount" in a time-limited retry loop?
> 
> for ... {
>    try_mount $FSTYPE
>    count++
> }
> 
> how do you think?

Maybe, but how do we discern ultimate from transient errors?

guessi:
no, we can't take care of all unexpected conditions,

btw, I will create another PR later

guessi:
@dmuhamedagic please have a look at #711 , thanks !!!

guessi:
Closing, replaced by #711

PR #711, opened 2015-12-05
Low: Filesystem: check filesystem backend before mount

guessi:


dmuhamedagic:
In general, the idea is good, but the implementation needs to be refined.

guessi:
@dmuhamedagic 

I've update the code, please spare some time to review it again,

If squash required, please let me know, thanks !!!

guessi:
@dmuhamedagic 
rebased & simplified, check backend support only

please review it again, thanks!

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+# Determine backend filesystem support state
+#
+backend_support()
+{
+	case "$FSYPE" in
+		overlay)
+			# it should fail immediately if kernel module not found
+			modinfo overlay >/dev/null 2>&1
+			;;
+		overlay)
```
krig: Unless I'm mistaken, the second "overlay" case would never match. Is this supposed to be "overlayfs"?

guessi: Oops... that's a stupid typo, nice catch, thanks!

guessi:
@krig update!

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+			modinfo overlayfs >/dev/null 2>&1
+			;;
+		nfs4|nfs|smbfs|cifs|gfs2|ocfs2)
+			# FIXME: how to determine if these service is available?
+			;;
+		*)
+			# any other?
+			;;
+	esac
+	[ $? -eq 0 ] && return 0
```
krig: I'm not sure what result `$?` will contain here for anything other than overlay/overlayfs, since nothing was executed in that case.

Maybe it would be better to check and return in each case? Something like

```
overlay)
    modinfo overlay >/dev/null 2>&1 && return 0
    ;;
```

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+			ceph)
+				# it should failed if ceph not proper configure
+				ceph health >/dev/null 2>&1
+				;;
+			*)
+				# any other?
+				;;
+		esac
+		[ $? -eq 0 ] && return 0
+
```
krig: Same as above

heartbeat/Filesystem
guessi: umm, another bug here,
when it run into ` *) ` case, ` $? ` will be 0

will be fixed in next commit

guessi:
@krig updated, but still leave several ` FIXME: ` out there,

need more reviewers.

guessi:
@dmuhamedagic @krig rebas & refreshed, please review again when you are free, thanks!

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+				# $FSTYPE not support.
+				pass=1
+				;;
+		esac
+
+		# exit immediately after test passed
+		[ $pass -eq 0 ] && break
+
+		# retry again until timeout
+		timeout=$((timeout - 1))
```
dmuhamedagic: This will unnecessary loop for non-transient error conditions.

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+
+#
+# Mount the filesystem.
+#
+try_mount()
+{
+	# determine for backend support
+	backend_support $FSTYPE
+	if [ $? -ne 0 ]; then
+		ocf_exit_reason "Backend [$FSTYPE] not support or not available"
```
dmuhamedagic: It would be better to capture the real exit reason in the backend_support function.

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+				# it should fail immediately if kernel module not found
+				lsmod | grep -q "$FSTYPE" && pass=0
+				;;
+			nfs4|nfs|smbfs|cifs|gfs2|ocfs2)
+				# assume all these type were ready when start
+				pass=0
+				;;
+			*)
+				# $FSTYPE not support.
+				pass=1
```
dmuhamedagic: Perhaps set "pass" here to 0. The generic problem with this, not really coming up with this patch, is that with support for various filesystems spread through the code, we're bound to forget to update all places.

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
 	return $OCF_ERR_INSTALLED
 }
 
+#
+# Determine backend filesystem support state
+#
+backend_support()
+{
+	# network backended service type need retry timeout (default: 15)
+	# please note the timeout value here mean "15 times", not "15 secs"
```
guessi: I'll make change it to OCF_KESKEY_max_retry_count at next commit

heartbeat/Filesystem
```diff
@@ -389,6 +389,82 @@ fstype_supported()
...
+	# 0: success 1: failed (default)
+	local pass=1
+
+	while [ $timeout -gt 0 ]; do
+		case "$FSTYPE" in
+			glusterfs)
+				# it should failed if service not ready
+				# here we set a general timeout value for waiting response
+				# since gluster will have slow response when starting up
+				timeout 5 gluster volume status >/dev/null 2>&1 && pass=0
```
guessi: error message from GlusterFS is not always helpful, setting a timed out value here is to avoid waiting too long

dmuhamedagic:
On Sat, Nov 05, 2016 at 08:33:26PM -0700, guessi wrote:
> error message from GlusterFS is not always helpful

That's a rather vague qualification. Wouldn't it still be prudent
to log the message in case of error?

> setting a
> timed out value here is to avoid waiting too long

Such ad-hoc timeout values are not always advisable and in
particular so if it's for a network service. It typically depends
on the actual environment. The timeout should perhaps be some
fraction of the resource start timeout value. I don't have
experience with gluster and don't really know what kind of
timeout/fraction is realistic.

guessi:
rebased

and make it possible to set time out value for detecting Gluster/Ceph status before mount.

guessi:
now it's rebase onto v4.0.0rc1

PR #713, opened 2015-12-07
High: mysql: fix grep failure on MySQL5.6 or higher when checking read_only variable

higashik:
On MySQL 5.6 or higher, the "SHOW VARIABLES" command returns the multiple rows that contains the "read_only".(innodb_read_only, read_only and tx_read_only)
So, I add "-w" to grep option.


dmuhamedagic:
OK. Thanks.

PR #714, opened 2015-12-07
Mid : Generation and error check of the container name at the time of the globally-unique designation.(repost)

HideoYamauchi:
The patch converts a container name at the time of the globally-unique
designation into a form supported in docker.
When docker RA cannot generate the name at the time of globally-unique
designation automatically, the patch returns an error.

If there is the method that is better for handling of sed of the patch,
I want you to change it.

Some problems still remain for the use with the clone of docker RA.
This is a temporary correction, but this correction blocks some problem
movement.
Best Regards,
Hideo Yamauchi.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -403,7 +403,27 @@ docker_validate()
 	return $OCF_SUCCESS
 }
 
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Thanks.

PR #715, opened 2015-12-07
Low: mysql: fix missing valiables

higashik:
the OCF_RESKEY_replication_user_default and OCF_RESKEY_replication_passwd_default
are used, but not defined.
So, I define those valiables in mysql-common.sh.


higashik:
Sorry for spell miss.("valiable" is wrong. )
I'll fix it.

dmuhamedagic:
On Mon, Dec 07, 2015 at 01:51:20AM -0800, Kazuhiko HIGASHI wrote:
> Sorry for spell miss.("valiable" is wrong. )
> I'll fix it.

Hi Kazuhiko-san,

You can also fix the existing pull request by amending commits
and then doing "git push -f".

PR #716, opened 2015-12-07
Low: mysql: fix missing variables

higashik:
The OCF_RESKEY_replication_user_default and OCF_RESKEY_replication_passwd_default
are used, but not defined.
So, I define those variables in mysql-common.sh.


dmuhamedagic:
Thanks.

PR #718, opened 2015-12-11
/var/lock/subsys is non-standard, check for it first

kmcfate:
/var/lock/subsys does not exist on debian/ubuntu. Primarily a RedHat-ism. Check for it first.


dmuhamedagic:
No harm with this, but is there more that needs to be done for Debian?

kmcfate:
Yes more needs to be done, primarily around 'nfs-common' and rpc.idmapd. I'll have another pull coming soon.

PR #727, opened 2015-12-22
nfsserver localconfig

dmuhamedagic:
The /etc/sysconfig/nfs editing code introduced in 2014 is
    specific to RH based linux. At best it has no effect on other
    distributions and at worst it can break the RA operation and/or
    nfs configuration.
    
    There is rudimentary support for distribution detection in
    ocf-distro. It should serve well for this purpose.
    
    NB: ocf-distro is intended to be used just here and just for this
    purpose and not to be further developed.
    
    Distribution-specific code is _not_ encouraged.


dmuhamedagic:
Tested (with ocft) on SLE12SP1. Please test on redhat based platforms.

tools/ocft/nfsserver
```diff
@@ -6,8 +6,15 @@ CONFIG
 	InstallPackage nfs-kernel-server
```
oalbrigt: This package is called nfs-utils on RHEL.

oalbrigt:
Works fine on RHEL7.2 (after changing the InstallPackage in the ocft config to nfs-utils as noted above).

tools/ocft/nfsserver
dmuhamedagic: On Wed, Dec 23, 2015 at 01:26:24AM -0800, oalbrigt wrote:
> > @@ -6,8 +6,15 @@ CONFIG
> >  	InstallPackage nfs-kernel-server
> 
> This package is called nfs-utils on RHEL.

Heh. InstallPackage should support redhat, suse, debian, but what
good is it when the package names differ. Anyway, it turned out
we can do some initialization in SETUP-AGENT. There's now a
new commit here which should make ocft nfsserver work on redhat
too.

dmuhamedagic:
On Wed, Dec 23, 2015 at 01:28:44AM -0800, oalbrigt wrote:
> Works fine on RHEL7.2 (after changing the InstallPackage in the ocft config to nfs-utils as noted above).

Great! Thanks for testing.

PR #730, opened 2015-12-30
Ocft update

dmuhamedagic:
Improve some ocft test suites.
Fix some ocft issues.
Add runocft to run a set of test suites.


dmuhamedagic:
This set of updates makes ocft pass on SLE12SP1, actually runocft. My guess is that the test suites would also run fine on other platforms but for package name differences. It would be important to improve the multiplatform support in ocft.

PR #733, opened 2016-01-04
Dev: VirtualDomain: switch to ocf-rarun to reduce complexity

dmuhamedagic:
There where several places testing how to properly exit depending on the action. That's handled properly by ocf-rarun.


dmuhamedagic:
Tested with ocft on SLE12.

Issue #734, opened 2016-01-04
Uncontrolled OCF monitor recursive calls when importing ocf-shellfuncs


bogdando:
Details are given here http://clusterlabs.org/pipermail/users/2016-January/002050.html
The dummy OCF resource https://github.com/bogdando/dummy-ocf-ra/blob/master/dummy
"works" as expected if I don't import ocf-shellfuncs at all. Otherwise, it causes nested monitor calls to grow ending up the node might turn completely unresponsive.

The work around is provided here https://github.com/bogdando/dummy-ocf-ra/blob/5f5cffe9b8b6c648bf92f78a4b1049254682b0a4/dummy#L7-L49

kskmori:
seems like ocf_start_trace() hitting a bug in dash: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=692611

kskmori:
@bogdando : Would you try this patch against to /usr/lib/ocf/lib/heartbeat/ocf-shellfuncs ? It would enable the bash specific trace info only when it's running under bash.
```
--- ocf-shellfuncs.org	2015-06-02 11:38:37.000000000 +0900
+++ ocf-shellfuncs	2016-01-05 16:06:35.489441829 +0900
@@ -820,7 +820,9 @@
 			return
 		ocf_trace_redirect_to_file "$__OCF_TRC_DEST"
 	fi
-	PS4='+ `date +"%T"`: ${FUNCNAME[0]:+${FUNCNAME[0]}:}${LINENO}: '
+	if [ -n "$BASH_VERSION" ]; then
+		PS4='+ `date +"%T"`: ${FUNCNAME[0]:+${FUNCNAME[0]}:}${LINENO}: '
+	fi
 	set -x
 	env=$( echo; printenv | sort )
 }
```

bogdando:
will do, thanks

bogdando:
For my version of the resource-agents the patch was a slightly different by line numbers, but it looks like have fixed the issue.

kskmori:
Thanks for testing! I've submitted the patch as a pull request and will be fix in the upstream.

jnpkrn:
See also http://oss.clusterlabs.org/pipermail/users/2016-January/002059.html

dmuhamedagic:
Fixed by Keisuke-san. Thanks!

PR #735, opened 2016-01-06
Medium: ocf_shellfuncs: suppress bash specific trace_ra log on dash

kskmori:
Fixes the issue: https://github.com/ClusterLabs/resource-agents/issues/734

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -902,7 +902,9 @@ ocf_start_trace() {
 			return
 		ocf_trace_redirect_to_file "$__OCF_TRC_DEST"
 	fi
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
OK. I suppose that this doesn't impair the tracing functionality significantly. Once that dash bug gets fixed, we can drop this patch again.

dmuhamedagic:
s/again//

kskmori:
Thanks for your ack and merging.

As a memorandum for when we get it back to dash:
* FUNCNAME is a bash only feature and causes Bad substitution error on dash
* LINENO is not yet supported and always empty, at least on dash-0.5.7-4ubuntu1/Ubuntu 14.04

Issue #692, opened 2015-10-23
Invalid log message when stop postgreSql by pgSql RA


m-ly4:
pgsql RA - pgsql_real_stop() function:
In log message value of ${OCF_RESKEY_stop_escalate} is used:
```bash
    if pgsql_status
    then
        #PostgreSQL is still up. Use another shutdown mode.
        ocf_log info "PostgreSQL failed to stop after ${OCF_RESKEY_stop_escalate}s using -m fast. Trying -m immediate..."
        runasowner "$OCF_RESKEY_pgctl -D $OCF_RESKEY_pgdata stop -m immediate"
    fi
```

But above initiated local variable stop_escalate, which can have another value for slave agent:
```bash
    stop_escalate=$OCF_RESKEY_stop_escalate
    if [ "$1" = "slave" ]; then
        stop_escalate="$OCF_RESKEY_stop_escalate_in_slave"
    fi
```
```

kskmori:
Thanks for reporting.
I and my colleague will look into this issue.

kskmori:
@m-ly4 : Please find that it has been fixed as the PR #736 . Thanks!
Closing.

PR #736, opened 2016-01-06
Low: pgsql : fix invalid log message

ytakeshita:
Fixes the issue: #692


kskmori:
Thanks for fixing it! Merging.

Issue #670, opened 2015-09-02
pgsql RA replication slots support - invalid chars in slot name


ilyas-ppg:
PostrgreSQL has some limitations in replication slot name:

postgres=# select pg_create_physical_replication_slot('a-b');
ERROR:  replication slot name "a-b" contains invalid character
HINT:  Replication slot names may only contain letters, numbers, and the underscore character.

Because hostname often contains symbol '-' resource agent in this case cannot create physical replication slot due to above error.

May be good idea here is replace '-' to underscore. This replacement will work here and now without changes in postgresql source code.

Probably this fix will help to avoid described problem:

$ diff -Naur pgsql.orig pgsql
```
--- pgsql.orig	2015-09-02 10:39:04.336886107 +0300
+++ pgsql	2015-09-02 11:00:42.556844908 +0300
@@ -1323,7 +1323,7 @@
         for target in $NODE_LIST
         do
             if [ "$target" != "$NODENAME" ]; then
-                replication_slot_name="$OCF_RESKEY_replication_slot_name"_"$target"
+                replication_slot_name="$OCF_RESKEY_replication_slot_name"_"${target//-/_}"
                 replication_slot_name_list_tmp="$replication_slot_name_list"
                 replication_slot_name_list="$replication_slot_name_list_tmp $replication_slot_name"
             fi
```
Sorry for bashism.

dmuhamedagic:
@kskmori Keisuke-san, what do you say? I suppose that this is a good suggestion.

kskmori:
Yes, this is indeed a good suggestion. I and my colleague @ytakeshita will take a look into this.
Probably we should also consider a period '.' character in  a hostname as well.

dmuhamedagic:
Fixed. Thanks!

PR #737, opened 2016-01-07
Low: pgsql: fix to be used valid charcters when the invalid characters is used in replication_slot_name.

ytakeshita:
Fixes the issue: #670


dmuhamedagic:
Thanks!

PR #741, opened 2016-01-14
Medium: nfsserver: fix monitor for systemd

dmuhamedagic:


ytakeshita:
Thanks for fixing!
I test your code, but `nfsserver_monitor()` and `systemctl is-active` doesn't detect the failure of nfsd process because `systemctl is-active` returns `active` and the return value is `0` as well.
````
# systemctl start nfs-server.service
# systemctl is-active nfs-server.service
active
#
# ps ax | grep nfsd
29744 ?        S<     0:00 [nfsd4]
29745 ?        S<     0:00 [nfsd4_callbacks]
29749 ?        S      0:00 [nfsd]
29750 ?        S      0:00 [nfsd]
29751 ?        S      0:00 [nfsd]
29752 ?        S      0:00 [nfsd]
29753 ?        S      0:00 [nfsd]
29754 ?        S      0:00 [nfsd]
29755 ?        S      0:00 [nfsd]
29756 ?        S      0:00 [nfsd]
29802 pts/0    S+     0:00 grep --color=auto nfsd
#
# pkill -9 nfsd
#
# systemctl is-active nfs-server.service
active
# echo $?
0
# ps ax | grep nfsd
29806 pts/0    S+     0:00 grep --color=auto nfsd
````

ytakeshita:
I think that it should not use systemctl for monitoring the status of kernel process like nfsd.
Maybe the systemctl status does't monitor the status of kernel process.

In case of using /etc/init.d/nfs, nfsserver_monitor may execute `pidof nfs nfsd` via init script.
So, I made a sample patch with using pidof.What about?

https://github.com/ytakeshita/resource-agents/commit/b8bfd15e0fa16e4a76580d22128d48d5e10c2b6e

I only tested on RHEL 7.1. sorry.

oalbrigt:
Fixed in #746.

PR #742, opened 2016-01-19
VirtualDomain: scope is not deleted in case of domain failure.

mgrzybek:
# Domain failure

In case of crash, the container's systemd unit is not removed by libvirt.
We need to delete it to be able to restart. We should not find any scope of the former domain.
# Whitespaces cleanup

Some whitespaces were left at the end of some lines.

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -171,7 +171,7 @@ Be sure to set the timeout of these operations to accommodate this delay.
 
 <parameter name="autoset_utilization_cpu" unique="0" required="0">
 <longdesc lang="en">
-If set true, the agent will detect the number of domainU's vCPUs from virsh, and put it 
+If set true, the agent will detect the number of domainU's vCPUs from virsh, and put it
 into the CPU utilization of the resource when the monitor is executed.
 </longdesc>
 <shortdesc lang="en">Enable auto-setting the CPU utilization of the resource</shortdesc>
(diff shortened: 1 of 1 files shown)


heartbeat/VirtualDomain
```diff
@@ -482,6 +493,11 @@ VirtualDomain_start() {
 	# is restored to an 'undefined' state before creating.
 	verify_undefined
 
+	# Make sure domain's systemd scope is deleted.
+	# In case of domain failure, libvirt does not delete them and cannot
+	# restart it.
+	verify_systemd_units
```
guessi: not all distro come with systemd support by default

heartbeat/VirtualDomain
```diff
@@ -427,6 +427,17 @@ VirtualDomain_status() {
 	return $rc
 }
 
+# In case of crash, the container's systemd unit is not removed by libvirt.
+# We need to delete it to be able to restart. We should not find any scope
+# of the former domain.
+verify_systemd_units() {
+	systemctl --no-legend -t scope | grep "$DOMAIN_NAME" 2>&1 > /dev/null
```
guessi: should be `>/dev/null 2>&1` or `... | grep -q "xxxx"`, right?

krig:
Hi, can you please move the whitespace cleanup into a separate commit?

Also, I wonder why you need this. The failed state is cleared when the service is restarted. Manually calling reset-failed should not be necessary.

mgrzybek:
I used to get the issue using KVM on CentOS / RHEL 7. I have not re-tested it yet.

PR #746, opened 2016-01-28
Medium: nfsserver: fix monitor for systemd

ytakeshita:
instead of PR #741

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -293,10 +293,42 @@ v3locking_exec()
 	fi
 }
 
+nfsserver_systemd_monitor()
+{
+	local threads_num
+	local rc
+
+	nfs_exec is-active
+	rc=$?
+
+	# Now systemctl is-active can't detect the failure of kernel process like nfsd.
+	# So, if the return value of systemctl is-active is 0, check the threads number
+	# to make sure the process is running really.
+	# /proc/fs/nfsd/threads has the numbers of the nfsd threads.
+	if [ $rc -eq 0 ]; then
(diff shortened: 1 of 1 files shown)


krig:
I checked on a SUSE system, and using `/proc/fs/nfsd/threads` should work there as well.

@dmuhamedagic, looks good to you too?

oalbrigt:
Line 307: $rc is not set

And the return codes needs to be updated so it doesnt fail in monitor() when the resource is disabled/not started on the node.

ytakeshita:
> Line 307: $rc is not set

Sorry, it is my mistake. Thank you for pointing out !
fixes in below.
https://github.com/ytakeshita/resource-agents/commit/e7b6c3fbe30e4e8b06d412c402373eb8cd2d9bc3

> And the return codes needs to be updated so it doesnt fail in monitor() when the resource is disabled/not started on the node.

When $rc is not set ? I think if $rc is set correctly, monitor() detect failure in such cases.

oalbrigt:
I tested with setting rc=$? and it still failed.

Also nfs_exec is-active wont work with SysV systems.

FYI. If you commit to fix_nfs_monitor_systemd your fixes will be added here instead of having to link to them.

oalbrigt:
It works fine now. I have no idea why it failed when I changed it myself.

Can you commit the update to this branch and squash it to a single commit?

https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit

ytakeshita:
Thanks you for checking and merging !

PR #748, opened 2016-02-02
galera: use mysql's --tc-heuristic-recover if crash recovery is needed

dciabrin:
If a Galera node is unexpectedly killed in a middle of a replication, InnoDB can retain the equivalent of a XA transaction in "prepared" state in its redo log. If so, mysqld cannot recover state (e.g. last seqno) automatically, and a special recovery flag has to be used on command line to unblock the node:

160129 14:47:27  InnoDB: Starting recovery for XA transactions...
160129 14:47:27  InnoDB: Transaction 7900 in prepared state after recovery
160129 14:47:27  InnoDB: Transaction contains changes to 1 rows
160129 14:47:27  InnoDB: 1 transactions in prepared state after recovery
160129 14:47:27 [Note] Found 1 prepared transaction(s) in InnoDB
160129 14:47:27 [ERROR] Found 1 prepared transactions! It means that mysqld was not shut down properly last time and critical recovery information (last binlog or tc.log file) was manually deleted after a crash. You have to start mysqld with --tc-heuristic-recover switch to commit or rollback pending transactions.
160129 14:47:27 [ERROR] Aborting

This pull request allows the galera resource agent to configure recovery heuristic to unblock a node if needed and to request a full state transfer (SST) a next restart for extra safety.

The pull request also enables to agent to keep track of recoveries to prevent bootstrapping a cluster from a recovered node when possible.

### DIFF
--- a/heartbeat/README.galera
+++ b/heartbeat/README.galera
@@ -25,7 +25,7 @@ restart a Galera cluster.
 
 ### Bootstrap the cluster with the right node
 
-When synced, the nodes of a galera clusters have in common a last seqno,
+When synced, the nodes of a galera cluster have in common a last seqno,
 which identifies the last transaction considered successful by a
 majority of nodes in the cluster (think quorum).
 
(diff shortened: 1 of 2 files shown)


beekhof:
Looks pretty sane to me

PR #751, opened 2016-02-05
v3.9.7 build failure

kskmori:
v3.9.7 fails to build the rpm by using the included configure scritps. Apparently runocft script is missing in the tarball which is introduced by #730.

Steps to reproduce:
```
[root@build-centos71 resource-agents (master)]# git checkout v3.9.7
[root@build-centos71 resource-agents ((v3.9.7))]# ./autogen.sh && ./configure && make rpm
(omit)
Making all in ocft
gmake[4]: Entering directory `/root/rpmbuild/devel/resource-agents/resource-agents-3.9.6.272-e697f/tools/ocft'
gmake[4]: *** No rule to make target `runocft', needed by `all-am'.  Stop.
gmake[4]: Leaving directory `/root/rpmbuild/devel/resource-agents/resource-agents-3.9.6.272-e697f/tools/ocft'
gmake[3]: *** [all-recursive] Error 1
(omit)
```

### DIFF
--- a/tools/ocft/Makefile.am
+++ b/tools/ocft/Makefile.am
@@ -17,7 +17,7 @@
 
 MAINTAINERCLEANFILES = Makefile.in
 
-EXTRA_DIST		= $(ocftcfgs_DATA) $(ocft_DATA)
+EXTRA_DIST		= $(ocftcfgs_DATA) $(ocft_DATA) $(ocft_SCRIPTS)
 
 sbin_SCRIPTS		= ocft
 


dmuhamedagic:
Ah, bad me.  Thanks, Keisuke-san!

dmuhamedagic:
But strange thing is that I'm quite sure I did many builds with OBS (OBS uses rpmbuild) and never saw a build failure. Is this only with make rpm?

kskmori:
It should be affected if the tarball was created by make dist. Maybe OBS archives it from the git repository directly?

dmuhamedagic:
On Wed, Feb 10, 2016 at 12:55:49AM -0800, Keisuke MORI wrote:
> It should be affected if the tarball was created by make dist. Maybe OBS archives it from the git repository directly?

Yes, right, the tarball is usually provided by 'git archive', but it's actually up to the user how to do that.

Now that I think about it, I never use 'make dist', always 'git archive' or 'hg archive'.

I wonder if this is a ground for a fix release?

PR #756, opened 2016-02-09
ldirectord: Don't send Inaccessible real server e-mails in fork=yes mode

mcnewton:
When starting up, if a service is not listed in the ipvsadm table the parent process will mark it as 'down'. Each child process will then go through its own services and mark them as 'up'.

However, if there is more than one service defined, the services from the other children are still marked as 'down'.  In this case each child will repeatedly e-mail out 'Inaccessible real server' messages for services now managed by the other children.

To prevent this, we go through the e-mail status hash as a child starts up and remove any services that now belong to other children.


mcnewton:
Thanks!

PR #759, opened 2016-02-10
mysql-common.sh: send SIGKILL to ppid, too

kleind:
The agent starts mysqld_safe, mysql's own wrapper around mysqld, which apparently is the best/safest way to start mysqld. The pid written to the pidfile is the pid of mysqld, the spawned child process.

If mysqld fails to stop within $timeout minus 5 seconds, the agent sends SIGKILL to the pid.

Unfortunately, mysqld_safe restarts mysqld when it is killed with SIGKILL (opposed to also dying when stopped by SIGTERM).

scenario 1
mysql running

root@dodo:~# ps -ef|grep mysql
root     19922     1  0 09:07 ?        00:00:00 /bin/sh /usr/bin/mysqld_safe
mysql    20178 19922  0 09:07 ?        00:00:00 /usr/sbin/mysqld --basedir=/usr --datadir=/var/lib/mysql --user=mysql --pid-file=/var/run/mysqld/mysqld.pid --socket=/var/run/mysqld/mysqld.sock --port=3306
root     20179 19922  0 09:07 ?        00:00:00 logger -t mysqld -p daemon.error
root     22769 22744  0 09:24 pts/1    00:00:00 grep mysql

send SIGTERM

root@dodo:~# kill 20178
root@dodo:~# ps -ef|grep mysql
root     22794 22744  0 09:24 pts/1    00:00:00 grep mysql

both processes ended. good.

scenario 2
mysql running

root@dodo:~# /etc/init.d/mysql start
Starting MySQL (Percona Server): mysqld.
Checking for corrupt, not cleanly closed and upgrade needing tables..
root@dodo:~# ps -ef|grep mysql
root     22826     1  0 09:24 pts/1    00:00:00 /bin/sh /usr/bin/mysqld_safe
mysql    22950 22826  3 09:24 pts/1    00:00:00 /usr/sbin/mysqld --basedir=/usr --datadir=/var/lib/mysql --user=mysql --pid-file=/var/run/mysqld/mysqld.pid --socket=/var/run/mysqld/mysqld.sock --port=3306
root     22951 22826  0 09:24 pts/1    00:00:00 logger -t mysqld -p daemon.error
root     23057 22744  0 09:24 pts/1    00:00:00 grep mysql

send SIGKILL 

root@dodo:~# kill -9 22950

mysqld_safe restarted mysqld

root@dodo:~# ps -ef|grep mysql
root     22826     1  0 09:24 pts/1    00:00:00 /bin/sh /usr/bin/mysqld_safe
mysql    23084 22826  3 09:25 pts/1    00:00:00 /usr/sbin/mysqld --basedir=/usr --datadir=/var/lib/mysql --user=mysql --pid-file=/var/run/mysqld/mysqld.pid --socket=/var/run/mysqld/mysqld.sock --port=3306
root     23085 22826  0 09:25 pts/1    00:00:00 logger -t mysqld -p daemon.error
root     23110 22744  0 09:25 pts/1    00:00:00 grep mysql

If we encounter a situation in which we need to send SIGKILL to mysqld, we should make sure it is actually stopped before reporting OCF_SUCCESS.

### DIFF
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -296,8 +296,17 @@ mysql_common_stop()
 
     mysql_common_status info $pid
     if [ $? != $OCF_NOT_RUNNING ]; then
-        ocf_log info "MySQL failed to stop after ${shutdown_timeout}s using SIGTERM. Trying SIGKILL..."
-        /bin/kill -KILL $pid > /dev/null
+        # parent (mysqld_safe) will respawn child (mysqld) if we only kill the child with -KILL
+        ppid=`ps -o ppid= $pid`
+        ocf_log info "MySQL failed to stop after ${shutdown_timeout}s using SIGTERM. Sending SIGKILL to parent $ppid and child $pid."
+        /bin/kill -KILL $ppid $pid > /dev/null
+    fi
+
+    # Double check if it is really dead now
+    mysql_common_status info
+    if [ $? = $OCF_SUCCESS ]; then
+      ocf_log err "MySQL still running after we sent SIGKILL. Nothing more we can do here."
+      return $OCF_ERR_GENERIC
     fi
 
     ocf_log info "MySQL stopped";


kleind:
I confirmed this mysqld_safe behaviour for mysql 5.1, 5.5, 5.6 and 5.7.

lge:
On Wed, Feb 10, 2016 at 12:48:10AM -0800, kleind wrote:
> The agent starts mysqld_safe, mysql's own wrapper around mysqld, which
> apparently is the best/safest way to start mysqld. The pid written to
> the pidfile is the pid of mysqld, the spawned child process.

Then, when started by pacemaker,
it should start mysql directly,
and skip the wrapper?

Or, given bad timing of some monitoring action,
mysql might die, mysql_safe might restart it,
during recovery the pacemaker monitor action might fail,
triggering a stop/start by pacemaker,
and afaicr mysql is extremely unhappy when killed during recovery.

   Lars

kleind:
I'm not sure skipping the wrapper is what we want. There might be things in those 650-ish lines of code that would need to be implemented in the resource agent if we don't want to introduce other regressions in behaviour.

smurfix:
Why do you use mysqld_safe in the first place?
The thing makes sense when you don't have a reasonable service manager. As soon as you do, it's superfluous, if not dangerous.

kleind:
Well ... If one is not supposed to use it in this environment, then I guess that information simply did not make it into my head :(

Since I was not aware that people might actually not use mysqld_safe, the patch has to be changed. If mysqld_safe is not used, this patch might kill init, which I guess would lead to fencing, which might not be necessary.

But ... if what you say is true ... maybe we have a bad default in the agent and that should be fixed as well? Are there any regression effects you know of not using mysqld_safe?

smurfix:
mysqld_safe basically does this:

```usage () {
        cat <<EOF
Usage: $0 [OPTIONS]
  --no-defaults              Don't read the system defaults file
  --core-file-size=LIMIT     Limit core files to the specified size
  --defaults-file=FILE       Use the specified defaults file
  --defaults-extra-file=FILE Also use defaults from the specified file
  --ledir=DIRECTORY          Look for mysqld in the specified directory
  --open-files-limit=LIMIT   Limit the number of open files
  --crash-script=FILE        Script to call when mysqld crashes
  --timezone=TZ              Set the system timezone
  --malloc-lib=LIB           Preload shared library LIB if available
  --mysqld=FILE              Use the specified file as mysqld
  --mysqld-version=VERSION   Use "mysqld-VERSION" as mysqld
  --nice=NICE                Set the scheduling priority of mysqld
  --no-auto-restart          Exit after starting mysqld
  --nowatch                  Exit after starting mysqld
  --plugin-dir=DIR           Plugins are under DIR or DIR/VERSION, if
                             VERSION is given
  --skip-kill-mysqld         Don't try to kill stray mysqld processes
  --syslog                   Log messages to syslog with 'logger'
  --skip-syslog              Log messages to error log (default)
  --syslog-tag=TAG           Pass -t "mysqld-TAG" to 'logger'
  --flush-caches             Flush and purge buffers/caches before
                             starting the server
  --numa-interleave          Run mysqld with its memory interleaved
                             on all NUMA nodes

All other options are passed to the mysqld program.

EOF
```
Besides that, it does a bunch of prefix and path handling if you have mysql installed in a non-standard location or with a non-standard data directory, to simplify startup -- and try to minimize the chance of running two mysql daemons on the same data, which would be Very Bad. It also tries to use "mysqladmin shutdown" if you send a SIGTERM to it.

If your startup handler already supports everything of the above which you actually need, then using mysqld_safe is essentially a no-op.

dmuhamedagic:
Not an expert on mysql, but I can vaguelly recall some
discussions on this matter. Found one here:

http://lists.linux-ha.org/pipermail/linux-ha/2011-November/044485.html

Couldn't find any previous discussions though.

> Besides that, it does a bunch of prefix and path handling if you have mysql installed in a non-standard location or with a non-standard data directory, to simplify startup -- and try to minimize the chance of running two mysql daemons on the same data, which would be Very Bad. It also tries to use "mysqladmin shutdown" if you send a SIGTERM to it.

The "prefix and path handling" sounds to me like it would be a
regression for some installations to change the default.

PR #760, opened 2016-02-10
ldirectord: Correctly suppress startup e-mails when emailalertstatus=running

mcnewton:
When the daemon is starting up it sets DAEMON_STATUS to STARTING, aligns its internal tables with those of ipvsadm (in ld_start), then sets DAEMON_STATUS to RUNNING.

This means that any services which are down (which is usually the case when ldirectord is not running, unless cleanstop=no is set) will be marked as down in STARTING state.

Then when the main loop in ld_main is run, the services will all be detected as 'up', but as we are now in RUNNING state, so e-mails will be sent out. This isn't really a running state, though - we've only just started up.

To prevent this from happening, do a one off initial check of all real servers in the ld_start, which means that services will be detected as 'up' when in DAEMON_STATUS STARTING. From then on the ld_main (or the children in fork=yes mode) will continue to check as usual and send e-mails out if applicable in RUNNING state.

In the other case, where emailalertstatus=starting, similarly now the startup e-mails will be sent, whereas previously e-mails would be sent to say the service is down, but not any to say that the servers were actually up and running.

### DIFF
--- a/ldirectord/ldirectord.in
+++ b/ldirectord/ldirectord.in
@@ -2662,24 +2662,8 @@ sub ld_main
 			check_signal();
 
 		} else {
-			my @real_checked;
-			foreach my $v (@VIRTUAL) {
-				my $real = $$v{real};
-				my $virtual_id = get_virtual_id_str($v);
+			_ld_main_check_all();
 
-				REAL: foreach my $r (@$real) {
-					my $real_id = get_real_id_str($r, $v);
-					check_signal();
-					foreach my $tmp_id (@real_checked) {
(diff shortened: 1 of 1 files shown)


krig:
This change look reasonable to me, and apologies for being troublesome, but could you tag each commit message with `ldirectord:`? With that change, I'll merge.

mcnewton:
@krig no problem! Hope it's OK now?

krig:
Great, thanks!

mcnewton:
Thanks!

PR #761, opened 2016-02-11
Low: SAPInstance: Update version support statement

krig:
Update stated SAP WebAS version support.

Also adds a clarification as to what the support statement means.


krig:
@dmuhamedagic If this looks good to you now, please merge :)

PR #762, opened 2016-02-11
galera: fix monitoring of joining node for long running SST

dciabrin:
mysqld's pidfile is only created after SST finished.
RA has to track mysqld process differently at startup to allow
monitoring long SST during "monitor" op.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -343,26 +343,39 @@ check_sync_needed()
     ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-sync-needed" -Q 2>/dev/null
 }
 
+
+# this function is called when attribute sync-needed is set in the CIB
 check_sync_status()
 {
-    local state=$(get_status_variable "wsrep_local_state")
-    local ready=$(get_status_variable "wsrep_ready")
+    # if the pidfile is created, mysqld is up and running
+    # an IST might still be in progress, check wsrep status
+    if [ -e $OCF_RESKEY_pid ]; then
+        local cluster_status=$(get_status_variable "wsrep_cluster_status")
+        local state=$(get_status_variable "wsrep_local_state")
+        local ready=$(get_status_variable "wsrep_ready")
 
-    if [ -z "$state" -o -z "$ready" ]; then
-        ocf_exit_reason "Unable to retrieve state transfer status, verify check_user '$OCF_RESKEY_check_user' has permissions to view status"
-        return $OCF_ERR_GENERIC
+        if [ -z "$cluster_status" -o -z "$state" -o -z "$ready" ]; then
+            ocf_exit_reason "Unable to retrieve state transfer status, verify check_user '$OCF_RESKEY_check_user' has permissions to view status"
+            return $OCF_ERR_GENERIC
+        fi
+
+        if [ "$cluster_status" != "Primary" ]; then
+            ocf_exit_reason "local node <${NODENAME}> is started, but not in primary mode. Unknown state."
+            return $OCF_ERR_GENERIC
+        fi
+
+        if [ "$state" = "4" -a "$ready" = "ON" ]; then
+            ocf_log info "local node synced with the cluster"
+            # when sync is finished, we are ready to switch to Master
+            clear_sync_needed
+            set_master_score
+            return $OCF_SUCCESS
+        fi
     fi
 
-    if [ "$state" = "4" -a "$ready" = "ON" ]; then
-        ocf_log info "local node synced with the cluster"
-        # when sync is finished, we are ready to switch to Master
-        clear_sync_needed
-        set_master_score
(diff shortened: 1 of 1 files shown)


heartbeat/galera
```diff
@@ -488,6 +501,65 @@ detect_first_master()
     set_bootstrap_node $best_node
 }
 
+galera_status()
```
dciabrin: mysqld may in fact be running (e.g. during SST) before it creates its pidfile. This new function adds check for that specific case.
It's implemented in the spirit of https://github.com/ClusterLabs/resource-agents/blob/master/rgmanager/src/resources/mysql.sh#L157

heartbeat/galera
```diff
@@ -516,39 +588,53 @@ galera_start_local_node()
...
 
-    if ocf_is_true $bootstrap; then
         clear_bootstrap_node
         # clear attribute heuristic-recovered. if last shutdown was
         # not clean, we cannot be extra-cautious by requesting a SST
         # since this is the bootstrap node
         clear_heuristic_recovered
     else
+        # only start server, defer full checks to "monitor" op
+        galera_start_nowait "$extra_opts"
```
dciabrin: We can't wait until the pidfile is created because long SST can exceed start time out. So we just start mysqld_safe here and ensure that it has spawned mysqld as expected.

heartbeat/galera
```diff
@@ -749,18 +837,21 @@ galera_monitor()
         return $OCF_ERR_GENERIC
     fi
 
-    is_primary
+    check_sync_needed
```
dciabrin: We can't poll any mysql property until mysqld finishes its start up, i.e. when pidfile is created.
So we switch the order of check_sync_status is_primary to avoid polling server before SST finishes.

zzzeek:
so does this address an SST of arbitrary duration?  e.g. does it finally fix the problem we had w/ aetna where rsync took like ten minutes?

dciabrin:
Affirmative, I can track 10min SST with this pull request.
The "monitor" op polls the mysqld process at regular intervals to ensure the SST is still progressing.
(SST failure is detected by galera itself if various rsync time outs are reached).
the "promote" op is triggered after SST finishes and wsrep status reads "Synced".

heartbeat/galera
```diff
@@ -488,6 +501,65 @@ detect_first_master()
...
+    local loglevel=$1
+    local rc
+
+    if [ -e $OCF_RESKEY_pid ]; then
+        mysql_common_status $loglevel
+        rc=$?
+    else
+        # if pidfile is not created, the server may
+        # still be starting up, e.g. running SST
+        ps auxww | grep -v -e "${OCF_RESKEY_binary}" -e grep | grep -qe "--pid-file=$OCF_RESKEY_pid"
```
zzzeek: so, we've given a solution to people when their galera was started by systemd, to copy /var/run/mariadb.pid to /var/run/mysqld.pid,which wouldn't be caught by this new test.  what do we lose if we just grep for "mysqld" directly, rather than excluding "mysqld_safe" etc. ?

heartbeat/galera
```diff
@@ -488,6 +501,65 @@ detect_first_master()
...
+    # Spin waiting for the server to be spawned.
+    # Let the CRM/LRM time us out if required.
+    start_wait=1
+    while [ $start_wait = 1 ]; do
+        if ! ps $pid > /dev/null 2>&1; then
+            wait $pid
+            ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?), please check your installation"
+            return $OCF_ERR_GENERIC
+        fi
+        ps auxww | grep -v -e "${OCF_RESKEY_binary}" -e grep | grep -qe "--pid-file=$OCF_RESKEY_pid"
```
zzzeek: maybe we want to make a single function that does the "grep for mysqld" feature?   this is a repeated line of code otherwise.

dciabrin: fixed

heartbeat/galera
dciabrin: The narrowing to the explicit "pidfile" is a fairly good guarantee that I'm monitoring the mysqld process I'm interested in (because the RA won't start a mysqld if the pidfile existed beforehand).
If we only monitor mysqld, we makes the start process more fragile: we may miss the original pid failure (e.g if two different mysql are legitimately running on the same machine) and keep the node in Slave state forever.

Note that the handy workaround of "copying pidfile" still work in the same extent as before:
  . one can still copy the pidfile while pacemaker resource is unmanaged, then cleanup, then manage. The monitor op will detect galera as running, as before.
  . in the old agent, if a slave node didn't finish sync, there is no pidfile to copy. So the strict grep doesn't restrain the use case. 

Last point, since the boot process now depends on a attribute "sync-needed" in the CIB, making the "copy pidfile" workaround works also during the sync would require additional manual steps. So weakining the grep for that didn't seem justified IMHO.

zzzeek:
awrite!

PR #766, opened 2016-02-16
Added iSER enabled portal support

d-zalewski:
Added ability specify IP of a portal which will have iSCSI Extensions for RDMA (iSER) feature enabled


krig:
Looks good to me

PR #767, opened 2016-02-19
Low: apache: ensure readable $TESTCONFFILE

guessi:
make sure target file existence and readable, but not just the string test

### DIFF
--- a/heartbeat/apache
+++ b/heartbeat/apache
@@ -594,9 +594,24 @@ apache_validate_all() {
 		return $OCF_ERR_INSTALLED
 	fi
 
-	if [ ! -f "$TESTCONFFILE" ] || [ ! -r "$TESTCONFFILE" ]; then
-		ocf_exit_reason "Configuration file $TESTCONFFILE not found, or not readable."
-		return $OCF_ERR_INSTALLED
+	# validate testconffile/testurl before apache_monitor_10()
+	if [ -n "$TESTCONFFILE" ]; then
+		if [ ! -f "$TESTCONFFILE" ] || [ ! -r "$TESTCONFFILE" ]; then
+			ocf_exit_reason "Configuration file $TESTCONFFILE not found, or not readable."
+			return $OCF_ERR_INSTALLED
+		fi
+	else
+		if [ -n "$TESTURL" ]; then
+			# remove leading or trailing spaces/tabs
+			local temp=$(printf "$TESTURL" | sed -e 's/^[ \t]*//g' -e 's/[ \t]*$//g')
+
+			if [ -z "$temp" ]; then
+				ocf_exit_reason "testurl: \"$TESTURL\" seems to be an empty string?"
+				return $OCF_ERR_CONFIGURED
+			fi
+		fi
+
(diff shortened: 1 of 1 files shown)


krig:
Hmm. Without this change, if `testconffile` is configured but does not exist, the monitor action will fail trying to read the non-existing file. But with this change, if `testconffile` is configured but does not exist, the monitor action will not even try to use the configured test file but will instead try to use `testurl` and `testregex10`.

Neither of these seem like exactly what you'd want.

Shouldn't the existence of `testconffile` be checked in the `validate-all` action instead?

guessi:
@krig update!

krig:
Won't this look for $TESTCONFFILE even if `testcconffile` hasn't been configured? That will probably fail then.

guessi:
according to `apache_getconfig()`,
`$TESTCONFFILE`  would be an empty string, if `OCF_RESKEY_testconffile` not set

if `$TESTCONFFILE` is an empty string,
` [ ! -f $TESTCONFFILE ] ` will failed, but not ` [ ! -f "$TESTCONFFILE" ] `

**note**, it's double-quoted variable

does it need to change the first line from `#!/bin/sh` to `#!/bin/bash` ?

krig:
A problem is that the user shouldn't have to configure `testconffile` since they can use `testregex10` instead. But with this change, the resource agent will report an error if `testconffile` is not a readable file.

I would first test if TESTCONFFILE is non-zero:

```
if [ -n "$TESTCONFFILE" ]; then
   # check if $TESTCONFFILE is a readable file in here
else
   # check testurl / testregex10 here
fi
```

Don't change `#!/bin/sh` to `#!/bin/bash`, the agents should be POSIX compatible as much as possible.

I can also recommend using `http://shellcheck.net` to check scripts, it catches many bashisms.

dmuhamedagic:
On Mon, Feb 22, 2016 at 10:52:42PM -0800, Kristoffer Grönlund wrote:
> Don't change `#!/bin/sh` to `#!/bin/bash`, the agents should be POSIX compatible as much as possible.

More, if an agent is /bin/sh, it has to remain /bin/sh.
Otherwise, we may introduce regressions for installations that
have only /bin/sh installed. Shells are not interchangeable.

guessi:
@krig @dmuhamedagic updated, with a simple validation for `$TESTURL` only,
but i leave `$TESTREGEX10` not check, to allow user get response from `$TESTURL` only

reference code:
[apache_monitor_10](https://github.com/ClusterLabs/resource-agents/blob/v3.9.7/heartbeat/apache#L317)

guessi:
as for the `shellcheck` topic,
I always run `shellcheck` locally, but for some historical reason, tons of `SCxxxx` out there !!!

`SC2086: Double quote to prevent globbing and word splitting`
`SC2006: Use $(..) instead of legacy ...`
`SC2003: expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]]`

no harm if we don't touch them, so I leave them not touch.

krig:
Yes, there are plenty of things to clean up in the agents, but it is probably best to leave them for now. :)

Thank you, I think this looks good now. Merging.

PR #773, opened 2016-02-24
quote vgchange options

dhoffend:
Without the options being you end up with the following error message during resource startup

    Feb 24 22:14:47 cluster-a lrmd[1717]:   notice: vg0_start_0:2344:stderr [ /usr/lib/ocf/resource.d/heartbeat/LVM: 415: local: --monitor: bad variable name ]

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -386,7 +386,7 @@ get_activate_options()
 ##
 retry_exclusive_start()
 {
-	local vgchange_options=$(get_activate_options)
+	local vgchange_options="$(get_activate_options)"
 
 	# Deactivate each LV in the group one by one cluster wide
 	set -- $(lvs -o name,attr --noheadings $OCF_RESKEY_volgrpname 2> /dev/null)
@@ -412,7 +412,7 @@ retry_exclusive_start()
 #	Enable LVM volume
 #
 LVM_start() {
-	local vgchange_options=$(get_activate_options)
+	local vgchange_options="$(get_activate_options)"
 	local vg=$1
 	local clvmd=0
 


krig:
Hmm, no, I don't think that is the real issue. `get_activate_options` potentially returns multiple options, quoting the return value would mean passing all of them as a single argument to vgchange.

I'm wondering if it could be a bash vs. dash problem, though.

krig:
I suspect that this is the same issue:

https://bugs.launchpad.net/ubuntu/+source/dash/+bug/139097

Splitting the local declaration and the assignment to separate lines should work around it:

```
local vgchange_options
vgchange_options=$(get_activate_options)
```

krig:
Actually, I think I am at least partially wrong - I think your fix would work as well. Apologies for the noise :)

(I still think that difference in how local is parsed is what the error message indicates)

dhoffend:
Yeah i it looks like the dash issue you linked. As I'm playing around with cLVM atm I can confirm that the fix above at least works:

I've added a bit of debug/echo code to show the result:
```
root@cluster-a:~# grep -A3 "LVM_start()" /usr/lib/ocf/resource.d/heartbeat/LVM
LVM_start() {
        local vgchange_options="$(get_activate_options)"
        echo "LVM_start vgchange_options = $vgchange_options"
        local vg=$1

root@cluster-a:~# /usr/lib/ocf/resource.d/heartbeat/LVM start
LVM_start vgchange_options = -aly --monitor y
INFO: Activating volume group vg0
INFO: Reading all physical volumes. This may take a while... Found volume group "vg0" using metadata type lvm2
INFO: 1 logical volume(s) in volume group "vg0" now active
```

dmuhamedagic:
On Wed, Feb 24, 2016 at 03:01:53PM -0800, Kristoffer Grönlund wrote:
> I suspect that this is the same issue:
> 
> https://bugs.launchpad.net/ubuntu/+source/dash/+bug/139097
> 
> Splitting the local declaration and the assignment to separate lines should work around it:
> 
> ```
> local vgchange_options
> vgchange_options=$(get_activate_options)
> ```

Yes, I think that this is anyway a good idea.

dmuhamedagic:
On Wed, Feb 24, 2016 at 03:10:33PM -0800, Kristoffer Grönlund wrote:
> (I still think that difference in how local is parsed is what the error message indicates)

Interesting discussion around `local`:

http://thread.gmane.org/gmane.linux.debian.devel.bugs.general/122267/focus=3271

krig:
Alright, I'll merge this since it fixes the immediate problem.

Thanks!

PR #775, opened 2016-02-26
ldirectord: Add new "servicename" and "comment" options to each virtual server

mcnewton:
This is solely for the purpose of adding more useful information into the alert e-mails so they become more useful for system administrators who have many load-balanced services.

Running a load balancer service for many different people and alerts can sometimes get confusing to try and work out exactly what service it is or who is responsible for it! Have been running this code for a couple of weeks and it's working fine here.

### DIFF
--- a/ldirectord/ldirectord.cf
+++ b/ldirectord/ldirectord.cf
@@ -23,6 +23,8 @@ quiescent=no
 
 # Sample for an http virtual service
 virtual=192.168.6.240:80
+	servicename=Web Site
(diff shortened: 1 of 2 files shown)


krig:
Code looks good to me and there haven't been any objections, so I'm merging this. Thanks!

mcnewton:
Thanks! I've got about 10 load balancers with over 30 services, so this makes it much easier to know what's up/down :-)

PR #776, opened 2016-02-26
ldirectord: Clean up protocol names/abbreviations

mcnewton:
Short patch to tidy some protocol names, e.g. previously could be all different capitalisation: "HTTP", "imap" and "Radius".


krig:
Looks good to me, thanks!

PR #778, opened 2016-03-02
Update OCF_RESOURCE_INSTANCE default for meta-data actions

kgaillot:
This grew out of the discussion for https://github.com/ClusterLabs/pacemaker/pull/920/files .

Resource agents often have parameters whose default values include the resource instance. Current approaches to representing that have various deficiencies. Defaulting OCF_RESOURCE_INSTANCE to "RESOURCE_ID" for meta-data actions allows agents to use ${OCF_RESOURCE_INSTANCE} in meta-data defaults and get something useful (UIs that know the instance name can pass it, and those that don't will get a fairly understandable string).

This request also updates .gitignore with files generated by ./autogen.sh && ./configure


krig:
Looks good to me

PR #779, opened 2016-03-03
Added pgagent RA

OlegPS:
Thanks!

### DIFF
--- a/heartbeat/pgagent
+++ b/heartbeat/pgagent
@@ -1,10 +1,8 @@
 #!/bin/sh
 #
-#       High-Availability pgagent OCF resource agent
+# High-Availability pgagent OCF resource agent
 #
 # Description:  Starts/stops pgagent
-#               This agent is for Debian systems only,
-#               as it rely on start-stop-daemon.
 # Author:       Oleg Selin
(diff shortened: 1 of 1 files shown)


heartbeat/pgagent
```diff
@@ -0,0 +1,121 @@
+#!/bin/sh
+#
+#       High-Availability nginx OCF resource agent
```
krig: This description seems wrong, this is not related to nginx, right?

OlegPS:
Ups... My bad, fixed! Kristoffer, thank you!

krig:
There are no other agents that rely on `start-stop-daemon`, and it is not available on all systems (for example, on my openSUSE system, I don't have `start-stop-daemon`).

That might be OK since this is the first version of the agent, but you should probably say in the description which distributions it is written for.

oalbrigt:
I think we should start the daemon directly instead, as that will work on all distributions, and only need changes to any path parameters that might differ from one distribution to the other.

OlegPS:
Reworked now, thanks for usefull comments!

oalbrigt:
Thanks.

PR #781, opened 2016-03-16
oralsnr: fix status check fail when username is more than 8 characters long

oalbrigt:


oalbrigt:
If $ORACLE_OWNER is "oracle123" it will be displayed in `ps -e -o pid,user,args` output as "oracle1+" and the check will fail causing the status check to fail.

dmuhamedagic:
Hmpf. There could be other places where ps -o user | grep $user is used to check for processes.

dmuhamedagic:
Thanks!

oalbrigt:
It seems like that was the only occurence (from a quick grep through the RA's).

PR #783, opened 2016-03-18
oracle: "shutdown immediate;" is needed after cleanup to be able to recover from the ORA-01081 error

oalbrigt:


dmuhamedagic:
On Fri, Mar 18, 2016 at 03:16:27AM -0700, oalbrigt wrote:
>   * oracle: "shutdown immediate;" is needed after cleanup to be able to recover from the ORA-01081 error

OK. Could you please elaborate why? What happened? If shutdown is
needed, why wasn't any oracle process running? Maybe there's a
deficiency in is_proc_running()?

oalbrigt:
It happens like 1 out of 3 times when we do "pkill ora_.*" to test recovery.

dmuhamedagic:
On Mon, Mar 21, 2016 at 03:54:48AM -0700, oalbrigt wrote:
> It happens like 1 out of 3 times when we do "pkill ora_.*" to test recovery.

But (some) processes remain running?

oalbrigt:
All the processes are dead, but somehow the shutdown is needed to make it register that it can be started again.

dmuhamedagic:
OK, I guess that an extra shutdown cannot hurt.

PR #784, opened 2016-03-20
High: sapdb.sh evaluate parameter DBOSUSER also in start and stop met…

AlexanderKrauth:
The last commit on sapdb.sh was not complete. It changed the monitor method, but left out the start and stop methods. This is required, if someone uses an entirely different DBOSUSER name and not one of the defaults.


krig:
Thank you!

PR #785, opened 2016-03-24
Low: clvm: Check use_lvmetad (bsc#970439)

krig:
lvmetad is not cluster aware.

If lvmetad is enabled in lvm.conf as "use_lvmetad = 1",
the agent must not start lvm resources, and must report
failure, and should provide a suggestion to the user to
disable lvmetad.

### DIFF
--- a/heartbeat/clvm
+++ b/heartbeat/clvm
@@ -151,6 +151,12 @@ clvmd_validate()
 		check_binary $LVM_VGSCAN
 	fi
 
+	LVM_CONF_F="/etc/lvm/lvm.conf"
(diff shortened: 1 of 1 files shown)


krig:
Turns out `lvmconf` should take care of this:

> Hmm, I see the following in ocf:heartbeat:clvm:
> 
>     # autoset locking type to clusted when lvmconf tool is available
>     if [ -x "$LVMCONF"  ]; then
>             $LVMCONF --enable-cluster > /dev/null 2>&1
>     fi
> 
> According to the manual of lvmconf:
> 
>     --enable-cluster
>         Set locking_type to the default clustered type on this system.  Also
> disable lvmetad use as it is not yet supported in clustered environment.
>

PR #787, opened 2016-04-05
VirtualDomain: fix unnecessary error when probing nonexistent domain

oalbrigt:
Fix for unnecessary logging when using LXC (only shows on the nodes where the container hasnt been run yet):
VirtualDomain(container1)[4736]: ERROR: Unable to determine emulator for lxc1


krig:
Looks good to me.

PR #788, opened 2016-04-10
heartbeat/mysql: Handle non-standard mysql server port

hackman:
In case the mysql master server is running on non-standard port, like 3308 for example, and on the same machine there is another mysql server listening on 3306, the ocf script sets the wrong details.

The problem comes from the fact, that this OCF script exports a parameter replication_port, which is actually never used.

Signed-off-by: Marian Marinov <mm@yuhu.biz>


krig:
Good catch, thanks!

PR #789, opened 2016-04-19
tickle_tcp: Fix "Failed to open raw socket (Invalid argument)" issue

oalbrigt:
Without patch:
```
# /usr/libexec/heartbeat/tickle_tcp -n 3 < /tmp/tickle/10.34.70.84
Failed to open raw socket (Invalid argument)
Error while sending tickle ack from '10.34.70.84:5000' to '10.34.71.128:6000'
# echo $?
255
```


krig:
`htons` in the socket call is a bug inherited from CTDB. It has already been fixed there:

https://lists.samba.org/archive/samba/2016-January/197434.html

PR #791, opened 2016-04-20
Medium: symlink: Handle missing directories in target (bsc#973054)

krig:
If the directory in which the symlink target is not existing, migration
after failure fails due to the probe reporting an error on the failed
node with a broken symlink.

To handle this, use "readlink -m" over "readlink -f". From the readlink
man page:

>        -f, --canonicalize
>               canonicalize by following every symlink in every component of the given name recursively; all but the last component must exist
>
>        -e, --canonicalize-existing
>               canonicalize by following every symlink in every component of the given name recursively, all components must exist
>
>        -m, --canonicalize-missing
>               canonicalize by following every symlink in every component of the given name recursively, without requirements on components existence

### DIFF
--- a/heartbeat/symlink
+++ b/heartbeat/symlink
@@ -123,7 +123,7 @@ symlink_monitor() {
             ocf_log debug "$OCF_RESKEY_link exists but is not a symbolic link, will be moved to ${OCF_RESKEY_link}${OCF_RESKEY_backup_suffix} on start"
             rc=$OCF_NOT_RUNNING
         fi
-    elif readlink -f "$OCF_RESKEY_link" | egrep -q "^${OCF_RESKEY_target}$"; then
+    elif readlink -m "$OCF_RESKEY_link" | egrep -q "^${OCF_RESKEY_target}$"; then
         ocf_log debug "$OCF_RESKEY_link exists and is a symbolic link to ${OCF_RESKEY_target}."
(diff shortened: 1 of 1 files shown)


oalbrigt:
Isnt it supposed to fail if the link is missing?

krig:
No, maybe I explained it badly:

1. The symlink points to a file on a moving device (like a DRBD disk)
2. The currently active node fails, leaving the symlink behind
3. Pacemaker starts the symlink resource on another node
4. The first node recovers
5. Pacemaker now wants to move the symlink back to the recovered node
6. The symlink RA crashes on `readlink -f` returning 1 since the device that the symlink is pointing to doesn't exist, returning a bad return code
7. Pacemaker fails to migrate the symlink back

This is not desirable: If the symlink exists and the target doesn't, but the directory that the target is in does exist, the RA cleans up the symlink and everything works. It is only if the directory that the symlink target is in doesn't exist that this error happens.

krig:
To adress the comment directly, yes: It is supposed to fail if the **link** is missing. In this case, the link is in place and pointing at the correct target, but the **target** is missing. That should be allowed, though, and disallowing it makes recovery after failure impossible. It was previously allowed as long as the only part of the target path that was missing was the last part. This fix ensures that the agent behaves correctly if more of the target path is missing.

PR #795, opened 2016-04-24
Do not wait for pg_ctl stop to complete.

ChenHuajun:
There's status checking after do pg_ctl stop, so waiting for pg_ctl stop to complete is not needed.
And if pg_ctl stop do wait, may cause pgsql_real_stop timeout.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -782,7 +782,7 @@ pgsql_real_stop() {
 
     # Stop PostgreSQL, do not wait for clients to disconnect
     if [ $stop_escalate -gt 0 ]; then
-            runasowner "$OCF_RESKEY_pgctl -D $OCF_RESKEY_pgdata stop -m fast"
+            runasowner "$OCF_RESKEY_pgctl -W -D $OCF_RESKEY_pgdata stop -m fast"
     fi
 
     # stop waiting
@@ -802,7 +802,7 @@ pgsql_real_stop() {
     then
         #PostgreSQL is still up. Use another shutdown mode.
         ocf_log info "PostgreSQL failed to stop after ${stop_escalate}s using -m fast. Trying -m immediate..."
-        runasowner "$OCF_RESKEY_pgctl -D $OCF_RESKEY_pgdata stop -m immediate"
+        runasowner "$OCF_RESKEY_pgctl -W -D $OCF_RESKEY_pgdata stop -m immediate"
     fi
 
     while :


kskmori:
Thanks for your suggestions.

Yes, the patch makes the intention of the code clear and better,
but a problem is that it changes the time to stop escalation shorter
and it would trigger to stop immediate earlier than expected as before,
which will eventually affects to the time of the crash recovery.

For example in the default configuration,
the time to stop escalation would be
 - current: 90s = 60s (pg_ctl stop timeout) + 30s (stop_escalate parameter)
 - patched: 30s = 0s (pg_ctl stop w/ -W) + 30s (stop_escalate parameter)

The advised stop timeout of pgsql RA is 120s so the default to 90s
is pretty reasonable. 30s seems too short to stop forcibly.


And now I realized that the current meaning of 'stop_escalate'
parameter is really confusing and not documented precisely.
People will want to set 90s to stop_escalate instead of 30s in such case.


So I think that we should not merge this patch solely, but instead
I would suggest to changes the meaning of 'stop_escalate'
and its default value to keep the current behavior,
along with this patch.

I'm going to write such a patch and send another pull request later.

Thanks,

ChenHuajun:
Hi, @kskmori 
I think your should also consider the situation of stop's timeout be setted to a value less then 60s.
Such as the example wrote at http://clusterlabs.org/wiki/PgSQL_Replicated_Cluster
-------------------------------------------------------
op stop    timeout="60s" interval="0s"  on-fail="block"
-------------------------------------------------------
I setup my pgsql cluster referenced the above wiki page and got a timeout when stopping pgsql.

kskmori:
> I think your should also consider the situation of stop's timeout be setted to a value less then 60s.

Right. Probably stop_escalate should be adjusted based on the configured timeout when stop_escalate > stop timeout.

> Such as the example wrote at http://clusterlabs.org/wiki/PgSQL_Replicated_Cluster
> op stop timeout="60s" interval="0s" on-fail="block"
> I setup my pgsql cluster referenced the above wiki page and got a timeout when stopping pgsql.

OK, the example page needs to be updated as well:
 - use stop(etc.) timeout=120s as advised in the meta-data of the RA.
 - adding stop_escalate=0 (disable graceful stop) is also preferable in the example - graceful stop (-m fast) is valuable for a shared-storage configuration, but no benefits and only makes the fail-over slower in a replication configuration.

Many thanks for your valuable comments!

kskmori:
@ChenHuajun please take a look PR #825 as a replacement of this PR. Thanks!

ChenHuajun:
@kskmori  I think PR #825 is OK, So i close this PR

PR #796, opened 2016-04-26
Medium: ldirectord: Remove IPv6 workaround (bsc#977193)

krig:
Remove workaround for missing IPv6 support in perl-libwww.

The workaround causes issues in a mixed IPv4 and IPv6 environment,
as it overwrites the global EXTRA_SOCK_OPTS.

The underlying issue is fixed with libwww-perl/net-http#10 and libwww-perl/libwww-perl#58.

I would suggest that users and distributions make sure that their
version of libwww-perl has those patches applied, instead of trying
to work around the problem in ldirectord.

Version 6.080 of libwww-perl includes the patches.

The related CPAN issue suggests an alternative workaround if that is
not possible, see https://rt.cpan.org/Public/Bug/Display.html?id=75618

### DIFF
--- a/ldirectord/ldirectord.in
+++ b/ldirectord/ldirectord.in
@@ -2864,10 +2864,6 @@ sub check_http
 	if (inet_pton(AF_INET6,&ld_strip_brackets($host))) {
 		no warnings 'once';
 		require Net::INET6Glue::INET_is_INET6;
-		# Workaround for Net-HTTP IPv6 Address URLs Broken
-		@LWP::Protocol::http::EXTRA_SOCK_OPTS = (PeerAddr => $host,
-							 PeerHost => &ld_strip_brackets($host),
-							 Host => &ld_strip_brackets($host));
 	}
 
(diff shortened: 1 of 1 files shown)


krig:
The workaround was added in #379 and #470.

krig:
@kskmori and @MichiroHibari, what is your opinion?

kskmori:
Thanks for the info. I will take a look.

kskmori:
I agree that the latest version of the library should be used, but unfortunately it does not seem to be widely available yet at this moment (such as in RHEL, Ubuntu), so just backing out the patch makes it unusable on such distros and probably it would take a while until all the distros upgrade the library.

My colleague is looking into this issue and he is now writing another patch so that it works both on the old version and the latest version of the library, as well as solves a problem in a mixed IPv4/IPv6 environment.

krig:
That sounds reasonable, sure.

krig:
Closing this as there is an alternative patch in #814 now.

PR #797, opened 2016-04-27
managing "Secondary Private IP" for EC2 instances

guessi:
sample config:

    primitive res_awsvip ocf:heartbeat:awsvip \
        params awscli="/usr/local/bin/aws" \
        params secondary_private_ip="172.31.0.123"

Signed-off-by: guessi <guessi@gmail.com>

### DIFF
--- a/heartbeat/awsvip
+++ b/heartbeat/awsvip
@@ -1,0 +1,222 @@
+#!/bin/sh
+#
+#
+#    Manage Secondary Private IP with Pacemaker
+#
+#
+# Copyright 2016 guessi <guessi@gmail.com>
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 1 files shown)


krig:
Looks good to me! My only comment would be the `sleep 3` in the `start` and `stop` actions. Maybe this time should be configurable? But that can be added later if necessary.

krig:
An option to sleeping might be to ping the assigned IP in a loop, something like

```
while [ . ]; do
    ping -nq -c 1 -W 1 ${SECONDARY_PRIVATE_IP} >/dev/null && break
    sleep 1
done
```

guessi:
@krig 

I've thought about the `ping` trick, but it need to execute `ifconfig eth0:vip 1.2.3.4` to get IP address up,

btw, I would recommend to use with the RA `IPaddr2` together, with the rule of `order` and `colocation`.

guessi:
@krig 

as for the `sleep 3`, I'll update later for making it configurable.

krig:
Aha, I see how it works. That sounds good then.

guessi:
@krig  please review it again, thanks

I will squash them into one when it is ready.

krig:
Looks good!

krig:
OK, merging these. Thanks!

PR #799, opened 2016-04-28
Low: Fix protential error for RA:nagios

guessi:
VAR="/etc/path with spcaes"

`[ ! -f ${VAR} ]`
equals to `[ ! -f /etc/path with spaces ]` ### syntax error

`rm -f ${VAR}`
equals to `rm -f /etc/path with spaces`    ### accidentally remove files/folders

`-eq` and `-ne` with number (not string)

Signed-off-by: guessi <guessi@gmail.com>

### DIFF
--- a/heartbeat/nagios
+++ b/heartbeat/nagios
@@ -130,7 +130,6 @@ nagios_start() {
         return $rc
     fi
 
-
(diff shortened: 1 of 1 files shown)


krig:
Looks good to me.

PR #803, opened 2016-05-03
High: SAPDatabase: Add START|STOP_TIMEOUT parameters

ionutgogu:
Pass the timeout on stop/start operation to saphostctrl in order to avoid that SAPDatabase will run into timeout with big databases.

-with big SAP Hana Database, saphostctrl is running into a timeout when a start/stop is triggered, because SAPDatabase is not passing  the timeout parameter to saphostctrl and saphostctrl is using default start/stop timeout which is not enough.

-saphostctrl running into a timeout is reporting the exit code to SAPDatabase and Pacemaker is interpreting the response as a fail, even if the SAP Hana is stopping / starting in background.

-saphostctrl accept timeout as argument:
StartDatabase
    -dbname <DB name> -dbtype <ada|db6|mss...> [-dbhost <hostname>] [-dbinstance <instance name>] [-dbuser <DB admin username>] [-dbpass <DB admin password>] [-timeout <timeout in sec>] [-service] [-instance] [-force]
  StopDatabase
    -dbname <DB name> -dbtype <ada|db6|mss...> [-dbhost <hostname>] [-dbinstance <instance name>] [-dbuser <DB admin username>] [-dbpass <DB admin password>] [-timeout <timeout in sec>] [-service] [-instance] [-force]

### DIFF
--- a/heartbeat/SAPDatabase
+++ b/heartbeat/SAPDatabase
@@ -2,7 +2,7 @@
 #
 # SAPDatabase
 #
-# Description:	Manages any type of SAP supported database instance
+# Description:  Manages any type of SAP supported database instance
 #               as a High-Availability OCF compliant resource.
 #
 # Author:       Alexander Krauth, October 2006
@@ -10,7 +10,7 @@
 # License:      GNU General Public License (GPL)
 # Copyright:    (c) 2006, 2007, 2010, 2012 Alexander Krauth
 #
-# An example usage: 
+# An example usage:
 #      See usage() function below for more details...
 #
 # OCF instance parameters:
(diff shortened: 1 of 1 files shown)


krig:
Could you combine this pull request and #802 into a single pull request?

Also, please change the commit messages to something more descriptive. I would suggest

    High: SAPDatabase: Add START|STOP_TIMEOUT parameters

For this commit, and 

    High: sapdb.sh: Pass timeouts to saphostctrl

For the commit in #802.

Thanks!

krig:
To change the commit messages, you can use this command while in your `patch-2` branch :

```
git rebase -i HEAD~2
```

This will open a text editor with a view that looks something like this:

```
pick 013872a Update SAPDatabase
pick 37e1008 Update SAPDatabase
```

Change `pick` to `reword`, save and close the file.

You will then be prompted to edit each commit message.

Once you have edited your messages, you will need to force push to this branch in your repository, and this pull request will be updated.

Hope that helps!

PR #809, opened 2016-05-17
Low NFS: add more default init scripts to check

dhoffend:
This change adds the Debian default NFS server init script to the resource agent.

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -16,11 +16,15 @@ if is_redhat_based; then
 	. ${OCF_FUNCTIONS_DIR}/nfsserver-redhat.sh
 fi
 
+DEFAULT_INIT_SCRIPT_LIST="/etc/init.d/nfsserver /etc/init.d/nfs /etc/init.d/nfs-kernel-server"
 DEFAULT_INIT_SCRIPT="/etc/init.d/nfsserver"
-if ! [ -f $DEFAULT_INIT_SCRIPT ]; then
-	# On some systems, the script is just called nfs
-	DEFAULT_INIT_SCRIPT="/etc/init.d/nfs"
-fi
+for script in $DEFAULT_INIT_SCRIPT_LIST
+do
+	if [ -f $script -a -x $script ]; then
+		DEFAULT_INIT_SCRIPT=$script
+		break
+	fi
+done
 
 DEFAULT_NOTIFY_CMD=`which sm-notify`
 DEFAULT_NOTIFY_CMD=${DEFAULT_NOTIFY_CMD:-"/sbin/sm-notify"}


heartbeat/nfsserver
```diff
@@ -16,11 +16,15 @@ if is_redhat_based; then
...
 
+DEFAULT_INIT_SCRIPT_LIST="/etc/init.d/nfsserver /etc/init.d/nfs /etc/init.d/nfs-kernel-server"
 DEFAULT_INIT_SCRIPT="/etc/init.d/nfsserver"
-if ! [ -f $DEFAULT_INIT_SCRIPT ]; then
-	# On some systems, the script is just called nfs
-	DEFAULT_INIT_SCRIPT="/etc/init.d/nfs"
-fi
+for script in $DEFAULT_INIT_SCRIPT_LIST
+do
+	if [ -f $script -a -x $script ]; then
```
krig: IIRC, -a for AND is not in the POSIX standard. I'd recommend changing this to
```
if [ -f $script ] && [ -x $script ]; then
```

dhoffend:
Well I can change it. But the same code `-a` is used in the same script 

here:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/nfsserver#L368

The for loop was copied from the apache ra
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/apache#L649

I can change it later (when I get back home) if that makes it better

heartbeat/nfsserver
dmuhamedagic: > IIRC, -a for AND is not in the POSIX standard.

'-a' is common. Documented in
http://www.unix.com/man-page/posix/1p/test/ too.

> if [ -f $script ] && [ -x $script ]; then

Of course, this would do too.

krig: So maybe the reason to avoid `-a` is folklore, but here is a justification I found:

> POSIX®/SUSv3 does not specify the behaviour of test in cases where there are more than 4 arguments. If you write a script that might not be executed by Bash, the behaviour might be different!

dmuhamedagic: On Wed, May 18, 2016 at 09:04:12AM -0700, Kristoffer Grönlund wrote:
> So maybe the reason to avoid `-a` is folklore, but here is a justification I found:
> 
> > POSIX®/SUSv3 does not specify the behaviour of test in cases where there are more than 4 arguments. If you write a script that might not be executed by Bash, the behaviour might be different!

Interesting, never saw the number of arguments mentioned, but
perhaps one needs to look at the code ;-)

krig:
No need, I guess cleaning that kind of thing up ought to be a separate pull request anyway.

I'll merge this.

PR #810, opened 2016-05-20
nfsserver: mount based on rpcpipefs_dir variable

oalbrigt:


ytakeshita:
I think the problem occurs if rpcpipefs_dir is default.

In this case after executing bind_tree() /var/lib/nfs/rpc_pipefs is lost because $fp/rpc_pipefs doesn't exist.
As a result mounting sunrpc is failed and starting nfs-idmapd.service is failed too.
Now, nfs-idmapd is invoked by nfsserver RA, so this cause failure of nfsserver.

Additionaly　on ```RHEL 7.2``` and systemd, var-lib-nfs-rpc_pipefs.mount is active in the start-up time.(maybe by dependency)
```
# mount | grep sunrpc
sunrpc on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
sunrpc on /var/lib/nfs/rpc_pipefs type rpc_pipefs (rw,relatime)
```
This causes double mount of sunrpc, but the impact of this is unknown to me...

PR #813, opened 2016-05-30
High: SAPDatabase: Revert previous changes

krig:
Reverts these submit requests: #802, #803, #806, #811 

I have learned that I handled these incorrectly and should have checked with the SAP Linux Lab before merging them.

I will let @AlexanderKrauth decide how to proceed with these changes suggested by @ionutgogu .


AlexanderKrauth:
Hi Kristoffer, as maybe already discussed with Fabian:
The idea is **not** to implement a second timeout mechanism inside the resource agent, beside the existing timeouts used by pacemaker for the different actions. This will be more confusing for the configuration, than it has any added value.
In fact, if no timeout is specified on the commandline of saphostctrl, the timeout should be infinite. This behaviour assures, that only the timeouts configured for the action itself are relevant.
The provider of this patch has also raised a customer message at SAP to get this fixed, which is the right way in this case. Currently the timeout for SAP HANA databases has a wrong default in saphostctrl. This is the point where it must be fixed. This has now also been directly forwarded to the responsible developers at SAP.
Thanks for your understanding.

krig:
Thank you for the update, that makes perfect sense.

PR #814, opened 2016-05-31
Medium: ldirectord: Fix the issue in IPv4/IPv6 mixed environment

kjtakeda:
This patch includes the following two points.
* Fix the issue (pointed out in #796) in IPv4/IPv6 mixed environment.
It was resolved by always overwrite EXTRA_SOCK_OPTS in either case IPv4/IPv6.
* Modify the IPv6 workaround (added in #379 #470) to be applied only some environment 
where old perl libraries are installed.
Libraries versions of interest are the following URL for reference.
https://rt.cpan.org/Public/Bug/Display.html?id=75618#txn-1390771

### DIFF
--- a/ldirectord/ldirectord.in
+++ b/ldirectord/ldirectord.in
@@ -2889,6 +2889,8 @@ sub check_http
 {
 	use LWP::UserAgent;
 	use LWP::Debug;
+	use Net::HTTP;
+	use URI;
 	if($DEBUG > 2) {
 		LWP::Debug::level('+');
 	}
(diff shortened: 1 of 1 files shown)


krig:
Looks OK to me.

kskmori:
@krig thanks for your review!
I will merge it after @kjtakeda has finished his extensive tests of the patch unless anybody else has other comments.
Thanks,

kskmori:
Merging as the tests has done by @kjtakeda .

krig:
Great, thank you!

PR #815, opened 2016-05-31
Low: extend support for AWS elastic ip association

guessi:
make it possible to associate secondary private ip address with elastic
ip address, if allocation-id.

if private-ip-address were both set, it will try to associate elastic
interface with the allocation-id, otherwise, it will try to associate
elastic ip with private ip instead (default).

reference docs:
http://docs.aws.amazon.com/cli/latest/reference/ec2/associate-address.html

### DIFF
--- a/heartbeat/awseip
+++ b/heartbeat/awseip
@@ -78,6 +78,22 @@ reserved elastic ip for ec2 instance
 <content type="string" default="" />
 </parameter>
 
+<parameter name="allocation_id" unique="1" required="0">
+<longdesc lang="en">
+reserved allocation id for ec2 instance
+</longdesc>
+<shortdesc lang="en">reserved allocation id for ec2 instance</shortdesc>
+<content type="string" default="" />
+</parameter>
+
+<parameter name="private_ip_address" unique="1" required="0">
+<longdesc lang="en">
+predefined private ip address for ec2 instance
+</longdesc>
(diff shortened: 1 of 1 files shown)


guessi:
@krig please review the changes

heartbeat/awseip
```diff
@@ -116,10 +132,24 @@ END
...
-    $AWSCLI ec2 associate-address  \
-        --instance-id ${INSTANCE_ID} \
-        --public-ip ${ELASTIC_IP}
-    RET=$?
+    ASSOCIATE_METHOD="ELASTIC_IP"
+    if [ -n "${ALLOCATION_ID}" ] && [ -n "${PRIVATE_IP_ADDRESS}" ]; then
+        ASSOCIATE_METHOD="ALLOCATION_ID"
+    fi
+
+    if [ "${ASSOCIATE_METHOD}" = "ALLOCATION_ID" ]; then
```
krig: Why the extra condition above, why not just check `[ -n "${ALLOCATION_ID}" ] && [ -n "${PRIVATE_IP_ADDRESS}" ]` here directly?

guessi: Uhh... what a stupid dummy code, I was debugging with this block, I'll remove it later.

krig:
Looks good to me now.

krig:
Squash the two commits together before merging?

guessi:
@krig squashed

PR #819, opened 2016-06-02
garbd: Introduces garbd resource-agent

dciabrin:
This pull request provides a resource agent for garbd, the Galera arbitrator.
It allows managing 2-nodes Galera clusters with an additional arbitrator for quorum.


oalbrigt:
Looks good. Thanks.

Issue #592, opened 2015-03-30
Consider renaming the doc/dev-guides/ra-dev-guide.txt file


beekhof:
Give it an .asc suffix and github will render it in a pretty format automatically.

PR #820, opened 2016-06-02
Low: Rename ra-dev-guide.txt to ra-dev-guide.asc

krig:
Renaming the file enables automatic syntax highlighting.

Fixes: #592


krig:
Any opinions on this from @oalbrigt or @dmuhamedagic ? Seems safe enough to me.

oalbrigt:
I think we should use the .md extension (as I guess we're using Markdown and not AsciiDoc).

https://github.com/github/markup/blob/master/README.md

krig:
No, the ra-dev-guide is written in Asciidoc. Try converting it to Markdown using a markdown converter and you'll see it won't look very good :)

I guess there's a separate pull request we could make to actually convert it to Markdown instead of Asciidoc, but that would take more than just renaming the file...

oalbrigt:
Then I'm fine with ".asc".

krig:
Alright, thanks! Merging.

PR #824, opened 2016-06-12
Dev: Clean up Makefile.am and add awseip, awsvip, pgagent and vsftpd to agent list

guessi:
RA missing: awseip, awsvip, pgagent, vsftpd


krig:
LGTM

PR #825, opened 2016-06-16
pgsql: fix stop_escalate parameter issue

kskmori:
This pull request fixes the issues below as discovered in #795.
 - stop escalation did not work and caused an unnecessary fencing when a stop timeout is configured too short (approx. <= 60s)
 - 'stop_escalate' parameter was obscure and hard to estimate the time to invoke stop escalation


krig:
Looks good to me.

PR #826, opened 2016-06-17
mysql: preserve mysql_common_validate return code 

kleind:
mysql: preserve mysql_common_validate return code if validate failed during monitor. OCF_NOT_RUNNING is not true.

We observed this when a node that did not have mysqld installed joined an existing cluster. monitor_0 happily returned OCF_NOT_RUNNING and the cluster never knew that node was not capable of running the mysqld resource.

```
Jun 17 12:05:33 node02 mysql(MysqlServer)[6848]: ERROR: Setup problem: couldn't find command: /usr/bin/mysqld_safe
Jun 17 12:05:33 node02 mysql(MysqlServer)[6848]: INFO: MySQL is not running
Jun 17 12:05:33 node02 lrmd[6708]:   notice: MysqlServer_monitor_0:6848:stderr [ ocf-exit-reason:Setup problem: couldn't find command: /usr/bin/mysqld_safe ]
Jun 17 12:05:33 node02 crmd[6711]:   notice: Operation MysqlServer_monitor_0: not running (node=node02, call=23, rc=7, cib-update=13, confirmed=true)
Jun 17 12:05:33 node02 crmd[6711]:   notice: node02-MysqlServer_monitor_0:23 [ ocf-exit-reason:Setup problem: couldn't find command: /usr/bin/mysqld_safe\n ]
```

### DIFF
--- a/heartbeat/mysql
+++ b/heartbeat/mysql
@@ -1019,8 +1019,8 @@ if [ $rc -ne 0 ]; then
                 ocf_exit_reason "environment validation failed, active pid is in unknown state."
                 exit $OCF_ERR_GENERIC
             fi
-            # validation failed and pid is not active, it's safe to say this instance is inactive.
-            exit $OCF_NOT_RUNNING;;
+            # validation failed
+            exit $rc;;
 
         status) exit $LSB_STATUS_STOPPED;;
         *) exit $rc;;


kleind:
Although nothing happened here in over three months, I just stumbled upon this thing again and think my patch is not correct. Instead it breaks probes in clusters where mysql is on not mounted storage (like drbd).
My intention was to fix the situation that the mysqld binary is not available. The patch does fix this single issue. It does however break probes since the my.cnf and the datadir (which mysql_common_validate checks for) are very likely not available on a passive node joining the cluster. So it will also return "not installed" if these files/dirs are not there although this situation may be perfectly okay in a drbd scenario.

Please do not apply this patch as is.

PR #827, opened 2016-06-17
[galera] Add /etc/default/clustercheck importation

goldyfruit:
On Debian and Ubuntu distributions, the ``/etc/sysconfig/`` directory doesn't exist.

This directory is replaced by ``/etc/default/``.


krig:
Looks good to me.

PR #831, opened 2016-06-23
RabbitMQ: Forget node before 2nd joining attempt

jistr:
If a first attempt at joining an existing cluster has failed and we
resort to wiping the local RabbitMQ data, make sure we also request the
local node to be forgotten from the existing cluster before we make the
join attempt, otherwise the node will be rejected.


jistr:
Did one more change after the last test but it was only indentation fix (i have my editor set to do spaces, had to convert to tabs). I don't think such a change warrants another retest.

oalbrigt:
Looks good. Thanks.

PR #836, opened 2016-07-21
nfsserver: systemd tests to make sure all services are running and monitored

oalbrigt:


krig:
Hmm, I don't think this will work as-is on SUSE systems, IIRC rpcbind is handled differently. I'll look into it.

oalbrigt:
Sure. Any suggestions for improvements are welcome.

krig:
I'll get back to you soon.

ytakeshita:
Great patch ! Thank you !

this while loop doesn't finish until resources's timeout (op start) although there is the counter i .
Is it as intended ?
```
+	case $EXEC_MODE in
+            [23]) nfs_exec start rpcbind
+		  local i=1
+		  while : ; do
+			ocf_log info "Start: rpcbind i: $i"
+			rpcinfo > /dev/null 2>&1
+			rc=$?
+			if [ "$rc" -eq "0" ]; then
+				break;
+			fi
+			sleep 1
+			i=$((i + 1))
+		  done
+		  ;;
+	esac
```

oalbrigt:
That is to be able to see easily if it is stuck waiting for a specific service, and that we might want to change (or remove) in the future.

ytakeshita:
OK. I see.

krig:
@oalbrigt I've checked, and I think this should work fine on SUSE as well. Merging this looks good to me. :+1:

oalbrigt:
Great. Thanks for checking.

PR #838, opened 2016-07-22
Make sure the RabbitMQ app is started on the bootstrap node

jistr:
We've seen RabbitMQ on the bootstrap node in a state where the service
is started, but the app is stopped, which prevents RabbitMQ from
forgetting nodes correctly and subsequently forming a cluster. This
commit tries to start the app on the bootstrap node if no running
cluster nodes are found.

More info and discussion at
https://bugzilla.redhat.com/show_bug.cgi?id=1343905


jistr:
Superseded by a more proper fix in #839

PR #839, opened 2016-07-26
rabbitmq: Allow automatic cluster recovery before forcing it

jeckersb:
When joining a node into an existing cluster, check to see if it is
already clustered before force removing it from the cluster and
re-adding.  If the clustering is already functional there's no need to
force it again.


lemenkov:
LGTM.

jistr:
Looks good, will try to test it tomorrow.

beekhof:
Me likey

fabbione:
thanks guys

PR #840, opened 2016-08-01
Update for version 4.0 of varnish

folcina:
Some of the parameters are no longer exist in the version 4.0 of varnish. This file is compatible with the latest version.


krig:
Looks good to me, not sure why it was overlooked.

PR #842, opened 2016-08-22
nfsserver: improve nfsserver monitor

ytakeshita:
This fix following 2 problems on monitor.
1. It is failed delete temp files and the files remain and increase on /tmp every monitoring.
2. The log message "Status: xxx" is output every monitoring. (the status is to /dev/null).

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -312,7 +312,7 @@ nfsserver_systemd_monitor()
 	local rc
 	local fn
 
-	ocf_log info "Status: rpcbind"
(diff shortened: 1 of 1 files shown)


kskmori:
The code looks good to me, and I'm going to merge this if no objections because it is necessary to make nfsserver work properly.

@oalbrigt : Would you make sure that the fixes are correct because those codes are introduced in #836 .
Thanks,

oalbrigt:
@kskmori: looks good to me.

kskmori:
@oalbrigt Thanks!

PR #845, opened 2016-09-16
portblock: create tickle_dir if it doesnt exist

oalbrigt:


krig:
Hmm, wouldn't it be better to check if the tickle_dir exists in the validate-all action?

Creating the directory if it doesn't exist masks typos and configuration errors, where someone mistyped the directory they meant or something like that. Checking in validate-all and setting exit-reason would be clearer, I think...

oalbrigt:
Updated to create it in the validate-all action.

heartbeat/portblock
```diff
@@ -466,8 +466,7 @@ IptablesValidateAll()
 		exit $OCF_ERR_CONFIGURED
 	fi
 	if [ ! -d "$OCF_RESKEY_tickle_dir" ]; then
```
dmuhamedagic: You could also get rid of this if statement. mkdir -p will just exit if the directory exists.

dmuhamedagic:
Kristoffer's note is also a valid concern and, even though we do have occasional RA that creates a directory automatically, in this case I lean more to reporting an error instead (as it is now). This directory must anyway be either shared or synced, hence it is necessary to create it on only one node.

oalbrigt:
I have updated the patch based on his concern.

I think it's an easy fix to avoid annoyances when you forget to create the directory.

PR #850, opened 2016-09-26
Low: slapd: Addition of the error handling of kill command.

HideoYamauchi:
Actually, it is unlikely to be the failure of the kill command.
However, it is handled as success because there is not error processing when an error happens.

Hideo Yamauchi.

### DIFF
--- a/heartbeat/slapd
+++ b/heartbeat/slapd
@@ -228,6 +228,10 @@ terminate()
   local waited=0
 
   kill -$signal $pid >/dev/null 2>&1; rc=$?
+  if [ $rc -ne 0 ]; then
+    ocf_exit_reason "Failed in the $signal signal of the kill command to slapd."
+    return 2 
+  fi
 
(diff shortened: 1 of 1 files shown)


krig:
It looks like with this change, return codes other than 1 or 2 are ignored, while before they would call `terminate`. Is that correct?

HideoYamauchi:
Hi Kristoffer,

Because the return code except 1 and 2 has to process it normally, I think that it is right.
Is there a point of the concern?
Possibly am I wrong?

Best Regards,
Hideo Yamauchi.

dmuhamedagic:
There's also ocf_stop_processes(). Perhaps you could consider using that?

HideoYamauchi:
Hi Dejan,
Hi Kristoffer,

> There's also ocf_stop_processes(). Perhaps you could consider using that?

Okay!

I send a new patch again.
This request withdraws it once.

Best Regards,
Hideo Yamauchi.

PR #852, opened 2016-09-28
improve robustness of sg_persist resource agent

sckobras:
Various changes to improve robustness of the resource agent:

- Improve parsing of sg_persist (the tool) output to only match node ids, not arbitrary other numbers that happen to fit the pattern;
- fix invalid attribute names when running in debug mode;
- don't send informational output to stderr;
- cleanup nesting of code.


krig:
LGTM. Thank you!

PR #857, opened 2016-10-10
New resource attribute "start_scripts"

berlevdv:
Create new resource attribute "start_scripts" for control domain start via scripts.


krig:
Why not create a separate resource for your start script (using the `anything` resource agent) and group that resource with your VirtualDomain resource? That way, your script will run first and only if it starts successfully will the VirtualDomain resource start.

No need to add the "start_scripts" parameter to VirtualDomain.

berlevdv:
I thought that once used the "monitor scripts", which functionality can be used by "anything" resource agent, it would be a good idea here is to implement the "start scripts". In addition it will prevent "excess" resources, their setting and specify same configuration file path at different resources.

krig:
The API for `start_scripts` is not the same as for `monitor_scripts`: It looks like a start script takes a domain name as parameter, while the monitor scripts seem to not take any parameters. Is this correct?

Either way, the difference should be documented in the `longdesc` for `start_scripts`, at least.

berlevdv:
Hi, Krig! Yes, start script section of VirtualDomain_start() function pass $DOMAIN_NAME variable as parameter to script for domain identification and further usage. With this option there is no need to manually specify the domain name (which may change in the future) when defining the VirtualDomain resource, such as in monitor_scripts.

I will add describe of this option in the longdesc for start_scripts. Thanks!

krig:
Alright, looks good to me now.

krig:
Could you collapse the commits into a single commit with a commit title like

```
Medium: VirtualDomain: Add start_scripts parameter
```

?

To do this, check out your branch (`master` in this case) and then run

```
git reset --soft HEAD~2
git commit --amend
```

Then once it looks good, `git push -f` updates this pull request.

PR #862, opened 2016-10-19
Delay: set default startdelay lower than start timeout to avoid timing out before starting with default values

oalbrigt:


krig:
Makes sense.

PR #867, opened 2016-10-28
High: pgsql: follow the specification change of "synchronous_standby_names" on PostgreSQL 9.6

ytakeshita:
This patch fix the following issue on PostgreSQL 9.6
On PostgreSQL 9.6 the 'synchronous_standby_names' changed to SQL identifier from plain text.
Now 'synchronous_standby_names' and 'application_name' is made from node name by pgsql RA.
There are some different rules between the node name and the SQL idenfier.
As a result, if the node name include "-", "." and the first character is number, the node
can not be allowed synchronous replication.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1120,10 +1120,19 @@ pgsql_notify() {
     return $OCF_SUCCESS
 }
 
+# convert the node name to SQL identifier
+convert_to_sqlidentifier() {
+    local identifier
+
+    identifier=$(echo "$1" | sed 's/\(^[0-9]\)/_\1/' | tr .- __)
+    echo "$identifier"
+}
+
 control_slave_status() {
     local rc
     local data_status
     local target
+    local tmp_target
     local all_data_status
     local tmp_data_status
     local number_of_nodes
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Looks good to me, shellwise. The rest I'd like to leave to @kskmori. In particular, can this introduce a problem with the running resource on RA upgrade?

kskmori:
We have already tested with an older version of PostgreSQL (9.5) and it's compatible, but I now realized that it could be a problem when a rolling upgrade of resource-agents if 1) the new RA is installed to only one node in the cluster and restarted, and 2) a node name contains special characters as described (which is the case if you are using the full domain name containing '.'), so it would be safer not to change the behavior in the running system.

@ytakeshita : Will you consider the version checking so that the conversion would be done only if the PG version is >= 9.6 ? We already have some version checking in pgsql_validate_all(). We should not need to care about mixing the version of PostgreSQL 9.5 and 9.6 because PostgreSQL does not support it AFAIK.

Thanks,

ssgelm:
In our tests, it seems that we were able to make it work by just quoting the fqdn.  E.g.:
```
synchronous_standby_names = '"HOSTNAME.FQDN"'
```
I think that might be a less invasive solution to the problem.

ytakeshita:
@ssgelm : Yes. quoting the value of synchrnous_standby_names make it works fine on PostgreSQL 9.6 and I think that it is better now.
I will create the new patch and PR.

Initially I wanted to avoid version check to unify processing.
Quoting the value needs version checking or complicated proccessing because the character ```""``` is meta character on PostgreSQL 9.6 but it is not meta character on PostgreSQL 9.5 or less.
Also, this cause inconsistency between synchrnous_standby_names and application_name in recovery.conf without version checking.

But now , in order to deal with the pointing by @dmuhamedagic and @kskmori , version check is essential.
So, I think that the quoting the value of synchrnous_standby_name with version checking is simple solution to this problem.

thanks,

ytakeshita:
#900 is better than this.
close this PR

PR #869, opened 2016-11-01
nfsserver: dont stop rpcbind, as other services might use it

oalbrigt:
Services like NIS/ypbind use rpcbind and will stop when it is stopped


krig:
Makes sense to me. There shouldn't be any reason for the agent to stop rpcbind. Just need it to be running before the server starts.

PR #872, opened 2016-11-01
Redis: restore rundir security context

gabrielecerami:
When selinux rules packages are installed, rundir does not yet exist,
and security context for it cannot be applied. Calling restorecon after
dir creation ensures that the proper context is applied to the rundir.
If the context is not applied, selinux denies write permission, the unix
socket cannot be created, and redis does not start

### DIFF
--- a/heartbeat/redis
+++ b/heartbeat/redis
@@ -324,6 +324,10 @@ function start() {
 
 	[[ ! -d "$REDIS_RUNDIR" ]] && mkdir -p "$REDIS_RUNDIR"
 	chown -R "$REDIS_USER" "$REDIS_RUNDIR"
+	if have_binary "restorecon"; then
+		restorecon -Rv "$REDIS_RUNDIR"
+	fi
+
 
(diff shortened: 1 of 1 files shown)


heartbeat/redis
```diff
@@ -324,6 +324,7 @@ function start() {
 
 	[[ ! -d "$REDIS_RUNDIR" ]] && mkdir -p "$REDIS_RUNDIR"
 	chown -R "$REDIS_USER" "$REDIS_RUNDIR"
+	restorecon -Rv "$REDIS_RUNDIR"
```
krig: What if SELinux is not available? In that `restorecon` won't be there and this will crash, right?
Shouldn't this be guarded by `have_binary`?

gabrielecerami: You're right, I'll change the pull request. What's the preferred method here ? amend and force push on the same branch, push to a different branch and change PR base, other ?

krig: Amend and force push on the same branch is what I usually do.

PR #875, opened 2016-11-07
Low: slapd: Change of the stop processing.

HideoYamauchi:
Hi All,

I change the stop processing to the ocf_stop_processes processing.
Because the time-out of the KILL stop becomes effective, it may become late than before, but is not different in movement.

This correction changed former pullrequest.
 - https://github.com/ClusterLabs/resource-agents/pull/850

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/heartbeat/slapd
+++ b/heartbeat/slapd
@@ -219,32 +219,6 @@ Maximum number of open files (for ulimit -n)
 END
 }
 
-terminate()
-{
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Looks good. Thanks!

HideoYamauchi:
Thanks!

PR #876, opened 2016-11-07
Dev: slapd: remove unused and confusing code from status

dmuhamedagic:


krig:
Right, changing the title here doesn't fix the typo in the commit message. Maybe change that before merging?

dmuhamedagic:
On Tue, Nov 15, 2016 at 01:48:09AM -0800, Kristoffer Grönlund wrote:
> Right, changing the title here doesn't fix the typo in the commit message. Maybe change that before merging?

Ah, right, will do that.

PR #879, opened 2016-11-11
rgmanager: fix up for wrongly generated RNG schema for cluster config

jnpkrn:
...wrt. "action" tag

Related: rhbz#1272587


oalbrigt:
Looks good to me. Thanks.

PR #884, opened 2016-11-18
galera: make last commit parsing compatible with mariadb 10.1.18+

dciabrin:
Starting with mariadb 10.1.18, mysqld_safe does no longer output the WSREP last commit to the standard output when called with command line option --wsrep-recover. Instead, it logs that information into mariadb's error log.

Change the way the resource agent recovers that information by forcing logging to a file and parsing info from this file.

The change is backward compatible. Tested with ra-tester.


oalbrigt:
Looks good to me. Thanks.

Issue #883, opened 2016-11-17
heartbeat/findif.sh: ifcheck must not suppress stdout


turchanov:
ifcheck must not suppress stdout because a subsequent invocation in findif.sh:141
reads stdout to *errmsg* to print it out in case of an error:

```
  if [ -n "$nic" ] ; then
    errmsg=`ifcheck $nic`
    if [ $? -ne 0 ] ; then
      ocf_log err "Invalid interface name [$nic]: $errmsg"
      return $OCF_ERR_CONFIGURED
    fi
  fi
```

Proposed fix:
```
ifcheck() {
  local ifname="$1"
-  $IP2UTIL link show dev $ifname >/dev/null 2>&1
+  $IP2UTIL link show dev $ifname 2>&1
}
prefixcheck() {
```

dmuhamedagic:
Good catch! Care to make a pull request? Thanks!

turchanov:
Why not? Just did that. Pull #886

krig:
Resolved by #866

PR #886, opened 2016-11-19
ifcheck must not suppress stdout of "ip" command …

turchanov:
so that it can be printed out in case of an error (e.g., usage in findif_check_params).

This resolves the issue #883


krig:
Thanks!

PR #888, opened 2016-11-21
add network namespace support to IPaddr2

phemmer:
Simple implementation of network namespace support. Basically the script changes its own network namespace if configured. Then any commands run by the script inherit the namespace.

Just FYI, we've been using this since around the time this PR was created (now 7 months as of mid-June) on 15 different servers, with 255 total different VIPs, and have not had any issues with it.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -368,6 +368,16 @@ Expects a value as specified in section 5.5.4 of RFC 4862.
 <content type="string" default="${OCF_RESKEY_preferred_lft_default}"/>
 </parameter>
 
+<parameter name="network_namespace">
+<longdesc lang="en">
+Specifies the network namespace to operate within.
+The namespace must already exist, and the interface to be used must be within
+the namespace.
+</longdesc>
+<shortdesc lang="en">Network namespace to use</shortdesc>
+<content type="string" default=""/>
+</parameter>
+
 </parameters>
 <actions>
 <action name="start"   timeout="20s" />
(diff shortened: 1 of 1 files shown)


phemmer:
Any status on this?
@oalbrigt Sorry for the direct ping, but this has been sitting for over 2 years without any feedback, and I'd really love if we (my company) could stop maintaining our own custom version of this resource agent, and get back to using an official version.

krig:
@phemmer Looking at that line breaks my brain a bit. :D Can you explain how that double exec construction works? I don't get why the assignment back to `OCF_RESKEY_network_namespace` is necessary, or how the first call works.

phemmer:
Heh, fair enough.

```
OCF_RESKEY_network_namespace= exec $IP2UTIL netns exec "$OCF_RESKEY_network_namespace" "$0" "$__OCF_ACTION"
```

The line basically re-executes the script, but from within the target namespace. But on the re-exec we clear the `OCF_RESKEY_network_namespace` variable to prevent an infinite loop.

The core command is
```
$IP2UTIL netns exec "$OCF_RESKEY_network_namespace" "$0" "$__OCF_ACTION"
```
which is something like
```
ip netns exec YOUR_NAMESPACE_HERE /path/to/resource-agents/IPAddr2 start
```
This invokes the `ip` utility, and says to execute the command `/path/to/resource-agents/IPAddr2 start` inside the given namespace.

The prefixed `exec` before that (between clearing the variable and the `$IP2UTIL`) is a shell-level exec because we don't need to do anything after re-exec, so the new script can just take over entirely.

krig:
Ah! Now it makes sense. That's clever. :+1:

oalbrigt:
LGTM.

Can you `git rebase upstream/master` and `git push -f` to solve the conflict?

phemmer:
Conflicts resolved.

PR #889, opened 2016-11-24
fs.sh: add parameter to be able to disable write check

oalbrigt:
For those cases where you want to disable the write check.

### DIFF
--- a/rgmanager/src/resources/bind-mount.sh
+++ b/rgmanager/src/resources/bind-mount.sh
@@ -96,6 +96,16 @@ do_metadata()
 			</shortdesc>
 		<content type="boolean"/>
 		</parameter>
+
+		<parameter name="write_check" required="0">
+			<longdesc lang="en">
+				Write check (default: on).
+			</longdesc>
+			<shortdesc lang="en">
+				Write check
+			</shortdesc>
+			<content type="boolean" default="on"/>
+		</parameter>
 	</parameters>
 
 	<actions>
--- a/rgmanager/src/resources/clusterfs.sh
+++ b/rgmanager/src/resources/clusterfs.sh
@@ -166,6 +166,16 @@ do_metadata()
             <content type="boolean"/>
         </parameter>
 
+        <parameter name="write_check" required="0">
+	    <longdesc lang="en">
+	        Write check (default: on).
+	    </longdesc>
+            <shortdesc lang="en">
+	        Write check
+            </shortdesc>
+	    <content type="boolean" default="on"/>
+        </parameter>
+
     </parameters>
 
     <actions>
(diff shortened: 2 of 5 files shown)


lge:
And the next line is:
[ $OCF_CHECK_LEVEL -lt 20 ] && return $YES

So what was the point here?
Use a check level < 20 if you don't want the write check.

oalbrigt:
check level 20 is there by default, so you'd have to set a very high interval to completely "disable" it.

lge:
why would check level 20 be the default? where?
why would that have anything do to with the interval?

why would introducing a new parameter (that you have to set in the config)
be easier than changing the "depth" (aka CHECK_LEVEL) in the config?

As I understand it, you define what status action(s) to run, in which intervals.
If none is defined, none is run.
If one is defined, with depth=10 and interval=30, then that one is run as specified.
If you chose to define three, with different depth and intervals, they are run as configured,
where, if several are "due" at the same time, only the one with the highest depth would be run.
If memory serves, something like this would do:
<resources>
  <fs name="foo" ... >
    <action name="status" depth=10 interval=30 timeout="10" />
  </fs>
</resources>

I have to admit that arguing with rgmanager has been a while,
so what am I missing?

Is this actually about some GUI shortcoming?

jnpkrn:
On 25/11/16 07:55 -0800, Lars Ellenberg wrote:
> why would check level 20 be the default? where?
> why would that have anything do to with the interval?
> 
> why would introducing a new parameter (that you have to set in the config)
> be easier than changing the "depth" (aka CHECK_LEVEL) in the config?

That's indeed nonsystemic, thanks for the noise.

> As I understand it, you define what status action(s) to run, in which intervals.
> If none is defined, none is run.
> If one is defined, with depth=10 and interval=30, then that one is run as specified.
> If you chose to define three, with different depth and intervals, they are run as configured,
> where, if several are "due" at the same time, only the one with the highest depth would be run.
> If memory serves, something like this would do:
> <resources>
>   <fs name="foo" ... >
>     <action name="status" depth=10 interval=30 timeout="10" />
>   </fs>
> </resources>
> 
> I have to admit that arguing with rgmanager has been a while,
> so what am I missing?

Resources have implicit action preconfiguration, so now it's the
question how it combines with the user's configuration in rgmanager.

It's being looked at.

> Is this actually about some GUI shortcoming?

Fortunately not, although it's not an optimal state that web UI (luci)
could not even display that.  At least this will become eventually fixed
(but likely that's the final stop here, provided its decline...).

One can configure actions using ccs (--addaction, --rmaction).

-- 
Jan (Poki)

oalbrigt:
I'll revert this for now, and redo it if we cant find a solution to this.

jnpkrn:
Solution to this is to configure an override:
https://bugzilla.redhat.com/show_bug.cgi?id=1310529#c25

The original change was rightfully reverted.
Thanks again, Lars, for an safety net you've provided here.

PR #890, opened 2016-11-24
redis: use "CLIENT KILL type normal" to notify clients of master being demoted

oalbrigt:


krig:
It seems this form of `CLIENT KILL` is only available in "Redis 2.8.12 or greater", but since the agent doesn't have any compatibility statement, that may not be a problem (and it's the previous major version anyway).

oalbrigt:
I'll add a version check to my next commit. We need to do some more testing with a new patch which will be part of my next commit.

oalbrigt:
Updated to pause connections to make the "client kill" work, and added version check.

PR #891, opened 2016-11-25
IPaddr2: return success on stop with invalid IP

oalbrigt:


dmuhamedagic:
This is an unusual way to solve the issue of failing stop in case
of an invalid configuration.

There's a lot of code being run in ip_stop(). Are you sure that
none of it is going to break with an invalid configuration?

If the configuration is invalid, we can assume that the resource
is stopped, as it never could have been started.

oalbrigt:
It's working fine in my tests, and the stop action skips it all due to it's if-statements, which is the reason I did it this way.

The reason for skipping it is because we've seen issues with Pacemaker hanging on shutdown/reboot due to it failing to stop the failed/already stopped process.

kskmori:
It does not seem to be working properly, even in a "normal stop" test case in ocft/IPaddr2 or ocft/IPaddr2v4; the stop op does return OCF_SUCCESS but apparently the assigned IP address is not removed ever.

A simple test with ocf-tester also fails:
```
[root@rhel72-1 ~]# ip -o a s eth0 | grep 192.168.100.112
[root@rhel72-1 ~]# ocf-tester -n test1 -o ip=192.168.100.112 -o cidr_netmask=24 -o nic=eth0 /usr/lib/ocf/resource.d/heartbeat/IPaddr2
Beginning tests for /usr/lib/ocf/resource.d/heartbeat/IPaddr2...
* Your agent does not support the notify action (optional)
* Your agent does not support the demote action (optional)
* Your agent does not support the promote action (optional)
* Your agent does not support master/slave (optional)
* rc=0: Monitoring a stopped resource should return 7
* rc=0: Monitoring a stopped resource should return 7
* rc=0: Monitoring a stopped resource should return 7
* Your agent does not support the reload action (optional)
Tests failed: /usr/lib/ocf/resource.d/heartbeat/IPaddr2 failed 3 tests
[root@rhel72-1 ~]# ip -o a s eth0 | grep 192.168.100.112
2: eth0    inet 192.168.100.112/24 brd 192.168.100.255 scope global secondary eth0\       valid_lft forever preferred_lft forever
[root@rhel72-1 ~]# 
```

I don't think that we can just skip ip_validate() and ip_init() because it does a lot of initializations of variables in order to work ip_start()/ip_stop() properly.

oalbrigt:
Good catch. I'll have to find a better way to do this.

oalbrigt:
It now passes the IPaddr2v4 and IPaddr2v6 ocft-tests without any issues.

heartbeat/IPaddr2
```diff
@@ -449,7 +449,6 @@ ip_init() {
 			exit $OCF_NOT_RUNNING
 		elif [ "$__OCF_ACTION" = stop ]; then
 			ocf_log warn "[$FINDIF] failed"
-			exit $OCF_SUCCESS
```
oalbrigt: It somehow failed on one of my tests with it, but now it appears to not fail with it, so I'll push an updated commit.

PR #892, opened 2016-11-25
pgsql: fix problem with creating replication slot names, when node count

oscarsix:
greather then 2

the problem is that tr changes the replication slot string not
correnctly:

opsdbatest03 -> repl_slot_opsdbatest]]

replication slot name "repl_slot_opsdbatest]]" contains
invalid character.
this fix resolve this problem


krig:
Looks good to me, thanks!

PR #895, opened 2016-11-30
Revert "fs.sh: add parameter to be able to disable write check"

oalbrigt:
Reverts ClusterLabs/resource-agents#889


jnpkrn:
Why it was sound to revert the original change: https://github.com/ClusterLabs/resource-agents/pull/889#issuecomment-272944174

PR #896, opened 2016-12-01
Dump/restore users even in case of RabbitMQ 3.6.x and also dump/restore user permissions during restarts

lemenkov:
This PR fixes dump/restore users from RabbitMQ 3.6.x where internal structure was changed a little. Also it allows dumping/restoring users' permissions (see RHBZ 1342376).


lemenkov:
This PR expands #804.

PR #897, opened 2016-12-07
RFC: RabbitMQ: Reset Mnesia before joining existing cluster

lemenkov:
--=== !!!! WARNING !!!! Don't merge it yet - we're still testing it. ==--

We should reset Mnesia always beore joining cluster. Any necessary data
will be fetched from live node(s), so why care? Not doing so could led
to stuck node which needs manual repair.

Why bother with these Mnesia bytes? Better kill it and refetch!

Also related:

* https://bugs.launchpad.net/fuel/+bug/1431761
* https://bugs.launchpad.net/fuel/+bug/1620649
* https://bugzilla.redhat.com/show_bug.cgi?id=1397393
* rabbitmq/rabbitmq-server#946

Signed-off-by: Peter Lemenkov <lemenkov@redhat.com>

### DIFF
--- a/heartbeat/rabbitmq-cluster
+++ b/heartbeat/rabbitmq-cluster
@@ -353,75 +353,71 @@ rmq_start() {
 		return $rc
 	fi
 
-	# first try to join without wiping mnesia data
+	# Try to join existing cluster
+	ocf_log info "wiping data directory before joining"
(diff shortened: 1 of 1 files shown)


lemenkov:
Ok, we've tested it and didn't find any regressions. Please consider merging it.

cc @oalbrigt

oalbrigt:
Thanks.

PR #898, opened 2016-12-08
portblock: Use -w (wait) to avoid "insufficient privileges" error

oalbrigt:
E.g. when running multiple portblock resources.


oalbrigt:
I'll add a version check for this, as it's only supported in iptables 1.4.20 and greater (this makes iptables version before 1.4.20 less usable if you need to run 2 or more portblock resources).

dmuhamedagic:
Looks good. Thanks!

PR #899, opened 2016-12-09
specfile: make requirements general and update comments

oalbrigt:
Most of these are required by heartbeat agents as well.

### DIFF
--- a/resource-agents.spec.in
+++ b/resource-agents.spec.in
@@ -79,31 +79,35 @@ BuildRequires:  libxslt docbook_4 docbook-xsl-stylesheets
 %endif
 
 ## Runtime deps
-## These apply to rgmanager agents only to guarantee agents
-## are functional
-%if %{with rgmanager}
 # system tools shared by several agents
 Requires: /bin/bash /bin/grep /bin/sed /bin/gawk
 Requires: /bin/ps /usr/bin/pkill /bin/hostname /bin/netstat
-Requires: /sbin/fuser
-Requires: /sbin/findfs /bin/mount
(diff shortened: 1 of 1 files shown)


krig:
Hmm, now that I see it, it does change things a bit... previously we didn't really require the dependencies for individual agents, but with this change we do - this means that you have to install NFS packages to get resource-agents even if you're not setting up the NFS agent, for example.

In practice, I don't know how much of an issue it really is - at least on SLE all those things are installed anyway. It might be more of a problem in resource-constrained situations, like running the stack on public cloud or something.

dmuhamedagic:
On Fri, Dec 09, 2016 at 08:35:54AM -0800, Kristoffer Grönlund wrote:
> Hmm, now that I see it, it does change things a bit... previously we didn't really require the dependencies for individual agents, but with this change we do - this means that you have to install NFS packages to get resource-agents even if you're not setting up the NFS agent, for example.

Agreed. There are so many different packages managed by resource
agents, it would be hard to draw the line. At any rate, the
actual specfile is ever going to depend on the distribution.

BTW, some of the requirements are really rgmanager specific.

oalbrigt:
Yeah. I think it makes sense to have dependencies on the most common utilities though (like netstat which is used in multiple agents) to make sure you get them if your cloud or minimal install doesnt include it from some other source.

I can go through the list and keep the rgmanager ones separate if you want.

krig:
OK, lets merge it. As @dmuhamedagic says, in the end, the specfile will probably be modified by the distribution anyway.

PR #900, opened 2016-12-12
pgsql: Double-quote value of synchronous_standby_names in rep_mode.conf

antaflos:
PostgreSQL 9.6 introduced a new syntax for specifying
synchronous_standby_names. The old syntax, used by the pgsql RA, is
still valid but PostgreSQL now treats the standby-names in
synchronous_standby_names as SQL identifiers. This means such values
need to be double-quoted since they can easily contain dashes or other
characters that are not valid in a bare SQL identifier.

See the docs for synchronous_standby_names in
https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html
for confirmation and
https://www.postgresql.org/message-id/21183.1481253534%40sss.pgh.pa.us
for a short discussion.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1493,7 +1493,7 @@ set_sync_mode() {
         ocf_log debug "$sync_node_in_conf is already sync mode."
     else
         ocf_log info "Setup $1 into sync mode."
-        runasowner -q err "echo \"synchronous_standby_names = '$1'\" > \"$REP_MODE_CONF\""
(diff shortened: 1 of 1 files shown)


kskmori:
@ytakeshita : Could you make sure that this fix is sufficient as a replacement of #867 ? Thanks!

ytakeshita:
I confirm this patch with PostgreSQL 9.5 and 9.6
I think this patch works fine and sufficify at all.
Thanks!

kskmori:
@antaflos @ytakeshita Thanks for your contribution!

PR #902, opened 2016-12-16
targetcli concurrent-safe

dmartinb:
targetcli is not concurrent-safe. Using iscsi resources in different groups and shutting down a node causes race conditions.
Basically this serializes access to targetcli via lock file.


dmuhamedagic:
On Fri, Dec 16, 2016 at 06:56:07AM -0800, dmartinb wrote:
> targetcli is not concurrent-safe.

Did you consider extending the existing approach to the
concurency issue? For instance (currently only for iet):

heartbeat/iSCSITarget:204             ocf_take_lock $LOCKFILE

dmartinb:
I see what you mean. I close this PR and I'll create another one.

PR #903, opened 2016-12-19
galera: allow names in wsrep_cluster_address to differ from pacemaker nodes' names

dciabrin:
Add a new option pcmk_host_map to the galera resource agent in case
names to be used in wsrep_cluster_address need to differ from names
used for the pacemaker nodes. (e.g. when galera names map to IP
from a specific network interface)

Option pcmk_host_map allows mapping pacemaker nodes to galera
nodes; it replicates the syntax used to set up fencing resources:

  pcmk1:node.1.galera;pcmk2:node.2.galera;pcmk3:node.3.galera

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -208,10 +208,27 @@ The galera cluster address. This takes the form of:
 gcomm://node,node,node
 
 Only nodes present in this node list will be allowed to start a galera instance.
-It is expected that the galera node names listed in this address match valid
-pacemaker node names.
+The galera node names listed in this address are expected to match valid
+pacemaker node names. If both names need to differ, you must provide a
+mapping in option cluster_host_map.
 </longdesc>
 <shortdesc lang="en">Galera cluster address</shortdesc>
+<content type="string" default=""/>
+</parameter>
+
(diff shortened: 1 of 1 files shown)


heartbeat/galera
```diff
@@ -208,13 +208,30 @@ The galera cluster address. This takes the form of:
 gcomm://node,node,node
 
 Only nodes present in this node list will be allowed to start a galera instance.
-It is expected that the galera node names listed in this address match valid
-pacemaker node names.
+The galera node names listed in this address are expected to match valid
+pacemaker node names. If both names need to differ, you must provide a
+mapping in option pcmk_host_map.
```
beekhof: Aside from the concept, the pcmk_ prefix is (or should be) reserved.
Even if its the same kind of functionality, pick a different variable name :-)

beekhof:
Apart from the naming, (perhaps s/pcmk/cluster/ ?) the concept seems reasonable given that we allow cluster nodes to have different names to the machines they're running on.

heartbeat/galera
dciabrin: Oh, okay, I've just amended the commit to name option cluster_host_map.

PR #904, opened 2016-12-21
IPsrcaddr: match exact route to avoid failing

oalbrigt:


htakeda:
Hi,
It works very well.
This is fixed required for  RHEL7(CentOS7), I guess.

htakeda:
Thank you!

PR #905, opened 2016-12-21
ocf_mkstatedir: fix path check

vvidic:
Check fails due to duble slashes in string comparison.

Path /var/run/apache2 is compared with /var/run//* giving the following error in the logs:

Dec 21 13:17:53 node1 apache(apache_/etc/apache2/apache2.conf)[826]: [868]: ERROR: cannot create /var/run/apache2 (does not start with /var/run/)

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -778,7 +778,7 @@ ocf_mkstatedir()
 	[ $(id -u) = 0 ] || return 1
 
 	case $path in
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
On Wed, Dec 21, 2016 at 07:04:52AM -0800, Valentin Vidić wrote:
> Check fails due to duble slashes in string comparison.
> 
> Path /var/run/apache2 is compared with /var/run//* giving the following error in the logs:

Ugh.

Thanks!

PR #907, opened 2017-01-06
Revert "exportfs: fix monitor function for IPv6 addresses"

oalbrigt:
Reverts ClusterLabs/resource-agents#906


oalbrigt:
My bad. This was already fixed in https://github.com/ClusterLabs/resource-agents/pull/828

PR #912, opened 2017-01-10
Fix grammar and spelling in some resource agents

ChristophBerg:
This patch had been part of the Debian package since at least 2011.


krig:
Thanks!

PR #913, opened 2017-01-11
IPaddr2 detect duplicate IP

oalbrigt:


krig:
Could you expand on the commit message a bit more? The `arping` call isn't entirely self-explanatory - at least, I'm not sure how that works reliably.

oalbrigt:
It returns 1 if the IP is in use.

It's basically the same code used in /etc/sysconfig/network-scripts/ifup-eth on RHEL systems.

krig:
I see, thanks!

dmuhamedagic:
This is going to introduce a 4 second wait in the start action. Is the extra check worth it? Shouldn't it be optional? I can vaguely recall people reporting running umpteen (as in hundreds) IP addresses.

Otherwise, we should also add 'check_binary arping'.

oalbrigt:
I've now added a run_arping parameter to be able to disable the check, and it also uses have_binary so it doesnt try to run the check if arping isnt available.

dmuhamedagic:
On Wed, Jan 18, 2017 at 04:36:54AM -0800, Oyvind Albrigtsen wrote:
> I've now added a run_arping parameter to be able to disable the check, and it also uses have_binary so it doesnt try to run the check if arping isnt available.

Great!

But I think that the default should be not to run the check, lest
we surprise people with the delay. It would be a regression of a
kind.

Then, if the run_arping is set to true and arping not available,
the RA should say OCF_ERR_INSTALLED.

oalbrigt:
I've updated it to default to false and check_binary if it's true.

PR #916, opened 2017-01-20
add nic_monitor function so VIP can takeover by the other node WHEN t…

liangxin1300:
…he net wire been pulled

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -970,6 +970,14 @@ ip_stop() {
 	exit $OCF_SUCCESS
 }
 
+nic_monitor()
+{
+        ps -ef|grep -w "NetworkManager"|grep -v grep &> /dev/null
+        if [ "$?" -ne 0 ];then
+                ethtool $NIC|grep "Link detected: yes" &> /dev/null
+        fi
+}
+
 ip_monitor() {
 	# TODO: Implement more elaborate monitoring like checking for
 	# interface health maybe via a daemon like FailSafe etc...
@@ -977,6 +985,7 @@ ip_monitor() {
 	local ip_status=`ip_served`
 	case $ip_status in
 	ok)
+                nic_monitor || return $OCF_ERR_GENERIC
 		$ARP_SEND_FUN refresh
 		return $OCF_SUCCESS
 		;;


dmuhamedagic:
Useful, but beware, ethtool depends on the hardware and network
drivers. ethtool(8) says:

	BUGS
	       Not supported (in part or whole) on all network drivers.

I am not sure whether this applies to the link detection, but
that may as well be the case.

liangxin1300:
Hi @dmuhamedagic ：
    I think this case(wire been pulled) is NOT rare,
    So, a more compatible way to the link detaction on different OS environment could be acceptable?

dmuhamedagic:
On Fri, Jan 27, 2017 at 11:47:40AM -0800, Xin Liang wrote:
> Hi @dmuhamedagic ：
>     I think this case(wire been pulled) is NOT rare,

No, it's not, though I cannot recall saying it was.

>     So, a more compatible way to the link detaction on different OS environment could be acceptable?

It's not about different OS environment, but different network
drivers and hardware. Maybe just checking the exit code would be
enough, sth like:

```
link_test() {
	local output
	output=$(ethtool $nic 2>/dev/null)
	if [ $? -ne 0 ]; then
		: interface most probably not supported
	elif echo "$output" | grep 'Link detected: no' > /dev/null; then
		return 1
	fi
	return 0
}

if ! link_test; then
	ocf_log err "ethtool reports no link"
	return $OCF_ERR_GENERIC
fi
```

I'm not sure whether this should be run on probes too: the driver
may not be loaded at the time.

NB: The code above may not be 100% correct.

bubble75:
This wont work on any interfaces except ethernet ones. bond, bridge, IPoOB, tunnels, ppp, etc.etc.etc.
There is my ifspeed agent in pacemaker extra/resources which solves the issue for any wired interfaces.

liangxin1300:
Hi @dmuhamedagic : 
     Thank you for your suggestion ！

Hi @bubble75 ：
     I just see the ifspeed agent you mentioned, great job!

     But, from the point of easy to use the HACluster,  I still think that both the IPaddr and IPaddr2 scripts could handle with the wired interfaces issues, so Users can add only one resource in the group, do not worry about ALL THE THINGS which are related to the IP or interfaces.

dmuhamedagic:
Do you care to update this pull request with the above suggestion? @bubble75, I find the code in my comment safe for non-ethernet interfaces too. Do you agree?

bubble75:
@dmuhamedagic Probably it is, but it should be thoroughly tested. Anyways, link status has nothing to IP addess IMHO.

PR #917, opened 2017-01-20
targetcli lockfile

dmartinb:
targetcli is not concurrent-safe. Using iscsi resources in different groups and shutting down a node causes race conditions.

### DIFF
--- a/heartbeat/iSCSILogicalUnit
+++ b/heartbeat/iSCSILogicalUnit
@@ -75,6 +75,8 @@ OCF_RESKEY_lio_iblock=${OCF_RESKEY_lio_iblock:-$OCF_RESKEY_lio_iblock_default}
 # OCF_RESKEY_tgt_bsopts
 # OCF_RESKEY_tgt_device_type
 
+# targetcli: iSCSITarget and iSCSILogicalUnit must use the same lockfile
+TARGETLOCKFILE=${HA_RSCTMP}/targetcli.lock
 #######################################################################
 
 meta_data() {
@@ -371,6 +373,8 @@ iSCSILogicalUnit_start() {
 		fi
 		;;
 	lio-t)
+		ocf_take_lock $TARGETLOCKFILE
+		ocf_release_lock_on_exit $TARGETLOCKFILE
 		# For lio, we first have to create a target device, then
 		# add it to the Target Portal Group as an LU.
 		ocf_run targetcli /backstores/block create name=${OCF_RESOURCE_INSTANCE} dev=${OCF_RESKEY_path} || exit $OCF_ERR_GENERIC
@@ -440,6 +444,8 @@ iSCSILogicalUnit_stop() {
 		fi
 		;;
 	lio-t)
+		ocf_take_lock $TARGETLOCKFILE
+		ocf_release_lock_on_exit $TARGETLOCKFILE
 		# "targetcli delete" will fail if the LUN is already
 		# gone. Log a warning and still push ahead.
 		ocf_run -warn targetcli /iscsi/${OCF_RESKEY_target_iqn}/tpg1/luns delete ${OCF_RESKEY_lun}
--- a/heartbeat/iSCSITarget
+++ b/heartbeat/iSCSITarget
@@ -50,6 +50,9 @@ OCF_RESKEY_portals_default="0.0.0.0:3260"
 
 # Lockfile, used for selecting a target ID
 LOCKFILE=${HA_RSCTMP}/iSCSITarget-${OCF_RESKEY_implementation}.lock
+
+# targetcli: iSCSITarget and iSCSILogicalUnit must use the same lockfile
+TARGETLOCKFILE=${HA_RSCTMP}/targetcli.lock
 #######################################################################
 
 meta_data() {
@@ -334,6 +337,8 @@ iSCSITarget_start() {
 		# number 1. In lio, creating a network portal
 		# automatically creates the corresponding target if it
 		# doesn't already exist.
+		ocf_take_lock $TARGETLOCKFILE
+		ocf_release_lock_on_exit $TARGETLOCKFILE
 		ocf_run targetcli /iscsi set global auto_add_default_portal=false || exit $OCF_ERR_GENERIC
 		ocf_run targetcli /iscsi create ${OCF_RESKEY_iqn} || exit $OCF_ERR_GENERIC
 		for portal in ${OCF_RESKEY_portals}; do
@@ -499,6 +504,8 @@ iSCSITarget_stop() {
 		ocf_run lio_node --deliqn ${OCF_RESKEY_iqn} || exit $OCF_ERR_GENERIC
 		;;
 	lio-t)
+		ocf_take_lock $TARGETLOCKFILE
+		ocf_release_lock_on_exit $TARGETLOCKFILE
 		ocf_run targetcli /iscsi delete ${OCF_RESKEY_iqn} || exit $OCF_ERR_GENERIC
 		;;
 	esac


dmuhamedagic:
Looks good.

Just a few suggestions/questions:

Is this targetcli issue documented anywhere?

The subject line (commit summary) should be a bit improved and
also follow the usual format. Say:

Mid: iSCSILogicalUnit/iSCSITarget: protect targetcli invocations with locks

Thanks!

dmartinb:
All I've found is this:
https://www.spinics.net/lists/target-devel/msg07464.html
https://bugs.launchpad.net/cinder/+bug/1460692

The issue is targetcli-rtslib writes the target configuration to a configfs filesystem with no synchronized access. This produces unexpected results with multiple resources.

dmuhamedagic:
On Thu, Feb 16, 2017 at 01:38:50PM -0800, dmartinb wrote:
> All I've found is this:
> https://www.spinics.net/lists/target-devel/msg07464.html
> https://bugs.launchpad.net/cinder/+bug/1460692
> 
> The issue is targetcli-rtslib writes the target configuration to a configfs filesystem with no synchronized access. This produces unexpected results with multiple resources.

Thank you for the links and explanation. It took me a while to
digest the information. Not that I have yet, but we can merge
this. It's a pity that such important information is only
available incidentally.

lge:
Note: ocf_take_lock is NOT actually safe to use.

As implemented, it uses "echo $pid > lockfile" to create the lockfile,
which means if several such "ocf_take_lock" happen at the same time,
they all "succeed", only the last one will be the "visible" one to future waiters.

Maybe we should change it to 
```
while ! ( set -C; echo $pid > lockfile ); do
    if test -e lockfile ; then
        : error handling for existing lockfile, stale lockfile detection
    else
        : error handling for not being able to create lockfile
    fi
done
: only reached if lockfile was successfully created
```

(or use flock or other tools designed for that purpose)

In this case, actually it would be best to fix the lio tools.

dmuhamedagic:
On Wed, Jun 07, 2017 at 05:52:33AM -0700, Lars Ellenberg wrote:
> Note: ocf_take_lock is NOT actually safe to use.
> 
> As implemented, it uses "echo $pid > lockfile" to create the lockfile,
> which means if several such "ocf_take_lock" happen at the same time,
> they all "succeed", only the last one will be the "visible" one to future waiters.

Ugh.

> Maybe we should change it to 
> ```
> while ! ( set -C; echo $pid > lockfile ); do
>     if test -e lockfile ; then
>         : error handling for existing lockfile, stale lockfile detection
>     else
>         : error handling for not being able to create lockfile
>     fi
> done
> : only reached if lockfile was successfully created
> ```
> 
> (or use flock or other tools designed for that purpose)

flock would probably be the easiest. mkdir would do too, but for
upgrade issues.

> In this case, actually it would be best to fix the lio tools.

Seems to be a long standing issue, I guess that somebody already
raised it before.

lge:
Repost to a wider audience, to raise awareness for this.
ocf_take_lock may or may not be better than nothing.

It at least "annotates" that the auther would like to protect something
that is considered a "critical region" of the resource agent.

At the same time, it does NOT deliver what the name seems to imply.

I think I brought this up a few times over the years, but was not noisy
enough about it, because it seemed not important enough: no-one was
actually using this anyways.

But since new usage has been recently added with
[ClusterLabs/resource-agents] targetcli lockfile (#917)
here goes:

On Wed, Jun 07, 2017 at 02:49:41PM -0700, Dejan Muhamedagic wrote:
> On Wed, Jun 07, 2017 at 05:52:33AM -0700, Lars Ellenberg wrote:
> > Note: ocf_take_lock is NOT actually safe to use.
> > 
> > As implemented, it uses "echo $pid > lockfile" to create the lockfile,
> > which means if several such "ocf_take_lock" happen at the same time,
> > they all "succeed", only the last one will be the "visible" one to future waiters.
> 
> Ugh.

Exactly.

Reproducer:
#############################################################
#!/bin/bash
export OCF_ROOT=/usr/lib/ocf/ ;
.  /usr/lib/ocf/lib/heartbeat/ocf-shellfuncs ;

x() (
	ocf_take_lock dummy-lock ;
	ocf_release_lock_on_exit dummy-lock  ;
	set -C;
	echo x > protected && sleep 0.15 && rm -f protected || touch BROKEN;
);

mkdir -p /run/ocf_take_lock_demo
cd /run/ocf_take_lock_demo
rm -f BROKEN; i=0;
time while ! test -e BROKEN; do
	x &  x &
	wait;
	i=$(( i+1 ));
done ;
test -e BROKEN && echo "reproduced race in $i iterations"
#############################################################

x() above takes, and, because of the () subshell and
ocf_release_lock_on_exit, releases the "dummy-lock",
and within the protected region of code,
creates and removes a file "protected".

If ocf_take_lock was good, there could never be two instances
inside the lock, so echo x > protected should never fail.

With the current implementation of ocf_take_lock,
it takes "just a few" iterations here to reproduce the race.
(usually within a minute).

The races I see in ocf_take_lock:
"creation race":
	test -e $lock
		# someone else may create it here
	echo $$ > $lock
		# but we override it with ours anyways

"still empty race":
	test -e $lock	# maybe it already exists (open O_CREAT|O_TRUNC)
			# but does not yet contain target pid,
	pid=`cat $lock` # this one is empty,
	kill -0 $pid    # and this one fails
	and thus a "just being created" one is considered stale

There are other problems around "stale pid file detection",
but let's not go into that minefield right now.

> > Maybe we should change it to 
> > ```
> > while ! ( set -C; echo $pid > lockfile ); do
> >     if test -e lockfile ; then
> >         : error handling for existing lockfile, stale lockfile detection
> >     else
> >         : error handling for not being able to create lockfile
> >     fi
> > done
> > : only reached if lockfile was successfully created
> > ```
> > 
> > (or use flock or other tools designed for that purpose)
> 
> flock would probably be the easiest. mkdir would do too, but for
> upgrade issues.

and, being part of util-linux, flock should be available "everywhere".

but because writing "wrappers" around flock similar to the intended
semantics of ocf_take_lock and ocf_release_lock_on_exit is not easy
either, usually you'd be better of using flock directly in the RA.

so, still trying to do this with shell:

"set -C" (respectively set -o noclober):
	If set, disallow existing regular files to be overwritten
	by redirection of output.

normal '>' means: O_WRONLY|O_CREAT|O_TRUNC,
set -C '>' means: O_WRONLY|O_CREAT|O_EXCL

using "set -C ; echo $$ > $lock" instead of 
"test -e $lock || echo $$ > $lock"
gets rid of the "creation race".

To get rid of the "still empty race",
we'd need to play games with hardlinks:

(
set -C
success=false
if echo $$ > .$lock ; then
	ln .$lock $lock && success=true
	rm -f .$lock
fi
$success
)

That should be "good enough",
and much better than what we have now.



back to a possible "flock" wrapper,
maybe something like this:

ocf_synchronize() {
	local lockfile=$1
	shift
	(
		flock -x 8 || exit 1
		( "$@" ) 8>&-
	) 8> "$lockfile"
}
# and then
ocf_synchronize my_exclusive_shell_function with some args

As this runs in subshells,
it would not be able to change any variales visible
to the rest of the script, which may limit its usefulness.


or maybe, have a ocf_rarun_synchronized(),
that will do (directly from the flock man page):
   [ "${FLOCKER}" != "$0" ] && exec env FLOCKER="$0" flock -e "$0" "$0" "$@" || :
with nonblocking and timeout variants?


I'd very much like some volunteer to step forward
and implement the actual patch...


Cheers,

	Lars

dmuhamedagic:
Thanks for the analysis. I'm also not sure how to handle the
stale lock files. Perhaps to skip the "empty pidfile" case as
normally it should never happen. Been tempted to turn to C as
anyway the whole business is around system or libc calls, but I'm
not sure whether that would help either.

dmuhamedagic:
I managed to make something to fix this, will make a pr soon.

dmartinb:
In order to use flock, the (ocf_take_lock, ocf_release_lock_on_exit) approach doesn't seem to work because the lock is dropped when the block exits. But maybe inserting this piece of code directly in the RA instead...

```bash
# prepare exclusive lock
local lockfile=/var/lock/dummy.lock
local lockfd=200
local rnd=$(ocf_maybe_random)

eval "exec $lockfd>$lockfile"

until flock -n $lockfd
do
   ocf_log info "Sleeping until $lockfile is released..."
   sleep 0.$rnd
done
# critical section

# exit block: end of critical section
```

It seems clean and reliable. What do you think?

PR #920, opened 2017-01-27
Adds support for fstype=zfs on Filesystem resource agent

pruiz:
This is requiered when using a zfs dataset with mountpoint=legacy in
order to mount/umount such dataset as a cluster resource.


oalbrigt:
Thanks.

PR #923, opened 2017-02-01
galera: revert "long SST monitoring" PR #684 and #762

dciabrin:
The current monitoring of long running SST does not work as expected.
A new monitoring approach is being implemented to tackle the issue;
until then, revert these PR.

Revert tested with ra-tester


oalbrigt:
Thanks.

PR #924, opened 2017-02-01
Medium: exportfs: Only strip brackets from edges of clientspec

krig:
regexps are valid inputs to clientspec (like `node[0-2].example.com`), which breaks if internal brackets are stripped.

Technically, removing from the beginning and end also breaks some regexps, but at least this should help some cases.

### DIFF
--- a/heartbeat/exportfs
+++ b/heartbeat/exportfs
@@ -266,11 +266,14 @@ is_exported() {
 
 exportfs_monitor ()
 {
+	local spec
+
 	if ! ha_pseudo_resource "${OCF_RESOURCE_INSTANCE}" monitor; then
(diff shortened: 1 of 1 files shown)


oalbrigt:
There's something wrong with the substitution:
$ OCF_RESKEY_clientspec="[[ddf::fff:ddd]]"
$ echo "${${OCF_RESKEY_clientspec#[}%]}"
bash: ${${OCF_RESKEY_clientspec#[}%]}: bad substitution

krig:
Bah, shell.. OK, I'll try to fix it :)

krig:
Alright, it seems to work now.

oalbrigt:
Thanks.

PR #927, opened 2017-02-05
Fix Route for IPv6 default gateway

tbe:
The IPv6 ressource checks the "destination" paramter for a ":" to decide if
the route is IPv4 or IPv6. If the destination is "default", it always assumes that
IPv4 is given, even it the gateway has a ":" in it.

The "gateway" parameter is more failsave for this, so we changed the logic to it.

### DIFF
--- a/heartbeat/Route
+++ b/heartbeat/Route
@@ -291,7 +291,7 @@ done
 
 route_validate || exit $?
 
-case $OCF_RESKEY_destination in
+case $OCF_RESKEY_gateway in
 *:*) addr_family="-6" ;;
   *) addr_family="-4" ;;
 esac


dmuhamedagic:
Obviously the "default" case wasn't covered correctly. But what
if the gateway is not specified (the route could also be via a
device)? Could the user provide a name instead of an IP address
for gateway? Would perhaps something like this be better:

case "$OCF_RESKEY_destination $OCF_RESKEY_gateway" in
*:*) addr_family="-6" ;;
  *) addr_family="-4" ;;
esac

tbe:
if names are a supported usecase, we would need an additional parameter like "addr_family".

If both, the destination and the gateway are names, we can not decide if the route is ipv4 or ipv6.

dmuhamedagic:
Yes, obviously.

However, you neither responded to other questions nor commented
on the proposed alternative.

CAPSLOCK2000:
This wil fail if destination is an aliased network interface, eg 'ip route add default dev eth0:4'

PR #928, opened 2017-02-09
Mid: pgsql: Change to crm_resource.

HideoYamauchi:
Hi All,

From the next release of Pacemaker, user and RA cannot appoint INFINTIY by crm_failcount command.

 * https://github.com/ClusterLabs/pacemaker/commit/95db10602e8f646eefed335414e40a994498cafd

I abolish the use of crm_failcount.
By this correction, crm_resource -U(--clear) is necessary for restoration from trouble.

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1064,7 +1064,7 @@ pgsql_pre_promote() {
                           sort | head -1`
             if [ "$cmp_location" != "$my_master_baseline" ]; then
                 ocf_exit_reason "My data is newer than new master's one. New master's location : $master_baseline"
-                exec_with_retry 0 $CRM_FAILCOUNT -r $OCF_RESOURCE_INSTANCE -U $NODENAME -v INFINITY
+                exec_with_retry 0 $CRM_RESOURCE -B -r $OCF_RESOURCE_INSTANCE -N $NODENAME -Q
                 return $OCF_ERR_GENERIC
             fi
         fi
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
Hi Hideo-san,

On Thu, Feb 09, 2017 at 03:31:29PM -0800, Hideo Yamauchi wrote:
> Hi All,
> 
> I abolish the use of crm_failcount.

I can vagely recall a discussion at the ML, but not details. It
would be good to give an explanation.

HideoYamauchi:
Hi Dejan,

> I can vagely recall a discussion at the ML, but not details. It
> would be good to give an explanation.

I discussed this in the next ML.
 * http://lists.clusterlabs.org/pipermail/users/2017-February/004958.html

The crm_resource -U command becomes the workaround when the next error was given.

　　"My data is newer than new master's one. New master's location : xxxxxxxxx"

Should I include an example by comment in RA?

Best Regards,
Hideo Yamauchi.

dmuhamedagic:
Hi Hideo-san,

Very sorry for the delay.

On Mon, Feb 13, 2017 at 12:51:42AM -0800, Hideo Yamauchi wrote:
> Hi Dejan,
> 
> > I can vagely recall a discussion at the ML, but not details. It
> > would be good to give an explanation.
> 
> I discussed this in the next ML.
>  * http://lists.clusterlabs.org/pipermail/users/2017-February/004958.html
> 
> The crm_resource -U command becomes the workaround when the next error was given.
> 
> 　　"My data is newer than new master's one. New master's location : xxxxxxxxx"
> 
> Should I include an example by comment in RA?

I read the thread and there are quite a few issues discussed.
I'd suggest to add more comments to the RA, perhaps something
like this:

# We used to set the failcount to INF for the resource here in
# order to move the master to the other node. However, setting
# the failcount should be done only by the CRM and so this use
# got deprecated in pacemaker version x.y. Now we do the
# "ban resource from the node".

Hopefully the comment is semantically good too :)

HideoYamauchi:
Hi Dejan,

Thank you for comment.

> I'd suggest to add more comments to the RA, perhaps something
like this:
>  #We used to set the failcount to INF for the resource here in
>  #order to move the master to the other node. However, setting
>  #the failcount should be done only by the CRM and so this use
>  #got deprecated in pacemaker version x.y. Now we do the
>  #"ban resource from the node".

Okay!

I contribute the correction including your example again by the weekend.

Best Regards,
Hideo Yamauchi.

HideoYamauchi:
This is closed to perform new pullrequest.

PR #929, opened 2017-02-12
Implement GTID based MariaDB Master/Slave resource-agent

pyssling:
This is a fork of the mysql resource-agent for use with MariaDB
with GTID support. This simplifies a number of steps in failover
scenarios as slaves no longer require notification of the log
positions in the master.

This GTID could in theory have been integrated into the mysql
agent, but as MariaDB and MySQL are diverging it seems wise
to likewise fork the resource-agents.

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -127,6 +127,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_lxc.7 \
                           ocf_heartbeat_lxd-info.7 \
                           ocf_heartbeat_machine-info.7 \
+                          ocf_heartbeat_mariadb.7 \
                           ocf_heartbeat_mysql.7 \
                           ocf_heartbeat_mysql-proxy.7 \
                           ocf_heartbeat_nagios.7 \
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -123,6 +123,7 @@ ocf_SCRIPTS	     =  AoEtarget		\
 			lxc			\
 			lxd-info		\
 			machine-info		\
+			mariadb			\
 			minio			\
 			mysql			\
 			mysql-proxy		\
--- a/heartbeat/README.mariadb.md
+++ b/heartbeat/README.mariadb.md
@@ -1,0 +1,156 @@
+Setting up the MariaDB resource agent
+=====================================
+
+This resource agent requires corosync version >= 2 and mariadb version > 10.2 .
+
+Before embarking on this quest one should read the MariaDB pages on replication
+and global transaction IDs, GTID. This will greatly help in understanding what
+is going on and why.
+
+Replication: https://mariadb.com/kb/en/mariadb/setting-up-replication/
+GTID: https://mariadb.com/kb/en/mariadb/gtid/
+semi-sync: https://mariadb.com/kb/en/mariadb/semisynchronous-replication/
+
+Some reading on failures under enhanced semi-sync can be found here:
+https://jira.mariadb.org/browse/MDEV-162
+
+Part 1: MariaDB Setup
+---------------------
+
+It is best to initialize your MariaDB and do a failover before trying to use
+Pacemaker to manage MariaDB. This will both verify the MariaDB configuration
+and help you understand what is going on.
+
+###Configuration Files
+
+In your MariaDB config file for the server on node 1, place the following
+entry (replacing my_database and other names as needed):
+```
+[mariadb]
+log-bin
+server_id=1
+log-basename=master
+binlog_do_db=my_database
+```
+
+Then for each other node create the same entry, but increment the server_id.
+
+###Replication User
+
+Now create the replication user (be sure to change the password!):
+```
+GRANT ALL PRIVILEGES ON *.* TO 'slave_user'@'%' IDENTIFIED BY 'password';
+GRANT ALL PRIVILEGES ON *.* TO 'slave_user'@'localhost' IDENTIFIED BY 'password';
+```
+
+The second entry may not be necessary, but simplified other steps. Change
+user name and password as needed.
+
+
+###Intialize from a database backup
+
+Initialize all nodes from an existing backup, or create a backup from the
+first node if needed:
+
+On the current database:
+```
+mysqldump -u root --master-data --databases my_database1 my_database2 > backup.sql
+```
+
+At the top of this file is a commented out line:
+SET GLOBAL gtid_slave_pos='XXXX...'
+
+uncomment his line.
+
+On all new nodes:
+```
+mysqldump -u root < backup.sql
+```
+
+###Initialize replication
+
+Choose a node as master, in this example node1.
+
+On all slaves, execute:
+```
+RESET MASTER;
+
+CHANGE MASTER TO master_host="node1", master_port=3306, \
+       master_user="slave_user", master_password="password", \
+       master_use_gtid=current_pos;
+
+SET GLOBAL rpl_semi_sync_master_enabled='ON', rpl_semi_sync_slave_enabled='ON';
+
+START SLAVE;
+
+SHOW SLAVE STATUS\G
+```
+
+In an ideal world this will show that replication is now fully working.
+
+Once replication is working, verify the configuration by doing some updates
+and verifying that they are replicated.
+
+Now try changing the master. On each slave perform:
+```
+STOP SLAVE
+```
+
+Choose a new master, node2 in our example. On all slave nodes execute:
+```
+CHANGE MASTER TO  master_host="node2", master_port=3306, \
+       master_user="slave_user", master_password="password", \
+       master_use_gtid=current_pos;
+
+START SLAVE;
+```
+
+And again, check that replication is working and changes are synchronized.
+
+
+Part 2: Pacemaker Setup
+-----------------------
+
+This is pretty straightforward. Example is using pcs.
+
+```
+# Dump the cib
+pcs cluster cib mariadb_cfg
+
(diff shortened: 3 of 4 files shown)


pyssling:
I plan to start looking at master election and semi-sync in the coming days. So this code can expect to change, I may drop support for async replication, depending on how the semisync implementation looks.

dmuhamedagic:
How do you justify a fork?

This patch gets rid of ocf_is_ms, i.e. it assumes that the RA is
to be run as master-slave. It also does
s/MySQL/MariaDB/;s/`(.*)`/$(\1)/. Once those changes are removed,
not that much remains to justify a completely new RA.

Note that a fork implies two RAs to maintain and at that they
share most of the code. That isn't for free.

krig:
I also think it would be preferrable to keep this in a single RA if possible, and to use boolean parameters to enable/disable particular features that may only be supported by MySQL or MariaDB. At least right now they seem similar enough. Another option might be to have a common implementation but separate agents that simply set different default values. So the MariaDB RA would call out to the base implementation but enable the MariaDB-specific features, and likewise for MySQL. What do you think about that?

pyssling:
@dmuhamedagic @krig I have updated this pull request now. Changes were quite substantial, please review. Ideally we should now have a fool-proof mariadb setup that doesn't lose data and relies completely on pacemaker for clustering.

kgaillot:
I'm not opposed to a single agent for mysql and mariadb (we could set a new option to select between the two, where the code needs to be different), but an equally valid alternative would be to put any shared code into heartbeat/mysql-common.sh. That approach especially makes sense if we think mysql and mariadb are likely to diverge more in the future.

I'd prefer the agent be named simply "mariadb" rather than "mariadb-semi-sync" -- if we eventually support other modes, it should be an option within the mariadb agent rather than a separate agent, since it's the same software being managed.

heartbeat/README.mariadb-semi-sync.md
```diff
@@ -0,0 +1,149 @@
...
+
+###Replication User
+
+Now create the replication user:
+```
+GRANT ALL PRIVILEGES ON *.* TO 'slave_user'@'%' IDENTIFIED BY 'password';
+GRANT ALL PRIVILEGES ON *.* TO 'slave_user'@'localhost' IDENTIFIED BY 'password';
+
+FLUSH PRIVILEGES;
+```
```
kgaillot: I'd explicitly mention that users should replace 'password' with their password, you'd be surprised how many people copy-and-paste without looking :-)

There's no need to flush privileges here -- that's a bit of voodoo handed down over the years that's almost never necessary. It's only needed when changing privileges for an account that is currently logged in, and you want the existing login session to get the new privileges.

heartbeat/README.mariadb-semi-sync.md
```diff
@@ -0,0 +1,149 @@
+Setting up the MariaDB Semi Sync resource agent
+===============================================
+
+Before embarking on this quest one should read the MariaDB pages on replication
+and global transaction IDs, GTID. This will greatly help in understanding what
+is going on and why.
```
kgaillot: It would be nice to include a link here. Since the docs tend to be version-specific, we're probably stuck with just linking to the top of the official docs, but that's better than nothing.

heartbeat/mariadb-semi-sync
```diff
@@ -0,0 +1,1064 @@
...
+END
+}
+
+# Convenience functions
+
+greater_than_equal_long()
+{
+    # there are values we need to compare in this script
+    # that are too large for shell -gt to process
+    local true=$(echo "$1 > $2" | bc)
```
kgaillot: Surpisingly, "local" isn't POSIX, and zsh (the default shell in Solaris) doesn't have it. The least-bad portable alternative is to simply use prefixed variable names in functions, e.g. gtel_true. (BTW I know many existing agents already use local, but we should avoid it when we can.)

heartbeat/mariadb-semi-sync
```diff
@@ -0,0 +1,1064 @@
...
+                  -s -N -e "show global variables like 'gtid_current_pos'" | cut -f 2)
+
+    # Ensure that we got somethine like a valid GTID
+    if ! echo $gtid | grep -q '-'; then
+        ocf_exit_reason "Unable to read GTID from MariaDB"
+        ocf_log err "Unable to read GTID from MariaDB"
+        return $OCF_ERR_GENERIC
+    fi
+
+    ${HA_SBIN_DIR}/attrd_updater -p -n ${OCF_RESOURCE_INSTANCE}-gtid -U $gtid
```
kgaillot: Not a problem, just an FYI: using attrd_updater with -Q or -p limits the agent to clusters using pacemaker with corosync 2. Older cluster stacks use a "legacy attrd" that doesn't support those options. And of course this agent is pacemaker-specific (a few other software packages can use OCF agents). It might be worthwhile to document these limitations in the README and/or the meta-data.

heartbeat/mariadb-semi-sync
```diff
@@ -0,0 +1,1064 @@
...
+    sed -ne "s/^.* $1: \(.*\)$/\1/p" < $2
+}
+
+get_slave_info() {
+    # Warning: this sets $tmpfile and LEAVES this file! You must delete it after use!
+    local mysql_options
+            
+    if [ "$master_log_file" -a "$master_host" ]; then
+        # variables are already defined, get_slave_info has been run before
+        return $OCF_SUCCESS
```
kgaillot: tmpfile isn't set in this case. maybe it should explicitly set tmpfile="" and then the remover can check -n $tmpfile.

heartbeat/mariadb-semi-sync
```diff
@@ -0,0 +1,1064 @@
...
+
+    get_slave_info
+    rc=$?
+
+    if [ $rc -eq 0 ]; then
+
+        # Check normal errors
+        if [ $last_errno -ne 0 ]; then
+            ocf_exit_reason "MariaDB slave replication has failed ($last_errno): $last_error"
+
```
kgaillot: need to remove $tmpfile in the error blocks

heartbeat/mariadb-semi-sync
```diff
@@ -0,0 +1,1064 @@
...
+            fi
+
+        fi
+
+        if [ "$slave_sql" != 'Yes' ]; then
+            # We don't have a replication SQL thread running. Not a
+            # good thing. Try to recoved by restarting the SQL thread
+            # and remove reader vip.  Prevent MariaDB restart.
+            ocf_exit_reason "MariaDB Slave SQL threads currently not running."
+            ocf_log err "See $tmpfile for details"
```
kgaillot: I think it's less than ideal to leave $tmpfile lying around, and to assume it'll still be there when someone gets around to reading the logs. Better to log the contents of $tmpfile then remove it.

heartbeat/README.mariadb-semi-sync.md
pyssling: Will do.

heartbeat/README.mariadb-semi-sync.md
pyssling: Will amend.

heartbeat/mariadb-semi-sync
pyssling: I explicitly invoke the bash shell at the top, is that ok? I can change this though, but I would prefer to retain my bashisms :-)

heartbeat/mariadb-semi-sync
pyssling: This requires a very recent version of mariadb, so I think demanding a recent version of pacemaker is ok. :-) I will document both mariadb versions and pacemaker versions.

heartbeat/mariadb-semi-sync
pyssling: ok. Will change to that approach. Thank you.

pyssling:
@kgaillot  @krig  Regarding the single vs multiple resource agents I would say that these are two separate pieces of software and in the replication area they have diverged, even on the terminology level (though confusingly they both have GTID but GTID work differently and look different). I think it's important (or at least it is to me) that the reader when looking at the resource-agent can say with confidence what it does with the software it controls and ideally reproduce this on the command line. This is one of the strengths with the scripting approach compared to others.

@kgaillot About the n-1 . As you say, in general quorum should be enough, but say you run a 5 node cluster of mariadb (I can't see a good reason to,) and you lose 2 nodes, 1 master & 1 slave: Pacemaker should handle this, but mariadb might not as it currently only guarantees that 1 of the slaves has the acknowledged the transaction, which may be the lost slave. MySQL can actually tune this, forcing acknowledgement from >1 slaves, so there one could do some algebra based on the cluster size and tune that variable, but with mariadb we can't. :-/

pyssling:
Hi, here is an updated resource agent. I think this addresses most concerns except for the BASHisms.

kgaillot:
Changes look good to me, though I haven't tested the functionality. Explicitly specifying bash is fine. You may want to remove "See $tmpfile for details", the file won't be there anymore. FYI, the pacemaker dependency (for private attributes) is >=1.1.13.

pyssling:
Hi,

I fixed the last comment and have done some (non-exhaustive) testing. How can I proceed to get this merged?

pyssling:
@krig let me know if there is anything more I need to do to get this merged.

krig:
@pyssling It looks good to me now! Only one small thing:

Please add the agent to the list in `heartbeat/Makefile.am` as well as the one in `doc/man/Makefile.am` in order to integrate the agent into the automatic CI checks as well as manpage generation.

pyssling:
@krig this should all be good now, makefile changes in place.

krig:
Great, thanks!

PR #936, opened 2017-03-01
ocf_log: use same log format as pacemaker

oalbrigt:
Changed log format to be close to what Pacemaker uses (incl. updated HA_DATEFMT to be able to grep for a time and get info from both the resource agents and Pacemaker)


krig:
:+1:

oalbrigt:
No. I didnt know crmsh expected a specific log format.

This was to unify the logging format so you can search the logs with a date/time and get both data from the Pacemaker services and the resource agents instead of having to use one format for resource agents and another for the Pacemaker services.

krig:
In my opinion, I think it is worth some pain in the tools to improve the logging and make the formatting more uniform across pacemaker and the agents. If crmsh needs updates to the log matching because of this, it should be fixed there.

dmuhamedagic:
On Thu, Mar 02, 2017 at 07:47:42AM -0800, Oyvind Albrigtsen wrote:
> No. I didnt know crmsh expected a specific log format.

It is not only about crmsh. The logs are also an interface,
however weak, to the users and they sometimes do try to handle
messages with scripts. Such changes may break those scripts. I'm
not saying that this particular will, but that it may and we
really cannot know.

krig:
That's true, this is an interface breaking change. It is a bit unfortunate that we don't really have a `stable` vs. `development` branch of resource-agents.. :/

dmuhamedagic:
I can only say that it would be wonderful if upstream pacemaker had the same attitude.

PR #937, opened 2017-03-02
added kamctl parameter as it was missing, causing issues if kamctl is…

scream314:
… not in PATH


krig:
Hi! Make sure to add metadata for the `kamctl` parameter as well, right now it seems to be missing.

scream314:
Metadata added.

krig:
Thanks!

Issue #932, opened 2017-02-28
Galera: mysql can't write into the mktemp created file


SDV-jwe:
Hello all!

The galera resource agent has a problem determining the last cluster commit id because the mysql can't write to the temporary created file using mktemp.
For example:
local tmp=$(mktemp) is creating the temp file with root u+rw:
-rw-------   1 root    root      0 Feb 28 13:21 tmp.Q5LAsdR8EE

Now the mysqld_safe process, which is usually running under other user like "mysql" can't write it's log in this file!

Not sure if thats a normal issue on all distris but its a problem on our centos 6 machine.

This is my fix for that issue using the current master:

`--- galera.orig 2017-02-28 13:18:39.274026052 +0100
+++ galera      2017-02-28 13:18:46.530401104 +0100
@@ -519,7 +519,7 @@
     ocf_log info "attempting to detect last commit version by reading ${OCF_RESKEY_datadir}/grastate.dat"
     last_commit="$(cat ${OCF_RESKEY_datadir}/grastate.dat | sed -n 's/^seqno.\s*\(.*\)\s*$/\1/p')"
     if [ -z "$last_commit" ] || [ "$last_commit" = "-1" ]; then
-        local tmp=$(mktemp)
+        local tmp=$(sudo -u $OCF_RESKEY_user mktemp)

         # if we pass here because grastate.dat doesn't exist,
         # try not to bootstrap from this node if possible
`

krig:
Hi! This looks like a good fix to me. Would you mind creating a pull request with the suggested patch?

dciabrin:
Hey @juergenwsdv,
Thanks for reporting that, it seems that this is due to mariadb commit 8fcdd6b0 where mysqld_safe hardened the permission needed when writing to log files.
I was rather thinking of doing:

--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -520,6 +520,8 @@ detect_last_commit()
     last_commit="$(cat ${OCF_RESKEY_datadir}/grastate.dat | sed -n 's/^seqno.\s*\(.*\)\s*$/\1/p')"
     if [ -z "$last_commit" ] || [ "$last_commit" = "-1" ]; then
         local tmp=$(mktemp)
+        chown $OCF_RESKEY_user:$OCF_RESKEY_group $tmp
+        chmod 0640 $tmp
 
         # if we pass here because grastate.dat doesn't exist,
         # try not to bootstrap from this node if possible

[...]

PR #938, opened 2017-03-02
galera: fix permission of temporary log file for mariadb 10.1.21+

dciabrin:
Since MariaDB/server@8fcdd6b0ecbb966f4479856efe93a963a7a422f7,
mysqld_safe relies on a helper subprocess to write into log files.
This new logging mechanism expects log file to be writable by the
user configured to run mysqld.

Fix the generation of temporary log file accordingly.


krig:
See discussion in #932.

Looks good to me.

PR #939, opened 2017-03-03
galera: fix the first bootstrap when cluster has no data

dciabrin:
The resource agent selects the first node to go into Master state
based on the biggest commit version found on each node. If case no
data were written yet into the galera cluster, the current node is
selected as a "fallback" node to bootstrap the cluster.

The way the "fallback" node is selected is wrong because every node
takes a different decision, and this ultimately yields to 3
single-node galera clusters being started. To fix that, let the
"fallback" node be the last one in the wsrep_cluster_address, so that
the selection algorithm yields coherent results across nodes.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -451,14 +451,24 @@ pcmk_to_galera_name()
 detect_first_master()
 {
     local best_commit=0
-    local best_node="$NODENAME"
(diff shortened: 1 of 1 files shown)


oalbrigt:
Looks good. Thanks.

PR #941, opened 2017-03-06
clvm: remove reload action from metadata

oalbrigt:
Remove reload-action as it's announced, but not available in the resource agent.

Decided to remove it as some settings require a restart, while others does not, so it's better to keep it consistent, as it's possible to do the reload changes yourself while unmanage'ing the resource.


krig:
Makes sense :+1:

PR #943, opened 2017-03-13
Changes to support load-balanced/distributed NFS v3 exports 

scottmayhew:
These changes allow NFS clusters to be configured for distributed NFS v3 exports (or more accurately they allow NFS v3 clients to reclaim their locks after the failover of an NFS v3 export on such a cluster).  

In this configuration, the nfsserver and root exportfs resources are configured as clones, and the remaining resources are configured in one or more groups of lvm+fs+export+ip+nfsnotify resources.  

The two main changes are:
1. Upon takeover of an export, it's necessary to merge any data about monitored clients from a hidden directory on the export into the local /var/lib/nfs/statd/sm directory.  This is necessary so that clients who may have had locks on the old node will be notified to reclaim their locks from the new node.  See note below about the creation of those monitor records.
2. Since the nfsserver resource is configured as a clone, then it won't enter a grace period after an individual export is moved, and clients attempting to reclaim locks in response to a notification would receive NFSERR_NO_GRACE.  Sending lockd a kill signal will cause it to drop its locks and and enter a grace period.  The most logical place to do this is right before sending out the notifications.

The other changes are bugfixes.

**Note**:  In order for there to be any monitor records to actually merge, it's necessary to run rpc.statd with an HA callout.  I'm using the clunfslock script from rgmanager with a few minor fixes.  Since rgmanager is no longer maintained (AFAIK) it might be a good idea to include the clunfslock script with the resource-agents.

### DIFF
--- a/heartbeat/exportfs
+++ b/heartbeat/exportfs
@@ -13,14 +13,20 @@
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
+STATD_PATH="/var/lib/nfs/statd"
+
 # Defaults
 OCF_RESKEY_unlock_on_stop_default=1
 OCF_RESKEY_wait_for_leasetime_on_stop_default=0
 OCF_RESKEY_rmtab_backup_default=".rmtab"
+OCF_RESKEY_merge_notify_list_default=0
+OCF_RESKEY_state_subdir_default=".clumanager"
 
 : ${OCF_RESKEY_unlock_on_stop=${OCF_RESKEY_unlock_on_stop_default}}
 : ${OCF_RESKEY_wait_for_leasetime_on_stop=${OCF_RESKEY_wait_for_leasetime_on_stop_default}}
 : ${OCF_RESKEY_rmtab_backup=${OCF_RESKEY_rmtab_backup_default}}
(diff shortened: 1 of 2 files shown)


heartbeat/exportfs
```diff
@@ -311,6 +344,28 @@ export_one() {
...
+merge_notify_list()
+{
+	local dir=$1
+	local subdir=${OCF_RESKEY_state_subdir}
+
+	if [ ! -d $dir/$subdir/statd ]; then
+		mkdir -p $dir/$subdir/statd/sm
+		chown -R rpcuser.rpcuser $dir/$subdir/statd
+	fi
+	if selinuxenabled; then
```
oalbrigt: This should be changed to avoid running selinuxenabled on non-systemd systems, e.g.:
https://github.com/ClusterLabs/resource-agents/blob/805bfa13543d2356eb4b7e72c040f829a24a3c69/heartbeat/nfsserver#L198-L202

oalbrigt:
@krig, @dmuhamedagic: Please review this when you have the time, and give your thoughts on whether to include his modified "clunfslock" script from RGManager here, or how you would prefer that to be done.

krig:
@oalbrigt I'm on vacation next week without github access, if it still needs reviewing when I'm back I'll check it out then

scottmayhew:
Closing this for now due to some concerns over the changes.   From the NFS standpoint there are concerns that storing client state on the exports themselves leaves them open to manipulation by the root user on other clients.  From the HA standpoint, there are concerns that these changes could cause instability problems for existing users of the agents.

Issue #944, opened 2017-03-14
Description of OCF_ERR_ARGS is wrong


beekhof:
Correct description is at http://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/s-ocf-return-codes.html

> The resource’s configuration is not valid on this machine. E.g. it refers to a location not found on the node.

PR #945, opened 2017-03-14
Doc: Fix description of OCF_ERR_ARGS

krig:
Correct description taken from:

http://clusterlabs.org/doc/en-US/Pacemaker/1.1/html/Pacemaker_Explained/s-ocf-return-codes.html

Fixes: #944


oalbrigt:
Looks good. Thanks.

PR #948, opened 2017-03-17
Midumu: docker: fix to deal with the image name correctly.

ytakeshita:
Fix the following issue.
Now image_exists() can not parse the image names which has port number.
For exsample "xxx.xxx.xxx.xxx:YYYY/IMAGE:latest".

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -360,24 +360,31 @@ docker_stop()
 
 image_exists()
 {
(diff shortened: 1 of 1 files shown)


ytakeshita:
I confirmed that this code passed the following pattern.
https://github.com/docker/docker/blob/master/client/image_tag_test.go

krig:
Looks good to me.

PR #949, opened 2017-03-22
Add support for rndc options in named resource agent.

sebastic:
The `named` resources agent only suppports options for `named`, but no [`rndc` options](https://ftp.isc.org/isc/bind9/cur/9.11/doc/arm/man.rndc.html).

I needed these changes to support multiple instances.

The `rndc` & `rndc-confgen` options are not identical, so the options configured for the resource are only used for the `rndc` commands.


dmuhamedagic:
Thanks!

PR #950, opened 2017-03-23
oraasm: new resource agent for Oracle ASM Disk Groups

oalbrigt:


heartbeat/oraasm
```diff
@@ -0,0 +1,184 @@
...
+
+oraasm_status() {
+	rc=$(oraasm_monitor)
+	return $rc
+}
+
+oraasm_validate_all() {
+	if [ x = "x$OCF_RESKEY_home" ]; then
+		ocf_exit_reason "home not set"
+		return $OCF_ERR_INSTALLED
```
dmuhamedagic: This should be OCF_ERR_CONFIGURED.

heartbeat/oraasm
```diff
@@ -0,0 +1,184 @@
...
+	validate-all
+	methods
+	meta-data
+	usage
+	!
+}
+
+oraasm_getconfig() {
+    [ x = "x$OCF_RESKEY_home" ] &&
+        OCF_RESKEY_home=`awk -F: "/^+$OCF_RESKEY_diskgroup:/"'{print $2}' /etc/oratab`
```
dmuhamedagic: Looks like you have indentation with spaces here.

heartbeat/oraasm
```diff
@@ -0,0 +1,184 @@
...
+        return $OCF_SUCCESS
+    fi
+
+    ocf_run -q /etc/init.d/ohasd start
+
+    while ! oraasm_monitor; do
+        sleep 1
+    done
+
+    if [ $? -eq 0 ]; then
```
dmuhamedagic: This test is superfluous. You won't get here unless the monitor succeeds.

oalbrigt:
Updated based on feedback from @dmuhamedagic.

Yeah, I did check how ohasd works.

heartbeat/oraasm
```diff
@@ -0,0 +1,179 @@
...
+<content type="string" default="" />
+</parameter>
+
+</parameters>
+
+<actions>
+<action name="start" timeout="120" />
+<action name="stop" timeout="120" />
+<action name="status" timeout="60" />
+<action name="monitor" depth="0" timeout="30" interval="10" start-delay="10" />
```
dmuhamedagic: Forgot to take a look at the operation timeouts. The timeouts are the minimum recommended values for some typical setup. Put differently, they should be good for most deployments. But they should not be too big either. Are these timeouts just a copy from another RA?

dmuhamedagic: Also, why "start-delay"?

oalbrigt: Yeah. They are copied from another agent, so I'll update them to 60 for start/stop and 30 for status/monitor and remove the start-delay.

PR #953, opened 2017-03-23
pgsql: fix regex in set async mode

gdecicco:
This will fix a bug where pgsql resource agent can't switch a node from sync to async mode because of wrong check.
Bug introduced in commit 6e91193

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1479,7 +1479,7 @@ set_async_mode_all() {
 }
 
 set_async_mode() {
-    cat $REP_MODE_CONF |  grep -q -e "[,' ]$1[,' ]"
+    cat $REP_MODE_CONF |  grep -q -E "(\"$1\")|([,' ]$1[,' ])"
     if [ $? -eq 0 ]; then
         ocf_log info "Setup $1 into async mode."
         runasowner -q err "echo \"synchronous_standby_names = ''\" > \"$REP_MODE_CONF\""


heartbeat/pgsql
```diff
@@ -1479,7 +1479,7 @@ set_async_mode_all() {
 }
 
 set_async_mode() {
-    cat $REP_MODE_CONF |  grep -q -e "[,' ]$1[,' ]"
+    cat $REP_MODE_CONF |  grep -q -e "\"$1\""
```
oalbrigt: I think you should change it to "[,'\" ]$1[,'\" $]" to be backwards compatible.

ytakeshita: I think that the backward compatibility is kept with` \"$1\"`.
The pattern `[,' ]$1[,' ]` is maybe to ensure uniqueness of $1. 

ex) if $1 is test, test matches both.
```
synchronous_standby_names = 'test, testtest'
```
Now, pgsql RA sets " " to all elements of synchronous_standby_names regardless of PostgreSQL version, and the pattern `\"$1\"` satisfies uniqueness.
```
synchronous_standby_names = '"test", "testtest"'
```

oalbrigt: echo "'1, 2'" | grep -e "\"$1\"" doesnt return any matches, so it'll probably break in configurations with single-quotes.

gdecicco: As reported in the postgresql [documentation](https://www.postgresql.org/docs/9.6/static/runtime-config-replication.html): 
> Each standby_name should have the form of a valid SQL identifier, unless it is *. 

and for an [identifier](https://www.postgresql.org/docs/9.1/static/sql-syntax-lexical.html):
> Quoted identifiers can contain any character, except the character with code zero. (To include a double quote, write two double quotes.) This allows constructing table or column names that would otherwise not be possible, such as ones containing spaces or ampersands.

With your suggestion you will match something like "don't match this" searching for "don't" , "match" or "this"
In my opinion the right backwards compatible change should be:
`grep -E "(\"$1\")|([,' ]$1[,' ])"`

gdecicco: `"'1, 2'"` from psql point of view is a single identifier

ytakeshita: $1 is should set hostname. I think [', ] are invalid hostname characters, so there is probably no case that these characters are included inside " ".
But I think the pattern `grep -E "(\"$1\")|([,' ]$1[,' ])"` is good for backward compatible.
I confirmed with 2 hostname pattern of slave node with the pattern `grep -E "(\"$1\")|([,' ]$1[,' ])"`
1. valid SQL identifier
pm02
2. invalid SQL identifier
pm02.test

This pattern works fine both.

gdecicco: @oalbrigt if it's ok for you too, I will update the changes

gdecicco:
updated changes based on the previous review.

There is anything else that should be done?

kskmori:
Looks good to me. 
I'm going to merge this very soon since this is a critical issue for pgsql users (unless someone has an objection).

Thanks,

kskmori:
Merged.
Thanks for your contributions!

antaflos:
@gdecicco thank you for spotting and fixing this! I completely overlooked that part in my PR (#900) :(

PR #956, opened 2017-03-24
pgsql: allow dynamic membership

gdecicco:
You can now add new nodes to the cluster and they will be automatically included in the resource management.
Allowing to specify parameter node_list in the resource creation to keep compatibility with older version.

changes:
- node_list parameter is now optional: if not specified nodes list is asked to cluster itself
- replication slots are created on master during promotion rather than on all nodes during resource's starting
- RA only use qualified replication slot

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -614,6 +614,15 @@ pgsql_real_start() {
         ocf_log debug "PostgreSQL still hasn't started yet. Waiting..."
     done
 
+    # delete replication slots on all nodes. On master node will be created during promotion.
+    if use_replication_slot; then
+        delete_replication_slots
+        if [ $? -eq $OCF_ERR_GENERIC ]; then
+            ocf_exit_reason "PostgreSQL can't clean up replication_slot."
+            return $OCF_ERR_GENERIC
+        fi
+    fi
+
     ocf_log info "PostgreSQL is started."
     return $rc
 }
@@ -678,7 +687,7 @@ pgsql_promote() {
 
     # create replication slots on master before promotion
     if use_replication_slot; then
-        create_replication_slot
+        create_replication_slots
         if [ $? -eq $OCF_ERR_GENERIC ]; then
             ocf_exit_reason "PostgreSQL can't create replication_slot."
             return $OCF_ERR_GENERIC
@@ -1348,7 +1357,34 @@ create_replication_slot_name() {
     echo $replication_slot_name_list
 }
 
-create_replication_slot() {
+delete_replication_slot(){
+    DELETE_REPLICATION_SLOT_sql="SELECT pg_drop_replication_slot('$1');"
+    output=`exec_sql "$DELETE_REPLICATION_SLOT_sql"`
+    return $?
+}
+
+delete_replication_slots() {
+    local replication_slot_name_list
+    local replication_slot_name
+
+    replication_slot_name_list=`create_replication_slot_name`
+    ocf_log debug "replication slot names are $replication_slot_name_list."
+
+    for replication_slot_name in $replication_slot_name_list
+    do
+        if [ `check_replication_slot $replication_slot_name` = "1" ]; then
+            delete_replication_slot $replication_slot_name
+            if [ $? -eq 0 ]; then
+                ocf_log info "PostgreSQL delete the replication slot($replication_slot_name)."
+            else
+                ocf_exit_reason "$output"
+                return $OCF_ERR_GENERIC
+            fi
+        fi
+    done
+}
+
+create_replication_slots() {
     local replication_slot_name
     local replication_slot_name_list
     local output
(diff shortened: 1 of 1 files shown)


ytakeshita:
I think that the variable `e` is not defined, so this condition may be always true.
```
+    if [ $(echo $e | wc -w) -eq 0 ]; then ← here
+        # defaulting nodelist to all nodes in the cluster if not specified
+        NODE_LIST=$(crm_node --list | cut -d ' ' -f 2)
+    else
+        NODE_LIST=`echo $OCF_RESKEY_node_list | tr '[A-Z]' '[a-z]'`
+    fi
```

gdecicco:
absolutely it should be `OCF_RESKEY_node_list`

krig:
Since there were no further comments, I will merge this. Thanks!

ioguix:
Oops, I just found this PR while looking for some other information.

It seems to me you probably introduced a regression for at least CentOS 6. The output of `crm_node --list` is different depending on the HA stack used. See my previous discussion about this on the mailing list: http://lists.clusterlabs.org/pipermail/developers/2016-November/000385.html

Maybe we could find some logic, comparing the output with the local nodename and OCF_RESKEY_node_list to build an agnostic function...

Thoughts?

krig:
I don't think that counts as a regression since `crm_node --list` isn't called if the `node_list` parameter is provided (it was required before).

ioguix:
Indeed, it's not a regression, but at least it's a bug I suppose.

krig:
Hm, I wonder if something like this would work...

```
cibadmin -Ql | grep '<node ' | grep -Eo 'uname=\"[^"]+\"' | grep -Eo '\"[^"]+\"' | tr -d \" | sort | uniq
```

ioguix:
Last time I asked on #clusterlabs chan back in january, Ken Gaillot answered: 

> yes, you could get a list of known nodes from the CIB. it wouldn't tell you which one is the local node, and (depending on the underlying cluster layer, e.g. using multicast) it might not have nodes that have not yet started and joined the cluster. it will still have nodes that have simply left the cluster, but it may not have past nodes that have been removed from the configuration (e.g. by crm_node -R)

It narrows down the problem, and sounds far better than the current code around `crm_node`, but this is not a perfect and clean solution. It sounds fragile and I feels it might have some race-condition hiding around the corner.

It seems to me the best solution is to ask the clustering layer, as `crm_node` does...A wrapper around
it might do the job by comparing the cib with its output to detect the format...

krig:
hm, good points.. it sounds like the thing needed is a wrapper script around `crm_node` which handles the different output formats. :/

antaflos:
@gdecicco @krig I fear this change introduced a bug somewhere that leads to replication slots *not* being created upon master promotion. I am still investigating but our Vagrant-based two-node test setup no longer works with the changes from this PR (c3d545cd90854224eb729d68ec5a4ab337ac1bdc). The slave node complains that there is no replication slot by that name (in our case `repl_slot_pgnode02`), which is correct; no such slot exists on the master when querying `select * from pg_replication_slots;`.

Going back to the previous commit (6ad25cf64e00cebe5d90ec96430d94a38b240d31) makes it work again. I am currently trying to find where exactly this problem occurs.

antaflos:
@gdecicco Seems to me the problem is that when `restart_on_promote=true` is set (as we have above) the replication slots are created correctly upon promoting the master (in `pgsql_promote()`) but they are deleted shortly afterwards when PostgreSQL is restarted (in `pgsql_real_start()`). Setting `restart_on_promote=false` "fixes" this problem but that is not a real solution I fear.

I am not sure how to fix this properly as I am not that familiar with the mechanics and timing of promotion and replication slots but I think either the replication slots need to be (re)created after restating PostgreSQL, somewhere around here:

https://github.com/ClusterLabs/resource-agents/blob/c3d545cd90854224eb729d68ec5a4ab337ac1bdc/heartbeat/pgsql#L703-L710)

Or the replication slots should not be deleted in the first place in https://github.com/ClusterLabs/resource-agents/blob/c3d545cd90854224eb729d68ec5a4ab337ac1bdc/heartbeat/pgsql#L617-L624 when `${OCF_RESKEY_restart_on_promote} == true`.

Or just move the block creating the replications slots in https://github.com/ClusterLabs/resource-agents/blob/c3d545cd90854224eb729d68ec5a4ab337ac1bdc/heartbeat/pgsql#L688-L695 down to https://github.com/ClusterLabs/resource-agents/blob/c3d545cd90854224eb729d68ec5a4ab337ac1bdc/heartbeat/pgsql#L735-L740 maybe?

gdecicco:
@antaflos I am really sorry for this lack.

The deletion of slots was made to resolve a problem when the dead old master keeps the replication slots on restart (if cleanup was not done manually)

but I think that the best solution (not very beautiful) should be to recreation of slots after pgsql restart when `restart_on_promote==true` as you suggested. In the if branch just before the else (line 710)

Speaking about timing we are just delaying the slot creations. So if a slave is already connected to the new master (which is starting) it will simply not found the slot for a limited time.
I think that we should test to find out if this is a problem, but I don't think it should be

I am sorry but at the moment I can't test this. Can you please try it and let me know?

Thank you for spotting this.

antaflos:
I have finally created #1192 to address this problem.

PR #958, opened 2017-03-28
IPaddr2: fix gratuitious ARP checks

oalbrigt:
It didnt work at all with arp_bg=true (no logging), and didnt fail if send_arp failed.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -710,10 +710,14 @@ run_send_arp() {
 	if [ $ARP_COUNT -ne 0 ] ; then
 		ARGS="-i $OCF_RESKEY_arp_interval -r $ARP_COUNT -p $SENDARPPIDFILE $NIC $OCF_RESKEY_ip $MY_MAC not_used not_used"
 		ocf_log $LOGLEVEL "$SENDARP $ARGS"
-		if ocf_is_true $OCF_RESKEY_arp_bg; then
-			($SENDARP $ARGS || ocf_log err "Could not send gratuitous arps")& >&2
-		else
-			$SENDARP $ARGS || ocf_log err "Could not send gratuitous arps"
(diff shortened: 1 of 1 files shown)


lge:
This patch completely defeats the purpose of arp_bg, namely to do a "best effort", but return OK as soon as the IP is configured. Not that I think that arp_bg was particular useful, but if you disable it, rather patch out the option, and maybe log a warning if it is still set.

oalbrigt:
From my tests it seemed to use the ocf_log return code, so it always returned OK.

kskmori:
I would like to continue to use arp_bg=true, so not to delay to start consecutive resources. It would impact to the failover time.

I personally think that the start op is not necessarily failed only by the failure of send_arp, because it's just a "gratuitous" ARPs, at least when it runs in a background. 
Of course it would be great if the output from send_arp is logged to somewhere. (Just for a note: in the old 1.0, an error output from send_arp was logged via lrmd even if it's background, but in 1.1 something seems have been changed and nothing logged now)

oalbrigt:
I think this is caused by ocf-functions not being inherited in the subshell.

dmuhamedagic:
On Fri, Apr 07, 2017 at 12:53:47AM -0700, Oyvind Albrigtsen wrote:
> I think this is caused by ocf-functions not being inherited in the subshell.

I really doubt that that is so. Furthermore, if an undefined function is invoked then appropriate error message should show up at stderr.

kskmori:
I would suggest be6359589bf9add380333c84158a90ea00514ff5 to leave the output  to the log as a part of #977 .

kskmori:
I believe that this issue is fixed in b0abfa3d626e3eb2e1a40c2ea75267bfc1535831 as a part of #977 .  I'm closing this but if the issue is still there please repopen it. Thanks!

PR #960, opened 2017-03-29
kamailio: More parameters added

scream314:
Added a few more parameters (`kamctl`, `kamgroup`, `extra_options`) to be able to use this resource agent in place of an old resource agent and be able to run `kamailio` with the same command line switches.
(For example: there was a need to supply the `-m` option to `kamailio`, with this change options like this can be put in `extra_options`.)
Tried to maintain backward compatibility.


krig:
Looks good to me.

scream314:
Any chance to accept and close?

krig:
Since no one has objected, I'll merge. Thanks!

PR #961, opened 2017-04-04
send_arp: update usage info

oalbrigt:
I havent had the chance to test the libnet version, so please test that it's working fine as well.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -708,7 +708,7 @@ run_send_arp() {
 		LOGLEVEL=info
 	fi
 	if [ $ARP_COUNT -ne 0 ] ; then
-		ARGS="-i $OCF_RESKEY_arp_interval -r $ARP_COUNT -p $SENDARPPIDFILE $NIC $OCF_RESKEY_ip $MY_MAC not_used not_used"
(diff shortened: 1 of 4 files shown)


lge:
While at it, do you intend to do all of this (see https://github.com/ClusterLabs/resource-agents/pull/654#issuecomment-128015953)?

> We may consider changing send_arp.libnet.c calling conventions to match those of iputils arping instead, drop the ipoibarping special case from IPaddr2, change the calling convention in IPaddr2, and possibly add an optional preference on the platform arping over our own send_arp.
> Then deprecate send_arp and at some point just depend on iputils being installed and working.

kskmori:
I have been working on another approach to address all of those issues related to send_arp on my private repo. at https://github.com/kskmori/resource-agents/issues/6 .

The basic approach of mine is that:
 - allow to choose arping command instead of send_arp binary by a parameter, so it could be a default in the future.
 - keep the send_arp code untouched as much as possible to keep the backward compatibility easier with minimal efforts

It is almost done but I am still working on it for some issues.
I will post a pull request soon after I have finished it, so that we can discuss what is the best way.

oalbrigt:
@kskmori: sounds good. I have copied the old code to .old.c-files in case someone still need it.

kskmori:
This is fixed in a way as #977 .

PR #964, opened 2017-04-07
DB2: fix HADR support for DB2 V98+

oalbrigt:
"db2pd -hadr" output changed in DB2 V98.

Reference: search for "In DB2 V97 and earlier releases, the output is like:" at: https://www.ibm.com/developerworks/community/wikis/home?lang=en#!/wiki/DB2HADR/page/HADR%20monitoring

### DIFF
--- a/heartbeat/db2
+++ b/heartbeat/db2
@@ -650,7 +650,9 @@ db2_hadr_status() {
     fi
 
     echo "$output" |
-    awk '/^HADR is not active/ {print "Standard/Standalone"; exit; }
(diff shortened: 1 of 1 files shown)


oalbrigt:
Tested with DB2 v11.1 in HADR and standalone mode.

dmuhamedagic:
Looks good to me. Not an expert on db2 though, so better to defer merge to somebody else.

krig:
Looks good to me. Merging.

PR #965, opened 2017-04-13
galera: fix master target during promotion with cluster_host_map

dciabrin:
When option cluster_host_map is in use, it is assumed that galera node
names map to pacemaker node names _because_ those galera names are not
part of the pacemaker cluster in the first place.

This is not always the case (e.g. when using pacemaker bundles), so
fix accordingly.


oalbrigt:
Looks good to me. Thanks.

Issue #560, opened 2015-01-26
IPaddr2 IPv6 preferred_lft option


krig:
Received this feature request for the IPaddr2 resource agent:

> Configure IPv6 address with option preferred_lft 0. This is required to mark the virtual ip as "deprecated" source address and thus to use only physical addresses. 

> See http://www.davidc.net/networking/ipv6-source-address-selection-linux

I don't know if this is actually something that you would want to do by default, but would it make sense to add an option to set this on the virtual IP? It seems to mostly make sense for IPv6.

dmuhamedagic:
I don't find the preferred_lft option neither in the ip-address(8) man page nor in the rest of the iproute package. But I couldn't find an alternative way to set an address as deprecated.
@kskmori Keisuke-san, do you want to take a look?

kskmori:
It was once requested in #68 and the conclusion is that it's preferable to use address labels instead of preferred_lft 0 for that purpose. 

So the solution would be either:
* enable lvs_ipv6_addrlabel option (introduced by #158 9b1c28c)
* or add an address label in the system configuration by doing something like:
```
ip addrlabel add prefix fd00:172:20:25::243 label 99
```

krig:
Thank you for the clear explanation! Closing this.

PR #968, opened 2017-04-26
IPaddr2: add option for specifying IPv6's preferred_lft

dciabrin:
This change allows setting the preferred_lft option when creating an
IPv6 address. This can be used to ensure that the created IP address
will not be used as a source address for routing.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -56,6 +56,7 @@
 #	OCF_RESKEY_arp_count
 #	OCF_RESKEY_arp_bg
 #	OCF_RESKEY_arp_mac
+#	OCF_RESKEY_preferred_lft
 #
 #	OCF_RESKEY_CRM_meta_clone
 #	OCF_RESKEY_CRM_meta_clone_max
@@ -80,6 +81,7 @@ OCF_RESKEY_arp_count_refresh_default=0
 OCF_RESKEY_arp_bg_default=true
 OCF_RESKEY_arp_mac_default="ffffffffffff"
 OCF_RESKEY_run_arping_default=false
+OCF_RESKEY_preferred_lft_default="forever"
 
 : ${OCF_RESKEY_lvs_support=${OCF_RESKEY_lvs_support_default}}
 : ${OCF_RESKEY_lvs_ipv6_addrlabel=${OCF_RESKEY_lvs_ipv6_addrlabel_default}}
@@ -92,6 +94,7 @@ OCF_RESKEY_run_arping_default=false
 : ${OCF_RESKEY_arp_bg=${OCF_RESKEY_arp_bg_default}}
 : ${OCF_RESKEY_arp_mac=${OCF_RESKEY_arp_mac_default}}
 : ${OCF_RESKEY_run_arping=${OCF_RESKEY_run_arping_default}}
+: ${OCF_RESKEY_preferred_lft=${OCF_RESKEY_preferred_lft_default}}
 #######################################################################
 
 SENDARP=$HA_BIN/send_arp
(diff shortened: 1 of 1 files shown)


dciabrin:
The idea behind this patch is that when we create an IPv6 address we now unconditionally pass the preferred_lft flag to "ip addr create". preferred_lft defaults to "forever" which is the default in "ip" command, so the semantics of IPv6 creation is unchanged when not using that option.

heartbeat/IPaddr2
```diff
@@ -350,6 +353,17 @@ Whether or not to run arping for IPv4 collision detection check.
...
 </parameter>
 
+<parameter name="preferred_lft">
+<longdesc lang="en">
+For IPv6, set the preferred lifetime of the IP address.
+This can be used to ensure that the created IP address will not
+be used as a source address for routing.
+Expects a value as specified in section 5.5.4 of RFC 4862.
+</longdesc>
+<shortdesc lang="en">IPv6 preferred lifetime</shortdesc>
```
beekhof: Might be worth indicating that the contents is a value in seconds

kskmori:
Would you try `lvs_ipv6_addrlabel=true` option and see if it works for you?

Using address labels is considered an appropriate way to control the source address selection rather than using preferred_lft according to the RFC description. And the option is just for it.

See:
* https://github.com/ClusterLabs/resource-agents/issues/560#issuecomment-71996810
* https://github.com/kskmori/resource-agents/pull/2#issuecomment-9133141

oalbrigt:
The preferred_lft option is consistent with the keepalived method of configuring VIPs, and it might be usable in other scenarios as well.

mbaldessari:
Also preferred_lft takes precedence compared to addrlabel so will work no matter how the system is configured. Also having ip -6 a showing you the option is kind of nice tbh. I am not entirely sure why you think it is a no-no in terms of RFC compliance. You state "By the definition of those rules in the RFC, rule 3 is not suitable for our purpose. " which I don't really grok.
You also add:
  An implementation MAY prevent any new communication from using a
  deprecated address, but system management MUST have the ability to
  disable such a facility, and the facility MUST be disabled by default.

which really does not imply at all that deprecated is a) either going away or b) that the behaviour will change (as this facility *must* be disabled by default)

Deprecated addresses are defined in rfc 4862 and deprecated addresses are just valid addresses that may become invalid. A VIP fits that definition.

kskmori:
I'm not intended to be too strict on RFCs, but the definition of a deprecated address in RFC 4862 goes:
```
deprecated address -  An address assigned to an interface whose use
      is discouraged, but not forbidden. (snip)
```
Also RFC 4862 and this terminology is a part of IPv6 Stateless Address for the autoconfiguration process, I'd hesitate to extend this definition as an use case of VIPs. Maybe ok but not guaranteed.

A VIP is nothing more than a "secondary" address managed by an user application (from the kernel point of view), some users want to use the VIP as the source address, wheres others may want to use the "primary" address as the source address, depending on their use cases. The network stack already provides the way to customize those precedence exactly for that purpose, that is address labels and guaranteed to work. That's why I choose this method.

Anyway, I am not totally against to introduce this option if there is an use case that can not be covered by address labels. "Compatibility with keepalived" might be the case, though I would like to know what is exactly the problem of "imcompatibility".

mbaldessari:
Sure but rfc 4862 also states that 
  A deprecated address should be used only by applications that have been using it and would have
  difficulty switching to another address without a service disruption.

which matches our case. Well I think the arguments for adding this option are:
- clearly displayed by "ip -6 a" (addrlabels are not)
- by matching rule 3 you have more guarantees than a later rule 
- it is independent of the system's configured addrlabel (aka it works even on more complex setups without needing to change the addrlabel configuration on the system)
- it uses the same logic as keepalived which makes things more consistent for admins

kskmori:
Yes but that statements in RFC 4862 is talking about IPv6 stateless address autoconfiguration and it does not seem to me that covers the case of VIPs for failover clusters.

> *    it is independent of the system's configured addrlabel 

Yes, I'd agree on this. 

> *    it uses the same logic as keepalived which makes things more consistent for admins

Well, it is "ok" for me. I can understand that many users want the same way that they're familier with. But it reminds me that some users complained to migrate to IPaddr2 because they can not see the VIP with `ifconfig` :)

Thanks,

PR #969, opened 2017-05-02
docker: Allow callers to specify a set of directories that should be …

beekhof:
…created if they don't exist


oalbrigt:
Looks good. Thanks.

PR #970, opened 2017-05-03
Refactor ocf-shellfuncs.in/ocf_run

matelakat:
When calling `ocf_run` people expect that the function returns with the
exit status of the subprocess. In case of 0 exit code we returned
$OCF_SUCCESS which is 0 as well, but that doesn't read well. Refactoring
the return statements, so it is evident that the function always returns
with the exit code of the subprocess.


dmuhamedagic:
Thanks.

PR #972, opened 2017-05-03
Medium: SAPInstance: Add IS_ERS parameter (bsc#1036486)

krig:
* `IS_ERS`

If `IS_ERS` is true, mark a per cluster attribute for a
specific ASCS/ERS pair describing which node is the
best place to failover a failed ASCS.

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -31,6 +31,7 @@
 #	OCF_RESKEY_POST_START_USEREXIT	(optional, lists a script which can be executed after the resource is started)
 #	OCF_RESKEY_PRE_STOP_USEREXIT	(optional, lists a script which can be executed before the resource is stopped)
 #	OCF_RESKEY_POST_STOP_USEREXIT	(optional, lists a script which can be executed after the resource is stopped)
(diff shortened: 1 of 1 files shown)


oalbrigt:
The AUTOMATIC_UNMIGRATE should be available in client utilities, and not in resource agents (available as e.g. "pcs resource move <resource> [lifetime=<lifetime>]").

From "pcs resource move" usage info:
If lifetime is specified then the constraint will expire after that time, otherwise it defaults to infinity and the constraint can be cleared manually with 'pcs resource clear' or 'pcs constraint delete'.

krig:
Constraint lifetimes isn't a popular alternative since the constraint will remain in the configuration indefinitely, confusing people. Maybe scheduling a regular cleanup of such constraints could be an option...

krig:
@oalbrigt We decided to drop the second parameter and follow the alternative approach of constraint lifetime + scheduled cleanup of the CIB.

oalbrigt:
@krig: Great. Looks good to me.

krig:
Great. Will wait for a decision from the SAP developers before merging.

krig:
Merging based on request from Fabian (and the positive feedback so far).

PR #974, opened 2017-05-03
systemd: add resource-agents-deps target

oalbrigt:
clvmd and LVM will create drop-ins on start to avoid nodes being fence during shutdown or reboot.

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -426,6 +426,12 @@ LVM_start() {
 	local vg=$1
 	local clvmd=0
 
+	# systemd drop-in to stop process before storage services during
+	# shutdown/reboot
+	if ps -p 1 | grep -q systemd ; then
+		systemd_drop_in "99-LVM" "After" "blk-availability.service"
+	fi
+
 	# TODO: This MUST run vgimport as well
 	ocf_log info "Activating volume group $vg"
 	if [ "$LVM_MAJOR" -eq "1" ]; then
--- a/Makefile.am
+++ b/Makefile.am
@@ -37,7 +37,7 @@ RGMANAGER = with
 endif
 
 if BUILD_LINUX_HA
-SUBDIRS	+= include heartbeat tools ldirectord doc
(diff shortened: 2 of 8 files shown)


oalbrigt:
After= and Wants= will be added to pacemaker.service.

heartbeat/LVM
```diff
@@ -426,6 +426,18 @@ LVM_start() {
 	local vg=$1
 	local clvmd=0
 
+	# systemd drop-in to stop process before storage services during
+	# shutdown/reboot
+	if ps -p 1 | grep -q systemd ; then
```
beekhof: Looks like something we want a function for.

dmuhamedagic:
On Wed, May 03, 2017 at 09:32:36PM -0700, Andrew Beekhof wrote:
> > @@ -426,6 +426,18 @@ LVM_start() {
>  	local vg=$1
>  	local clvmd=0
>  
> +	# systemd drop-in to stop process before storage services during
> +	# shutdown/reboot
> +	if ps -p 1 | grep -q systemd ; then
> 
> Looks like something we want a function for.

IIRC, there's also a bunch of related tests in nfsserver. Best to
be done by somebody who's running systemd (I don't).

oalbrigt:
He meant the drop-in, so I have now made a systemd_drop_in function that is being used by the 2 agents.

dmuhamedagic:
Not sure what does "drop-in" mean. At any rate, what I meant is a function to test whether the system is running systemd. In nfsserver there's a test for systemctl existence which is not unequivocal (i.e. if systemctl exists doesn't necessarily mean that systemd is used to manage services).

oalbrigt:
drop-in is a systemd "override" file that can add After= or similar dependencies.

heartbeat/LVM
wferi: From the man page of `sd_booted()`:

> Internally, this function checks whether the directory /run/systemd/system/ exists. A simple check like this can also be implemented trivially in shell or any other language.

PR #975, opened 2017-05-08
Make ocf_is_ver() handle dist. specific versions

greemo:
Comment on this function says letters are ignored, but that was not the case. 
For example: 1.1.14-8.el6_8.1

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -511,7 +511,7 @@ ocf_is_ms() {
 # letters and such are effectively ignored
 #
 ocf_is_ver() {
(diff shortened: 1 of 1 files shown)


oalbrigt:
This makes it match longer versions as well, so it would be better to undo this and fix it where ocf_is_ver or ocf_version_cmp is used to strip .el and similar info.

krig:
Hmm, okay.

oalbrigt:
Reverting.

@greemo: please inform us where this is an issue, so it can be fixed properly instead of something that might match longer version numbers.

greemo:
Hi, I gave the example in the original issue comment: https://github.com/ClusterLabs/resource-agents/pull/975#issue-226976056

oalbrigt:
@greemo: which resource agent is failing the ocf_is_ver()-check for dist specific versions?

Issue #860, opened 2016-10-18
send_arp.linux ignores arguments


phemmer:
The [`send_arp` utility for linux](https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c) ignores the `src_hw_addr`, `broadcast_ip_addr`, and `netmask` arguments. This results in the utility sending out the wrong mac address when called by `IProute2` in clone (clusterip) mode.

https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c#L1105-L1115
^ note that the only parameters that are actually used are `device` and `src_ip_addr`.

Also, on an unrelated note, the usage documentation is completely wrong: https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c#L140-L150

lge:
> The [`send_arp` utility for
> linux](https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c)
> ignores the `src_hw_addr`, `broadcast_ip_addr`, and `netmask`
> arguments. This results in the utility sending out the wrong mac
> address when called by `IProute2` in clone (clusterip) mode.

Depends on exact configuration, but generally: yes, you are right.

> https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c#L1105-L1115
> ^ note that the only parameters that are actually used are `device` and `src_ip_addr`.

"Always been broken that way" :-(

> Also, on an unrelated note, the usage documentation is completely
> wrong:
> https://github.com/ClusterLabs/resource-agents/blob/master/tools/send_arp.linux.c#L140-L150

For the interested, let me add some historic context:
on "inception", heartbeat came with its own, homegrown, libnet based
send_arp tool (because apparently a working, cross-platform, reliably
available tools with similar functionality.

Its ancestor is still available as the "send_arp.libnet.c" variant.
Later, the libnet dependency was found to be undesirable (for whatever
reason), so on linux at least, the iputils arping was canibalized,
and put into send_arp.linux.c, with mangling of the options parsing,
and added "pidfile", so it could be used as a drop-in replacement of its
existing send_arp.libnet tool.

The usage() displayed is that of iputils arping,
but the way it is used by IPaddr2 is how send_arp.libnet.c was used.
It basically has two usages, but the project uses the one not documented.

Last time I touched IPaddr2, I suggested that we would instead
change the send_arp.libnet.c variant to follow the arping command line
[...]

PR #977, opened 2017-05-12
IPaddr2: fix send_arp issues and support arping command

kskmori:
This series of patches is to fix a bunch of issues related to send_arp so far (__Known issues:__ below) as well as to support `arping` command instead of send_arp.
This pull request should not change any behavior for the existing configuration (except for bug fixes).

I would appreciate your comments and feedback, particularly from Debian user (which has two versions of arping and I don't know how they are used), and from Cluster IP users (which I haven't used so much).

Regards,

### Known issues: ###
* need an option for  sending ARP REPLY (`arping -A`) 
  -  some switches does not accept ARP REQUEST(arping -U, send_arp) but do only ARP REPLY (arping -A)
* send_arp: arp_interval does not work
  - http://lists.clusterlabs.org/pipermail/users/2016-October/004180.html
* send_arp: src_hw_addr, broadcast_ip_addr, and netmask are not set properly which causes clusterip unavailable
  - ​https://github.com/ClusterLabs/resource-agents/issues/860
* send_arp: improper documentation
  - https://github.com/ClusterLabs/resource-agents/issues/860
* IPaddr2: arp_mac is no longer in use but still described in the metadata
  - removed with the commit 2852f51 and 0295fa1
* an output from send_arp is not logged when run in a background
  - https://github.com/ClusterLabs/resource-agents/pull/958

### Proposals to fix: ###
* extend `arp_sender` parameter to allow to choose the standard arping bundled in the distro. 
  - aims to replace send_arp binary with arping as default in the future
  - also considering a variant of arping command (iputils_arping, libnet_arping)
* add a new `send_arp_opts` parameter 
  - allow to send ARP REPLY instead of ARP REQUEST
* send_arp.{libnet, linux} are left only for the backward compatibility with minimal changes
  - basically fixes documentation errors only
  - the usage of send_arp.linux shows options that are currently used as a drop-in replacement of send_arp.libnet
* let `arp_interval` parameter as deprecated and document it
  - it does not seem to be used frequently and only limited tools support it
* do not send gratuitous ARPs in the Cluster IP configuration
  - it does not seem to be necessary because the cluster MAC will not change and nothing needs to be updated on the peers (point me if it's wrong)
  - sending a physical MAC does cause a problem 
* use ocf_log for the output from send_arp command

### Related duscussion: ###
* The great summary of the history by lge
  - ​https://github.com/ClusterLabs/resource-agents/issues/860#issuecomment-263335126
* suggestion of replacing send_arp with arping in the past
  - https://lists.gt.net/linuxha/pacemaker/58110#58110
  - https://github.com/ClusterLabs/resource-agents/issues/860#issuecomment-263335126
  - https://github.com/ClusterLabs/resource-agents/pull/654#issuecomment-128015953
* a different approach to fix send_arp usage
  - https://github.com/ClusterLabs/resource-agents/pull/961


heartbeat/IPaddr2
```diff
@@ -701,6 +703,23 @@ log_arp_sender() {
...
+    local cmdline
+    local pid
+    local rc
+
+    cmdline="$@"
+
+    $cmdline &
+    pid=$!
+    echo "$pid" > $SENDARPPIDFILE
+    wait $pid
```
dmuhamedagic: I'm a bit confused here. What's the purpose of running the program in the background when immediately we wait for it?

kskmori: This is to create and remove the pid file of arping as similar manner  as the send_arp binary does.
Running it in the background once is to store the pid of arping via $! to the pid file (which could be used later to clean up if it's still running when arp_bg=true), and waiting it here is to remove the pid file as soon as arping exits at the next line.

kskmori:
Rebased to the master so that it could be merged easily.
I'm going to merge this until  4.1.0-rc1 if there is no objection.

I'm wondering why Travis-CI fails and now looking into it. There is no issue here, neither before rebasing.

krig:
Hi, the CI failure is unrelated, currently all CI is failing due to a problem with the Travis base installation.

kskmori:
Oh, I see. Thanks!

PR #978, opened 2017-05-14
fix nginx resource

antblant:
try to restart resource on different node if nginx configtest fails


krig:
This matches my understanding of `OCF_ERR_CONFIGURED` vs. `OCF_ERR_INSTALLED`: `OCF_ERR_CONFIGURED` is meant to indicate problems with the CIB configuration, not configuration files per se.

PR #981, opened 2017-05-29
LVM: status check for missing VG

oalbrigt:
Detect if VG is missing (e.g. for multipath where the device is still present when it's unavailable).

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -320,6 +320,18 @@ LVM_status() {
 			fi
 		fi
 	fi
+
+	# Check if VG is still available (e.g. for multipath where the device
+	# doesn't disappear)
+	if [ "$LVM_MAJOR" -eq "1" ]; then
+		output=$(vgscan $vg 2>&1)
+	else
+		output=$(vgscan --cache 2>&1)
+	fi
+	if ! echo "$output" | grep -q "Found.*\"$1\""; then
+		ocf_exit_reason "LVM Volume $1 is not available"
+		return $OCF_ERR_GENERIC
+	fi
 	
 	if [ -d /dev/$1 ]; then
 		test "`cd /dev/$1 && ls`" != ""


heartbeat/LVM
```diff
@@ -320,6 +320,18 @@ LVM_status() {
...
 		fi
 	fi
+
+	# Check if VG is still available (e.g. for multipath where the device
+	# doesn't disappear)
+	if [ "$LVM_MAJOR" -eq "1" ]; then
+		output=$(vgscan $vg 2>&1)
+	else
+		output=$(vgscan --cache 2>&1)
+	fi
```
renzhengeek: I think 'vgscan --cache' is very expensive lvm command, especially if there are a large of LVs on system. LVM_status will be invoked periodically. I don't think it's good idea to ad it in status checking.

heartbeat/LVM
```diff
@@ -320,6 +320,18 @@ LVM_status() {
...
+	if ! echo "$output" | grep -q "Found.*\"$1\""; then
+		ocf_exit_reason "LVM Volume $1 is not available"
+		return $OCF_ERR_GENERIC
```
renzhengeek: It seems to me that  'status action' expects either OCF_SUCESS or OCF_NOT_RUNNING...

renzhengeek:
From my perspective, in status action, it's always good to use short/quick code to *accurately* check the LV/VG activation status directly :)

dmuhamedagic:
> I think 'vgscan --cache' is very expensive lvm command, especially if there are a large of LVs on system. LVM_status will be invoked periodically. I don't think it's good idea to ad it in status checking.

Actually, _all_ lvm commands were at one point in time removed from the status action exactly because timeouts were reported. The timeouts were apparently also due to volumes unrelated to the cluster. Note that they were not even vgscan, but something looking very innocuous. There should be a thread somewhere on the linux-ha ML.

oalbrigt:
Thanks for your feedback. I agree, and will close this PR.

PR #984, opened 2017-05-30
LVM: warn when cache mode is not writethrough

oalbrigt:


krig:
Haha, I don't think I've seen all-caps warning messages in OCF agents before. What motivated this change? :)

oalbrigt:
To improve the chances of users seeing it.

krig:
Isn't the chance greater if the agent fails with a clear exit reason, as before? That way, the tools will report it and the resource won't start.

krig:
Log messages in all-caps seems like an arms race kind of situation to me, I'm not sure I really get the situation where it is useful. To see it, a user would have to be digging through their logs line-by-line anyway.

oalbrigt:
The reason is that we dont want to have the parameter where users forget about it, and also avoid regression issues for users who might already use it.

krig:
I still find it strange:

* Why is this particular log warning the one that deserves to be all caps?
* Now there is a `check_writethrough` parameter which does nothing.

At the very least, the description for the `check_writethrough` parameter should be updated to state that it is deprecated and that writethrough is always checked.

Another option could be to change the `check_writethrough` parameter to a string, and to set the default to "warn":

* if the value of `check_writethrough` matches `ocf_is_true`, report the error
* if the value is `warn`, write the warning to the log
* if it is false, do nothing

oalbrigt:
Oh. I was supposed to remove the "check_writethrough" parameter but forgot, so I'll fix that.

oalbrigt:
Updated patch to remove the check_writethrough parameter as it's no longer needed.

krig:
OK, I see. That makes more sense now. Alright, I'll merge it.

lge:
A non "writethrough" configuration would lead to data corruption or unavailability after failover,
so I'd recommend to leave the "exit" in there (and lose the ALL CAPS again).

lge:
Also, lvs -o cache_mode is not available on older lvm2 tools (e.g. debian jessie).
Then again, those distributions are very unlikely to be used for any "cache" segtype, so that is probably ok.

Is my statement even correct, about "potential corruption or unavailability after failover"?
Or what exactly is the reason for the "UNSUPPORTED" warning?

oalbrigt:
The reason for not using "exit" is to avoid causing issues for people upgrading, hence the ALL CAPS in the hopes they'll notice it.

Good point regarding the lvs -o cache_mode. I'll try and find a better way to do this.

Unsupported is because it might cause data corruption if there's any issues with devices or similar, so it cant be supported by e.g. distributions.

PR #986, opened 2017-06-01
fs.sh: fix builds when srcdir and builddir are seperated

jackiehjm:
It fails to find fs.sh.in when srddir and builddir are not the same:

make[5]: Entering directory '/path/to/builddir/rgmanager/src/resources'
cat fs.sh.in | sed \
    -e 's#@''LOGDIR@#/var/log/cluster#g' \
    > fs.sh.out
    cat: fs.sh.in: No such file or directorychmod +x fs.sh.out
    mv fs.sh.out fs.sh

Add abs_srcdir to fix this.

Signed-off-by: Jackie Huang <jackie.huang@windriver.com>


oalbrigt:
Looks good to me. Thanks.

PR #987, opened 2017-06-07
Add support for LIO-T SCSI tpu,3pc and caw emulation attributes

### LINKED ISSUE #610: iSCSILogicalUnit - No way to set iblock Parameters
It appears there is no way to set the iblock parameters for iSCSILogicalUnit.
For example, if you want to set emulate_tpu=1, you can't!

This means you can't:
- Enable TRIM / Discard (emulate_tpu)
- Enable Compare and Write (emulate_caw)
- Enable Extended Write (emulate_3pc)

sammcj:
Fixes #610 

Ideally I'd make it work for all iSCSI subsystems that can support them, but that's a bit beyond the time I'm currently able spend on.

### DIFF
--- a/heartbeat/iSCSILogicalUnit
+++ b/heartbeat/iSCSILogicalUnit
@@ -147,6 +147,33 @@ The default is a hash of the resource name, truncated to 8 bytes.
 <content type="string" default="${OCF_RESKEY_scsi_sn_default}"/>
 </parameter>
 
+<parameter name="emulate_tpu" required="0" unique="0">
+<longdesc lang="en">
+The SCSI UNMAP command to be configured for this Logical Unit.
+Setting this integer to 1 will enable TPU IOCTL emulation.
+</longdesc>
+<shortdesc lang="en">SCSI UNMAP (for TRIM / DISCARD)</shortdesc>
+<content type="integer" />
+</parameter>
+
+<parameter name="emulate_3pc" required="0" unique="0">
+<longdesc lang="en">
+The SCSI EXTENDED COPY command to be configured for this Logical Unit.
+Setting this integer to 1 will enable 3PC IOCTL emulation.
+</longdesc>
+<shortdesc lang="en">SCSI extended write</shortdesc>
+<content type="integer" />
+</parameter>
+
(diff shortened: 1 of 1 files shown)


sammcj:
Current workaround is a disgusting `anything` resource:

```bash
#!/bin/bash
#
# Dodgy script to enable additional iSCSI settings
# Can be removed when https://github.com/ClusterLabs/resource-agents/issues/610 is resolved.

trap "exit 0" SIGTERM

# Enable discard etc...
nice -10 find /sys/kernel/config/target/core/ -regextype awk -regex '(.*)emulate_(tpu|3pc|caw)' -exec sh -c 'echo "1" > {}' \;

# safety net incase anyone runs targetcli by hand
nice -10 targetcli set global auto_save_on_exit=false

while :; do
    nice -10 sleep 1
done
```

🤢

dmuhamedagic:
The three OCF_RESKEY_..._default variables are superfluous because there is no default, at least not as the feature is implemented. Either remove them or change the implementation to use these defaults in case the parameters are not set. My guess is that the former would be preferred.

sammcj:
@dmuhamedagic, does that look better to you?

dmuhamedagic:
There are still references to the now non-existing variables OCF_RESKEY_emulate_..._default. Can you also squash the commits into a single one. With "git push -f" you can push that over the existing commits in your branch. Finally, the commit message should be something like:

Medium: iSCSILogicalUnit: add emulate parameters

Thanks!

sammcj:
omg @dmuhamedagic you're going to laugh at me... guess who forgot to push their re-based update.... 😆

sammcj:
Also, after using GitLab for so long, I find Github really dorky to use :/

sammcj:
Hi @dmuhamedagic , are you happy with this MR now?

dmuhamedagic:
I somehow missed the update. Yes, looks good now. Thanks for the contribution!

sammcj:
Thanks for merging this @dmuhamedagic , just rolled out to a heap of storage servers and it's really cleaned things up.

dmuhamedagic:
Thanks are to you for contributing!

PR #988, opened 2017-06-07
iSCSITarget: targetcli rely on defaults

oalbrigt:
Delete targetcli settings to rely on default settings (e.g. auto_add_default_portal gets saved even with save_on_exit=false and clearconfig doesnt clear it either).


oalbrigt:
My bad. Closing this.

PR #990, opened 2017-06-08
sapdb.sh: add nameserver to the HANA default monitor services

fmherschel:
We found that only monitoring '**hdbindexserver**' on many SAP HANA installations you get during the resource start, because SAPDatabase does not find any service to be monitored. A workaround is to set  **MONITOR_SERVICES="hdbindexserver|hdbnameserver"**. The proposed change will take the monitoring of the indexserver OR nameserver as **default**.

### DIFF
--- a/heartbeat/sapdb.sh
+++ b/heartbeat/sapdb.sh
@@ -360,7 +360,7 @@ then
          ;;
     SYB) export OCF_RESKEY_MONITOR_SERVICES="Server"
          ;;
-    HDB) export OCF_RESKEY_MONITOR_SERVICES="hdbindexserver"
+    HDB) export OCF_RESKEY_MONITOR_SERVICES="hdbindexserver|hdbnameserver"
          ;;
   esac
 fi


oalbrigt:
Looks good to me, but can you update the commit message and PR title with a short description of the change?

oalbrigt:
@fmherschel: ping. Can you update commit and PR message as I've suggested?

fmherschel:
I am not sure, if I really managed the requested changes. At least the pull request title seams to be good now. I also tried to change the commit message, but I am not sure if that one worked.

oalbrigt:
Thanks. The commit message didnt go through.

You should be able to do it by doing "git commit --amend", update the message and then "git push -f".

fmherschel:
As the change of the commit message does not work as expected I close (widthdrow) this request and create a new one with a corrected commit message. Github help says that they do not recommend to use a force push after a commit amend, so I do not want to damage the upstream repository.

PR #993, opened 2017-06-18
Properly Detect read_only State

richotero:
Many MySQL system variables include "read_only" in their names. Use a
stricter query in get_read_only() to ensure that the value of read_only
is obtained, not innodb_read_only, tx_read_only, or some other
similarly-named variable.


krig:
LGTM

PR #995, opened 2017-06-26
Medium: ocf-shellfuncs: improve locking (ocf_take_lock)

dmuhamedagic:
This change improves locking by ocf_take_lock(). It uses mkdir(1)
to prevent two instances from creating the same directory (named
by the lock).

The major difficulty is to prevent a race when a stale lock is
discovered. If two processes try to remove the stale lock at
about the same time, the one which runs slightly later can remove
the lock which just got created by the one which run slightly
earlier. The probability of this race is significantly reduced by
testing for stale lock twice with a random sleep in between.

Though this change does not exclude a race entirely, it makes it
extremely improbable. In addition, stale locks are result of only
abnormal circumstances and occur seldom.

The function providing random numbers has been modified to use
either /dev/urandom or awk (with the process pid as the seed).

It was thoroughly tested with both stale lock simulation and
without, by running 64 instances of processes trying to get the
lock on a workstation with 4 cpus.

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -72,10 +72,11 @@ ocf_is_root() {
 }
 
 ocf_maybe_random() {
-	local rnd="$RANDOM"
-	# Something sane-ish in case a shell doesn't support $RANDOM
-	[ -n "$rnd" ] || rnd=$$
-	echo $rnd
+	if test -c /dev/urandom; then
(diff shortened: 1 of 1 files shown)


krig:
Since there have been no further comments, I will merge this. It looks good to me, but I don't know if it's possible to fully grasp the possible races in this... at least the earlier code was known to have issues.

Thank you!

PR #1000, opened 2017-07-12
Medium: Raid1: Handle case when mddev is a symlink

zhilongliu:
from mdadm rule, it would create a symlink to /dev/md(default as 127, likely used "int name--" until find a valid number) if user specify cluster-md name like /dev/md/xxxx, thus the function raid1_monitor_one() should parse whether or not the $mddev is a symlink.

### DIFF
--- a/heartbeat/Raid1
+++ b/heartbeat/Raid1
@@ -371,7 +371,6 @@ raid1_monitor_one() {
 			return $OCF_NOT_RUNNING
 		fi
 	fi
-
 	if ! grep -e "^$md[ \t:]" /proc/mdstat >/dev/null ; then
 		ocf_log info "$md not found in /proc/mdstat"
 		return $OCF_NOT_RUNNING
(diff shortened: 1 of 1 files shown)


krig:
Could you change the commit message to:

```
Medium: Raid1: Handle case when mddev is a symlink
```

zhilongliu:
thanks, this title is better. I would update it.

Regards,
-Zhilong

zhilongliu:
we'd better ignore and close this request, I would re-send the pull-request in a new branch on tomorrow morning. I cann't "amend" the commit message via to the web interface and then the mistake happens.
Sorry for that.

krig:
@zhilongliu You can amend the commit message from the command line and then force push the new commit to the branch: That will update the pull request. No need to create a new pull request!

zhilongliu:
@krig I have checkout my branch, and have "git commit --amend" the commit message, then git push -f.
Thanks very much.

krig:
Hm well, now it seems we are hitting a github bug: https://github.com/isaacs/github/issues/361

I guess you will have to create a new pull request anyway. If we had reopened the request and THEN force pushed, it would have worked. :)

zhilongliu:
Yes, I cannot reopen this request any more, so have to submit a new request.

PR #1001, opened 2017-07-13
Medium: Raid1: Handle case when mddev is a symlink

zhilongliu:
For the mddev name, the function raid1_monitor_one()
should parse whether or not the $mddev is a symlink,
because the symlink name wouldn't show in /proc/mdstat
and result in "raid1_monitor_one()" returns failure.


zhilongliu:
@krig Here I create a new request, thanks very much.

krig:
For clarity, this is an updated version of #1000.

PR #1002, opened 2017-07-13
Low: sg_persist: Read empty value when when no attribute exists

Werkov:
bsc#1048288

Tested on local cluster, happens only when in probe monitor operations.


krig:
Looks good, thanks!

PR #1004, opened 2017-07-19
VirtualDomain: new attributes migrateuri, remoteuri, migration_user 

dhoffend:
The `remote_user` attribute allows to specify a username that might be useful if you want to use a different username then root when using `qemu+ssh://` transport and restrict virsh access to a non-root user. (if root login is not permitted)

The `remote_network_suffix` attribute is similar to the `migration_network_suffix` attribute and modifies the remoteuri if the virsh signaling should be moved to a different network.

Example: Move virsh control and migration traffic (if combined with `migration_network_suffix`) to a more reliable network (f.e. crosslink bonding) and allow live migration while the main network is down (this is where the virsh connection would fail before the migration could even begin).

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -108,6 +108,17 @@ use libvirt's default transport to connect to the remote hypervisor.
 <content type="string" default="" />
 </parameter>
 
+<parameter name="migration_user" unique="0" required="0">
+<longdesc lang="en">
+The username will be used in the remote libvirt remoteuri/migrateuri. No user will be
+given (which means root) in the username if omitted
+
+If remoteuri is set, migration_user will be ignored.
+</longdesc>
+<shortdesc lang="en">Remote username for the remoteuri</shortdesc>
+<content type="string" />
+</parameter>
+
 <parameter name="migration_downtime" unique="0" required="0">
 <longdesc lang="en">
 Define max downtime during live migration in milliseconds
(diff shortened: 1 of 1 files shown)


heartbeat/VirtualDomain
```diff
@@ -195,6 +195,30 @@ This port will be used in the qemu migrateuri. If unset, the port will be a rand
...
+The username will be used in the remote libvirt remoteuri/migrateuri. No user will be
+given (which means root) in the username if omitted
+</longdesc>
+<shortdesc lang="en">Remote for remoteuri/migrateuri</shortdesc>
+<content type="string" />
+</parameter>
+
+<parameter name="remote_network_suffix" unique="0" required="0">
+<longdesc lang="en">
+Use a dedicated network to communitce with the remote hypervisor. The remote
```
krig: Small typo: `s/communitce/communicate`

dhoffend:
Sure. I'll take a closer look. I'm also looking to add the attribute remote_uri so we can make use of virsh Uri parameters like keyfile.

I'm just unsure because migrate Uri is integrated into migrate_options. Adding remote_uri would be different to migrateuri but in the end would make sense

Later :-)

Today the remote_user stuff works flawless in our 2node setup and we can disable root login again

dhoffend:
I'm unsure about the parameter naming. Some are really confusing and missleading. First we have 2 different URI.

* remoteuri: the connection to the remote hypervisor to initate transfer/migration
* migrationuri: usally used with highport to sync the state/ram

Now we're existing parameters:
* migration_transport: (example ssh) this is used as part the remoteuri (not migrationuri)
* migration_network_suffix: used only for the migrationuri
* migration_options: if `--migrateuri` is included that parameter will be used as migrateuri for state transfer

instead of adding lots of new parameters maybe I should reuse them
* rename remote_user to migration_user
* remove remote_network_suffix and reuse migration_network_suffix for both remoteuri and migrateuri.
* add remote_uri for full customization
* add migrate_uri which overrules `--migrateuri` if set. It makes things cleaner IMO

Let me take a look. Thoughts?

oalbrigt:
@dhoffend: It makes sense to reuse and improve the existing parameters and use the same kind of naming for the new ones, and improve metadata for them if they're unclear.

dhoffend:
Okay I've refactored my changes. We now have seperate `migrateuri` and `remoteuri` attributes to have the full flexibility of libvirt connections uris.

For simple changes the attribute `migration_user` adds the username to the auto generated remoteuri.

The parameter `migration_network_suffix` also affects the remoteuri (control connection) not only the migrationuri (state transfer). This might affect existing usages but I feel that the amount of users is fairly low who choose to separate control and state traffic and those will be more happy to have clean and dedicated parameters.

The `--migrationuri` parameter within `migration_options` still exists for backward compatibility but IMO it should be deprecated. If migrationuri or remoteuri is configured parameters like transport/user/suffix are ignored.

dhoffend:
@krig would you be so kind and take another look/review into the merge requests. I would like to know if there's anything I should change.

krig:
Looks OK to me now, @oalbrigt ?

oalbrigt:
Looks good to me too.

PR #1006, opened 2017-07-30
Move "set read-only" operation from pre-demote to demote

ikedaj:
Pacemaker will call "pre-demote" when its transition start.
At this time, some MySQL clients don't stop nomary.
They might try to update their status during stop operation, but MySQL is already "read-only" mode.
It's better to do "set read-only" at, not "pre-demote" but "demote".

### DIFF
--- a/heartbeat/mysql
+++ b/heartbeat/mysql
@@ -897,6 +897,30 @@ mysql_demote() {
         return $OCF_NOT_RUNNING
     fi
 
+    # Demote Master
+    demote_host=`echo $OCF_RESKEY_CRM_meta_notify_demote_uname|tr -d " "`
+    if [ $demote_host = ${NODENAME} ]; then
+        ocf_log info "Set read-only mode on"
+        set_read_only on
+        if [ $? -ne 0 ]; then
+            ocf_exit_reason "Failed to set read-only";
+            return $OCF_ERR_GENERIC;
+        fi
+
+        # Must kill all existing user threads because they are still Read/write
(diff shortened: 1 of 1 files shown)


heartbeat/mysql
```diff
@@ -949,30 +973,7 @@ mysql_notify() {
...
-
-                for thread in `awk '$0 !~ /Binlog Dump|system user|event_scheduler|SHOW PROCESSLIST/ && $0 ~ /^[0-9]/ {print $1}' $tmpfile`
-                do
-                    ocf_run $MYSQL $MYSQL_OPTIONS_REPL \
-                        -e "KILL ${thread}"
-                done
-            else
-               ocf_log info "Ignoring post-demote notification execpt for my own demotion."
-            fi
+            ocf_log info "Ignoring pre-demote notification for my own demotion."
```
dmuhamedagic: This log message doesn't seem to be correct, i.e. there's a case of "my demotion" but also one of "demotion of the other".

dmuhamedagic:
On Sat, Jul 29, 2017 at 10:28:08PM -0700, Junko IKEDA wrote:
> Pacemaker will call "pre-demote" when its transition start.
> At this time, some MySQL clients don't stop nomary.

"nomary"? Cannot find it in my dictionary ;-)

> They might try to update their status during stop operation, but MySQL is already "read-only" mode.
> It's better to do "set read-only" at, not "pre-demote" but "demote".

So, set read only at the demote time? Sounds good to me. Perhaps
just reformulate slightly the last sentence.

PR #1007, opened 2017-07-31
Add hdbnameserver to monitor services

fmherschel:
Adding hdbnameserver to the list of monitor services for HANA databases (dbtype HDB/hdb). The goal is to change the default to address older and newer HANA versions. In newer HANA versions the hdbindexserver is typically not listed as an own service.


oalbrigt:
Thanks.

PR #1013, opened 2017-08-07
lvmlockd: add resource agent for lvmlockd daemon

renzhengeek:
"lvmlockd" is a pretty new feature from LVM2 project.
It plays the same role as "clvmd" whose resource agent
is "clvm", but with much more proper design. The primary
benefit of "lvmlockd" is that it supports two lock
managers: dlm and sanlock. Refer to lvmlockd.8 for more
information.

This new agent only aim to support "dlm + lvmlockd",
because "sanlock" lock mananger plugin is suggested
to be used in scenario where HA clster softwares, i.e.
corosync/pacemaker/dlm are not already being needed/used.

Signed-off-by: Eric Ren <zren@suse.com>

### DIFF
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -65,6 +65,7 @@ ocf_SCRIPTS	     =  AoEtarget		\
 			IPsrcaddr		\
 			LVM			\
 			LinuxSCSI		\
+			lvmlockd		\
 			MailTo			\
 			ManageRAID		\
 			ManageVE		\
--- a/heartbeat/lvmlockd
+++ b/heartbeat/lvmlockd
@@ -1,0 +1,438 @@
+#!/bin/bash
+#
+#
+#	lvmlockd OCF Resource Agent
+#
+# Copyright (c) 2017 SUSE LINUX, Eric Ren
+#			All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.	Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+
+#######################################################################
+# Initialization:
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+#######################################################################
+
+meta_data() {
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="lvmlockd">
+<version>1.0</version>
+
+<longdesc lang="en">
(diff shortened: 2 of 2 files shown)


renzhengeek:
@krig I've tested this RA for several rounds. There was one big issue inside lvmlockd daemon found during testing. I reported it upstream, and it has been fixed in the lvm2 side by David Teigland.

Could you help review this? The work almost completes. I can fixup small issues afterward if I find something. Thanks.

heartbeat/Makefile.am
```diff
@@ -144,7 +144,8 @@ ocf_SCRIPTS	     =  AoEtarget		\
 			varnish			\
 			vmware			\
 			vsftpd			\
-			zabbixserver
+			zabbixserver		\
+			lvmlockd
```
krig: Please add lvmlockd at the right place in the list of agents to keep the list sorted alphabetically, thank you :)

heartbeat/lvmlockd
```diff
@@ -0,0 +1,438 @@
...
+meta-data)		meta_data
+			exit $OCF_SUCCESS
+			;;
+usage|help)		usage
+			exit $OCF_SUCCESS
+			;;
+esac
+
+# Anything other than meta-data and usage must pass validation
+lvmlockd_validate || exit $?
```
krig: I'm not sure about this. No other agents run validation for each action, only in the validate-all call. Are you worried that pacemaker does not act correctly?

heartbeat/lvmlockd
```diff
@@ -0,0 +1,438 @@
...
+		daemon_is_running "$pid"
+	else
+		# No pid file
+		false
+	fi
+}
+
+shared_vgs()
+{
+	local vg_list=()
```
krig: Ah, this is the problem: Your shebang is `/bin/sh`, but this is not valid POSIX shell syntax. Please either rewrite this as POSIX code, or change the shell used in the shebang to `/bin/bash`.

krig:
So to clarify what the main problem is: Travis CI runs on Ubuntu, which uses `dash` as `/bin/sh`. `dash` is a lot more strict about POSIX syntax than `bash`, so it fails to run the `lvmlockd` script with this error:

```
./lvmlockd: 156: ./lvmlockd: Syntax error: "(" unexpected (expecting "}")
```

The fix is either to rewrite the script to run with `/bin/bash` as the parser, or to change the code into something POSIX compatible.

feist:
@renzhengeek 
I'm wondering if it makes sense for this to be split into 2 resource agents (or 1 new RA and an expansion of the existing lvm RA).

I think we want 1 RA that just starts and monitors lvmlockd and then a separate RA that does the vgchange (optionally using -asy to use the LV in shared mode for GFS/OCFS file systems).

This would let us run lvmlockd on all nodes of the cluster, but mount different file systems on different LVs on a subset of all the nodes in the cluster.  It would also help us deal with certain LVs that should be marked as shared (for GFS/OCFS).  I *think* it would be possible to expand the existing lvm RA with some extra options to do this.

heartbeat/Makefile.am
renzhengeek: OK, thanks!

heartbeat/lvmlockd
renzhengeek: Aha, I learned this pattern from resource agent guide:

[4. Resource agent structure](http://www.linux-ha.org/doc/dev-guides/_execution_block.html)
```

# Anything other than meta-data and usage must pass validation
foobar_validate_all || exit $?
```

heartbeat/lvmlockd
renzhengeek: OK. I tend to use `/bin/bash`, noticed many RAs have used `bash`, hah

renzhengeek:
@feist 
Thanks for taking a look at this.  I keep this idea in mind that `lvmlockd` should behave similarly like `clvmd` - they both try to activate `clustered/shared` VGs on only `start` action, optional with parameter `activate_vgs`.

Yes, like `clvmd`, `lvmlockd` also need to use together with `lvm` RA to monitor each VG individually.

And yes, in order to make `lvm` RA work for `shared` VG, I also plan to make respective changes in `lvm` RA once this PR accepted, which should be easy.

Does it sound good to you? Thanks again.

heartbeat/lvmlockd
krig: Yes, that's OK :)

feist:
@renzhengeek 
Yes, I think that sounds good to me.

PR #1016, opened 2017-08-09
Add ZFS Resource Agent

utopiabound:
Adds a resource agent to import/export ZFS pools.

This resource agent has a long history of use and is well tested.

### DIFF
--- a/heartbeat/ZFS
+++ b/heartbeat/ZFS
@@ -1,0 +1,211 @@
+#!/bin/sh
+#
+# License:      GNU General Public License (GPL)
+# Support:      zfs@lists.illumos.org
+# Written by:   Saso Kiselkov
+#
+#	This script manages ZFS pools
+#	It can import a ZFS pool or export it
+#
+#	usage: $0 {start|stop|status|monitor|validate-all|meta-data}
+#
+#	The "start" arg imports a ZFS pool.
+#	The "stop" arg exports it.
+#
+#       OCF parameters are as follows
+#       OCF_RESKEY_pool - the pool to import/export
+#
+#######################################################################
+# Initialization:
+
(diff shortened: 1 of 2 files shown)


krig:
Looks generally clean and good to me, but I'm curious why it has a separate debug log mechanism? There is the `ocf_log debug` command built into the OCF library, and the pacemaker `trace_ra` mechanism to get even more detail.

heartbeat/ZFS
```diff
@@ -0,0 +1,198 @@
...
+</longdesc>
+<shortdesc lang="en">Import arguments</shortdesc>
+<content type="string" default="" />
+</parameter>
+<parameter name="importforce" unique="1" required="0">
+<longdesc lang="en">
+zpool import is given the -f option.
+</longdesc>
+<shortdesc lang="en">Import is forced</shortdesc>
+<content type="boolean" default="true" />
```
oalbrigt: You should define and assign "_default" parameters to use here and assign if not set by user (grep for _default in the other agents for examples).

heartbeat/ZFS
```diff
@@ -0,0 +1,198 @@
...
+        debug_log "starting import"
+
+        # The meanings of the options to import are as follows:
+        #   -f : import even if the pool is marked as imported to another
+        #        system - the system may have failed and not exported it
+        #        cleanly.
+        #   -o cachefile=none : the import should be temporary, so do not
+        #        cache it persistently (across machine reboots). We want
+        #        the CRM to explicitly control imports of this pool.
+	if [ ${OCF_RESKEY_importforce} = 1 ]; then
```
oalbrigt: ocf_is_true would be better here, as that matches true/on/1 and so on.

oalbrigt:
Looks good to me.

PR #1017, opened 2017-08-11
awseip/awsvip: fixes and improvements

oalbrigt:
--public-ip doesnt work (tested with latest awscli from pip as well), so it had to be changed to --allocation-id and --association-id.


krig:
Looks good to me.

PR #1018, opened 2017-08-16
VirtualDomain - virsh shutgdown shutdown_mode support 

vaLski:
This commit allows operators to define virsh shutdown --mode in case traditional 'virsh shutdown' does not work.

virsh 3.5.0 supports one of the following --mode parameters - acpi|agent|initctl|signal|paravirt

By default shutdown_mode is empty and --mode is passed on the command line only in case is set

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -229,6 +229,19 @@ Restore state on start/stop
 <content type="string" default=""/>
 </parameter>
 
+<parameter name="shutdown_mode">
+<longdesc lang="en">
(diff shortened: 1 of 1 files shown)


krig:
This looks OK to me.

berlevdv:
Works with virsh 3.2.0 version.

Is it worth checking the status of the "virsh shutdown" command execution? For example, if a guest agent is not installed in the guest system, and after an unsuccessful run "virsh shutdown" it can be executed without "--mode" parameter?

PR #1019, opened 2017-08-18
redis: when containerized, use the host name rather than the bundle name

dciabrin:
Redis records the last known master in the CIB. When run in a bundle,
the host name used is that of the bundle, which is incorrect. One must
use the name of the host on which the bundle ran, because that's where
the Redis server's state is stored on disk.

### DIFF
--- a/heartbeat/redis
+++ b/heartbeat/redis
@@ -545,7 +545,7 @@ function validate() {
 	fi
 }
 
-NODENAME=$(ocf_local_nodename)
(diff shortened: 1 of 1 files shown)


dciabrin:
FTR: I just opened this PR while @mbaldessari and myself are testing that it does what we want. Please hold off a bit until we validated that :)

dciabrin:
Obsoleted, closing that PR now

PR #1020, opened 2017-08-18
galera: when containerized, use the host name rather than the bundle name

dciabrin:
Galera resource agents records various attributes in the CIB that are
linked to the state of the local mysql database (e.g. last-committed).
When run in a bundle, the host name that the resource agent receives is
that of the bundle, but it cannot use it to stored attributes because
bundle replicas can move across physical hosts when restarted.

Use name of the node which is hosting the container to store to stores
attributes that reflect the state of the local mysql server. Keep
using the bundle's name for master/slave state attribute.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -68,6 +68,17 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 . ${OCF_FUNCTIONS_DIR}/mysql-common.sh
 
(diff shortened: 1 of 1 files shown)


dciabrin:
FTR: I just opened this PR while @mbaldessari and myself are testing that it does what we want. Please hold off a bit until we validated that :)

dciabrin:
Obsoleted, closing that PR now

Issue #915, opened 2017-01-18
Galera agent doesn't work when grastate.dat contains safe_to_bootstrap


gyurco:
The galera agent doesn't work with the safe-to-bootstrap feature:
http://galeracluster.com/2016/11/introducing-the-safe-to-bootstrap-feature-in-galera-cluster/

A simple fix can be a
````
sed -i 's/^safe_to_bootstrap:.*/safe_to_bootstrap: 1/' ${OCF_RESKEY_datadir}/grastate.dat
````
in galera_start_local_node(), after 'if ocf_is_true $bootstrap; then'

I think it is fairly safe, because it enables bootsrapping for the already detected best node.

dciabrin:
So I'm not familiar with that new flag yet, but your proposal seems to essentially bypass it entirely rather than use it for detecting the bootstrap node.

Have you checked where this flag is stored and used in galera?

My concern with that proposal is that by doing so, we deny the existence of this new "feature", and if that value is also stored elsewhere (e.g. innodb), circumventing it will have bad results. To illustrate my point, "seqno" value is store both in grastate.dat and in innodb (fetchable with mysqld_safe --wsrep-recover).

gyurco:
I'm not aware of any other locations where it is stored. 

I think the existing detection via the --wsrap-recover is good enough. As the above docs says, it is a convenient flag for sysadmins to start the cluster, and the existing methods determining the "best" node still works. And if all nodes marked with "safe_to_bootstrap: 0" by galera, then the proposed change is a must (I've tested it with Percona XtraDB Cluster 5.7).
One can argue, if a node is marked with "safe_to_bootstrap: 1" already, and the found best node is not the same, then what to do.

dciabrin:
> And if all nodes marked with "safe_to_bootstrap: 0" by galera, then the proposed change is a must (I've tested it with Percona XtraDB Cluster 5.7).
Do you recall the condition which triggered that? What is Codership's recommend way to restart the cluster in such condition?

> One can argue, if a node is marked with "safe_to_bootstrap: 1" already, and the found best node is not the same, then what to do.
Suppose there's a network partition impacting all the node, they will go non-primary and the resource agent will shut them down. Will they all have "safe_to_bootstrap: 1" afterwards?

gyurco:
Actually stopping the cluster with the crm_resource command leaves all nodes with "safe_to_bootstrap: 0" for me. In this case, the docs says:
````
[...]

PR #1022, opened 2017-08-22
galera: Honor "safe_to_bootstrap" flag in grastate.dat

rhafer:
With version 3.19 galera introduced the "safe_to_bootstrap" flag to
the grastate.dat file [1]. When a cluster is shutdown cleanup, the last
not to shutdown gets this flag set to 1. (All other get a 0).

This commit enhances the galera resource agent to make use of that flag
when selecting an appropriate node for bootstrapping the cluster.  When
any of the cluster nodes has the "safe_to_bootstrap" flag set to 1, that
node is immediately selected as the boostrap node of the cluster.

When all nodes have safe_to_bootstrap=0 or the flag is not present the
current bootstrap behaviour mostly unchanged. We just set
"safe_to_bootstrap" to 1 in grastate.dat on the selected bootstrap node
to a allow for galera to start, as outlined in the galera documentation
[2].

Fixes: #915

[1] http://galeracluster.com/2016/11/introducing-the-safe-to-bootstrap-feature-in-galera-cluster
[2] http://galeracluster.com/documentation-webpages/restartingcluster.html#safe-to-bootstrap-protection

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -332,6 +332,27 @@ get_last_commit()
     fi
 }
 
+clear_safe_to_bootstrap()
+{
+    ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -D
+}
+
+set_safe_to_bootstrap()
+{
+    ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -v $1
+}
+
+get_safe_to_bootstrap()
+{
+    local node=$1
+
+    if [ -z "$node" ]; then
+        ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -Q 2>/dev/null
+    else
+        ${HA_SBIN_DIR}/crm_attribute -N $node -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -Q 2>/dev/null
+    fi
+}
+
 wait_for_sync()
 {
     local state=$(get_status_variable "wsrep_local_state")
@@ -465,6 +486,7 @@ detect_first_master()
     local all_nodes
     local best_node_gcomm
     local best_node
+    local safe_to_bootstrap
 
     all_nodes=$(echo "$OCF_RESKEY_wsrep_cluster_address" | sed 's/gcomm:\/\///g' | tr -d ' ' | tr -s ',' ' ')
     best_node_gcomm=$(echo "$all_nodes" | sed 's/^.* \(.*\)$/\1/')
@@ -492,6 +514,19 @@ detect_first_master()
     done
 
     for node in $nodes_recovered $nodes; do
+        safe_to_bootstrap=$(get_safe_to_bootstrap $node)
+
+        if [ "$safe_to_bootstrap" = "1" ]; then
+            # Galera marked the node as safe to boostrap during shutdown. Let's just
+            # pick it as our bootstrap node.
+            ocf_log info "Node <${node}> is marked as safe to bootstrap."
+            best_node=$node
+
+            # We don't need to wait for the other nodes to report state in this case
+            missing_nodes=0
+            break
+        fi
+
         last_commit=$(get_last_commit $node)
 
         if [ -z "$last_commit" ]; then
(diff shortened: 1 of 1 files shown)


krig:
At a glance, this change looks incompatible with versions < 3.19... am I wrong about that or is that an issue?

jsuchome:
Commit message: I do not really understand 2nd sentence (_When cluster ..._ );don't you mean 'node' instead of 'cluster'?

krig:
Nevermind my previous comment, I think I get what it does now.

heartbeat/galera
```diff
@@ -332,6 +332,27 @@ get_last_commit()
...
+{
+    ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -v $1
+}
+
+get_safe_to_bootstrap()
+{
+    local node=$1
+
+    if [ -z "$node" ]; then
+        ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -Q 2>/dev/null
```
jsuchome: What is `-Q`? Don't you want `-G`?

heartbeat/galera
```diff
@@ -522,6 +557,22 @@ detect_first_master()
...
+detect_safe_to_bootstrap()
+{
+    local safe_to_bootstrap=""
+
+    if [ -f ${OCF_RESKEY_datadir}/grastate.dat ]; then
+        safe_to_bootstrap=$(sed -n 's/^safe_to_bootstrap:\s*\(.*\)$/\1/p' < ${OCF_RESKEY_datadir}/grastate.dat)
+    fi
+
+    if [ "$safe_to_bootstrap" = "1" ] || [ "$safe_to_bootstrap" = "0" ]; then
+        ocf_log info "attempting to read safe_to_bootstrap flag from ${OCF_RESKEY_datadir}/grastate.dat"
```
jsuchome: This is bit misleading imho - it was already read and now you're adapting crm attribute to that saved value...

heartbeat/galera
rhafer: It is a (legacy) alias for `-q/--quiet`. I used it because all the other crm_attribute calls here use it.

jsuchome: Ah, than it makes sense (kind of)

jsuchome:
Tested it manually, seems to work as it should. Thanks!

heartbeat/galera
rhafer: Hm, yeah. That Comment endup up on the wrong line after I did move some code around. I adjusted it.

rhafer:
> Commit message: I do not really understand 2nd sentence (When cluster ... );don't you mean 'node' instead of 'cluster'?

I rephrased it. I hope is clearer now.

heartbeat/galera
jsuchome: Thanks

heartbeat/galera
```diff
@@ -332,6 +332,27 @@ get_last_commit()
...
+}
+
+get_safe_to_bootstrap()
+{
+    local node=$1
+
+    if [ -z "$node" ]; then
+        ${HA_SBIN_DIR}/crm_attribute -N $NODENAME -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -Q 2>/dev/null
+    else
+        ${HA_SBIN_DIR}/crm_attribute -N $node -l reboot --name "${INSTANCE_ATTR_NAME}-safe-to-bootstrap" -Q 2>/dev/null
```
jhesketh: (nit) Could reduce duplication by setting $node to $NODENAME when $node is empty

heartbeat/galera
```diff
@@ -492,6 +514,19 @@ detect_first_master()
     done
 
     for node in $nodes_recovered $nodes; do
+        safe_to_bootstrap=$(get_safe_to_bootstrap $node)
+
+        if [ "$safe_to_bootstrap" = "1" ]; then
+            # Galera marked the node as safe to boostrap during shutdown. Let's just
+            # pick it as our bootstrap node.
```
jhesketh: Does Galera marking nodes safe to bootstrap mean they are at the latest version or just that they shutdown cleanly. If it's the latter, then this isn't sufficient. I suspect it's the former but just checking/curious.

heartbeat/galera
```diff
@@ -605,6 +656,11 @@ galera_promote()
         bootstrap=$(is_bootstrap)
 
         if ocf_is_true $bootstrap; then
+            # The best node for bootstrapping wasn't cleanly shutdown. Allow
+            # bootstrapping anyways
```
jhesketh: What are the consequences of this? If it wasn't clean, why are we using it?
I presume if we get to this point it means no nodes are clean. However does that mean we need support intervention?

heartbeat/galera
```diff
@@ -605,6 +656,11 @@ galera_promote()
...
+            if [ "$(get_safe_to_bootstrap)" = "0" ]; then
+                sed -ie 's/^\(safe_to_bootstrap:\) 0/\1 1/' ${OCF_RESKEY_datadir}/grastate.dat
```
jhesketh: Should we be modifying this file outside of galera's control?

krig: Hm, yeah. This does look problematic. Wouldn't an admin need to interfere if the data got corrupted somehow? This seems like it's circumventing the feature.

rhafer: Actually the resource agent does exactly what the galera documentation outline. If no node has the safe_to_bootstrap flag, check the latest sequence id on all nodes and use the node with the latest for bootstrapping. I think that perfectly fine to automate (and pretty much what the agent already did in the past)

If the agent would not do that a simple `crm resource restart galera` would not work. If mysql is restarted at pretty much the same time on all node the flag is not updated.

heartbeat/galera
rhafer: "safe_to_bootstrap" means it's was the last one leaving the cluster. So it was cleanly shutdown and has the latest data.

heartbeat/galera
rhafer: Hm, true. I just copy'n pasted it form a few lines above.

heartbeat/galera
rhafer: > What are the consequences of this? If it wasn't clean, why are we using it? I presume if we get to this point it means no nodes are clean.

Right

> However does that mean we need support intervention?

Not at this point. We were still able to figure which node has the latest data and we can recover from there. (It's just fallback to the old behaviour of the resource-agent, which is still valid)

jhesketh:
Thanks for addressing my comments :-). Sounds good to me!

PR #1023, opened 2017-08-23
Route: Configurable address family

### LINKED ISSUE #996: Route agent failed when adding default IPv6 route
Actually, the Route agent uses the 'OCF_RESKEY_destination' key to detect the `addr_family`

Unfortunately, when adding an IPv6 gateway, the destination is `default` and the route script incorrectly consider this is an IPv4 address. As the `addr_family` is used to check the route status, it then report a failure.

CAPSLOCK2000:
Make the address family for routes configurable. This is required because it is not always possible to auto-detect the correct family, eg 'ip route add default eth0' is valid for both IPv4 and IPv6.
Falls back to the old (fallible) behaviour when not specified.

Fixes #996

### DIFF
--- a/heartbeat/Route
+++ b/heartbeat/Route
@@ -318,8 +318,6 @@ case $OCF_RESKEY_family in
         esac ;;
     *) ocf_exit_reason "Address family '${OCF_RESKEY_family}' not recognized." ;;
 esac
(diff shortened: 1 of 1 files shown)


krig:
Looks good to me, thank you!

PR #1024, opened 2017-08-24
galera: Fix instance name in master_exists()

rhafer:
Under certain circumstances $OCF_RESOURCE_INSTANCE as passed to the
resource agent contains and instance number (e.g. "galera:0"). This
happens e.g. when using "crm_resource --force-stop/--force-start". In
those cases "master_exists()" doesn't work correctly.

Use $INSTANCE_ATTR_NAME which is set in "mysql-common.sh", based on
$OCF_RESOURCE_INSTANCE with the instance number stripped.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -386,7 +386,7 @@ master_exists()
         return 1
     fi
     # determine if a master instance is already up and is healthy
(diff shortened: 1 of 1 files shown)


krig:
Looks good to me

jhesketh:
Should this also be changed on line 611?

rhafer:
> Should this also be changed on line 611?

I think it's ok to have the instance number visible as part of the error message.

PR #1029, opened 2017-08-28
High: exportfs: Make fsid optional (bsc#1045161)

krig:
The fsid export option is not needed any more.  There
might be some strange corner cases where it can be
useful, but it definitely shouldn't be assumed to be
needed by default.

The `fsid=0` usage used to be important for NFSv4, but
this is no longer the case.  It is best *not* to use
`fsid=0` and to let `exportfs/mountd` create a
"NFSv4 pseudo root" themselves.  This is (almost)
always what is wanted and is least confusing.

The `fsid=N N!=0` usage was important some years ago
when the device number of hard drives started to change.
NFS used to use the device id, which was problematic.
Setting an explicit fsid= removed the instability.
However for quite some years now NFS has been using the
UUID of the filesystem, rather than the device id, to
identify a filesystem.  So unless you have multiple
filesystems with the same uuid, or are using some weird
type of filesystem which doesn't have a uuid, `fsid=`
isn't needed and is best avoided.

So exportfs_test should be changed to not require
`fsid=` any more.


krig:
@oalbrigt Would be grateful for some extra eyeballs on this one, in particular the `cleanup_export_cache` function seems to need the fsid to do cleanup. Maybe it's an acceptable tradeoff for that not to work if fsid is not explicitly configured..

oalbrigt:
Looks good to me, but I think we should update cleanup_export_cache accordingly as well.

PR #1034, opened 2017-09-11
rabbitmq: Use RMQ_CTL var everywhere instead of rabbitmqctl

lemenkov:
We're using $RMQ_CTL instead of a direct rabbitmqctl invocation almost everywhere. Let's do it everywhere - for the sake of consistency.


krig:
Looks good, thanks

PR #1035, opened 2017-09-15
findif: improve IPv6 NIC detection

oalbrigt:
Without this patch findif fails to find NIC for IPv6 addresses with 128 netmask.

### DIFF
--- a/heartbeat/findif.sh
+++ b/heartbeat/findif.sh
@@ -233,6 +233,8 @@ findif()
     fi
     case $1 in
     */*) : OK ;;
+    # "ip route" doesnt show netmask for IPv6 /128
+    *:*:*) : OK ;;
     *)
       ocf_log err "Unable to find cidr_netmask."
       return $OCF_ERR_GENERIC ;;
@@ -248,7 +250,7 @@ findif()
       fi
     fi
   else
-    if [ -z "$OCF_RESKEY_nic" -a "$netmask" != "${1#*/}" ] ; then
+    if [ -z "$OCF_RESKEY_nic" -a -z "$OCF_RESKEY_cidr_netmask" -a "$netmask" != "${1#*/}" ] ; then
       ocf_log err "Unable to find nic, or netmask mismatch."
       return $OCF_ERR_GENERIC
     fi
--- a/heartbeat/findif.sh
+++ b/heartbeat/findif.sh
@@ -250,7 +250,7 @@ findif()
       fi
     fi
   else
-    if [ -z "$OCF_RESKEY_nic" -a -z "$OCF_RESKEY_cidr_netmask" -a "$netmask" != "${1#*/}" ] ; then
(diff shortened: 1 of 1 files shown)


dmuhamedagic:
@kskmori Mori-san, I think that it was you who did most of development here. Could you please take a look?

heartbeat/findif.sh
```diff
@@ -233,24 +233,19 @@ findif()
...
-  if [ $family = "inet" ] ; then
-    if [ -z "$brdcast" ] ; then
-      if [ -n "$7" ] ; then
-        set -- `ip -o -f $family addr show | grep $7`
-        [ "$5" = brd ] && brdcast=$6
-      fi
-    fi
-  else
-    if [ -z "$OCF_RESKEY_nic" -a "$netmask" != "${1#*/}" ] ; then
-      ocf_log err "Unable to find nic, or netmask mismatch."
```
kskmori: Is there any reason to totally eliminate this validation check?
It should detect a wrong configuration for the netmask parameter (which is the case 13 in IPaddr2v6 ocft test cases).

I think that this validation check is still useful and should be retained somehow so that cooperate with prefix /128,
but if you have a good reason to remove it, that might be ok. In that case, we would need to modify the IPaddr2v6 ocft script along with it with some explanations.

heartbeat/findif.sh
```diff
@@ -233,24 +233,19 @@ findif()
...
-      return $OCF_ERR_GENERIC
+  if [ -z "$brdcast" ] ; then
+    if [ -n "$7" ] ; then
+      set -- `ip -o -f $family addr show | grep $7`
+      [ "$5" = brd ] && brdcast=$6
```
kskmori: This part is not necessary to run for inet6 because ```brd``` is only shown when inet.
It would not be harm for the results but it would add a wasteful invocation of the ip command when inet6.

heartbeat/findif.sh
oalbrigt: Do you have any suggestions for how to make it work with /128?

oalbrigt:
@kskmori I've updated the patch, so it should solve the issues you mentioned.

heartbeat/findif.sh
```diff
@@ -248,7 +250,7 @@ findif()
       fi
     fi
   else
-    if [ -z "$OCF_RESKEY_nic" -a "$netmask" != "${1#*/}" ] ; then
+    if [ -z "$OCF_RESKEY_nic" -a -z "$OCF_RESKEY_cidr_netmask" -a "$netmask" != "${1#*/}" ] ; then
```
krig: I'd recommend rewriting this as

```
if [ -z "$OCF_RESKEY_nic" ] && [ -z "$OCF_RESKEY_cidr_netmask" ] && [ "$netmask" != "${1#*/}" ] ; then
```

The `-a` operator doesn't short circuit and is not well-defined with more than 4 arguments, etc.

http://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o

oalbrigt: Nice catch. I'll fix that.

PR #1039, opened 2017-10-10
aws-vpc-move-ip

oalbrigt:
aws-vpc-move-ip from https://github.com/mguertler/AWS-VPC-move-IP with cleanup and improvements.


krig:
Thanks!

PR #1040, opened 2017-10-12
Request for review of LVM-activate RA

renzhengeek:
Hi all,

The LVM-activate RA supports 4 activation mode: lvmlockd, clvmd,
systemID and lvm tag (hosttag?). I have tested them one by one on
my two-node pacemaker cluster with shared disks. But, I didn't stack
filesystem or other service on top of LVM yet. The resource agents
I tried variant combinations of them are:
controld (dlm), clvm, LVM, lvmlockd and LVM-activate. So far, they
works well for me. There were several problems during working on this,
but we (David and me) have solved/worked around them. Thanks David's
support!

Of course, there are still some potential issues to improve. I marked
them with "TODO" comments in the code.

One thing still not clear to me is about the "tagged-based" exclusive
activation. The old LVM RA just uses normal tag feature and activation/
volume_list to achieve the exclusive activation. But, as David mentioned
previously, "hosttag" feature should be used like in this documentation[1].
Currently, the "tag" activation logic in this RA is nearly copied from
LVM. I don't see much need to change it yet, but the naming of it looks
confusing :-/ 

Welcome any feedback and comments!

[1]
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/           Cluster_Logical_Volume_Manager/tag_activation.html

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -1,0 +1,849 @@
+#!/bin/sh
+#
+#
+# Copyright (c) 2017 SUSE LINUX, Eric Ren
+#			All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.	Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+# LVM-activate OCF Resource Agent:
+#
+# Logical volume manager (LVM) provides new features for cluster enviroment:
+# lvmlockd and system ID, which aims to replace clvmd and tagged-exclusive
+# activation. Accordingly, we have created a new resource agent named "lvmlockd"
+# to manage lvmlockd daemon. In addition, this new resource agent "LVM-activate"
+# is created to take care of LVM activation/deactivation work. This agent supports
+# the new features: lvmlockd and system ID, and also supports the old features:
+# clvmd and lvm tag.
+#
+# For LVM activation, we already have an old resource agent "LVM" that supports
+# LVM1, clvmd and lvm tag. Some reasons why we tend to create a new one rather
+# than changing the old:
+# 1. "LVM" supports clvmd and some legacy things. The code looks very messy,
+# which make it difficult to extend for new features, and ensure everything
+# works correctly with huge changes.
+#
+# 2. We want this new RA to do the activation work in a much more safe way by
+# leveraging the new features and activating at VG or LV basis depending on
+# your choice.
+#
+# We will leave the old RA alone while the new one is phased in. In order to replace
+# the old RA smoothly, this RA also tries to support the functionalities of the old
+# RA and inevitably copies some code from the LVM RA.
+#
+# WARNINGS:
+# 1. We support two configure combinations: lvmlockd+LVM-activate and clvm+LVM-activate.
+#    But, you cannot use them in your cluster at the same time!
+# 2. Please put all "lvmlockd"/"clvmd" volume group into auto management by RA,
+#    once you choose pacemaker to manage one of them.  If you manage some
+#    by hand, the stop action of lvmlockd RA may fail and the node may get
+#    fenced consequently, because some DLM lockspaces might be in use and
+#    cannot be closed automatically.
+#
+# Thanks David Teigland! He is the author of these LVM features, giving valuable
+# idea/feedback about this resource agent.
+############################################################################
+
+# Initialization:
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+# If LV is given, only activate this named LV; otherwise, activate all
+# LVs in the named VG.
+VG=${OCF_RESKEY_vgname}
+LV=${OCF_RESKEY_lvname}
+
+# How LVM controls access to the VG:
+#
+# 0: place-holder for any incorrect casees; To be safe, we enforce the VG
+#    must use any of the following protection methods in cluster environment.
+# 1: vg is shared - lvmlockd (new)
+# 2: vg is clustered - clvmd (old)
+# 3: vg has system_id (new)
+# 4: vg has tagging (old)
+VG_access_mode=${OCF_RESKEY_vg_access_mode}
+
+# Activate LV(s) with "shared" lock for cluster fs
+# or "exclusive" lock for local fs
+LV_activation_mode=${OCF_RESKEY_activation_mode:-exclusive}
+
+# For system ID feature
+SYSTEM_ID=""
+
+# For tagging activation mode
+OUR_TAG=${OCF_RESKEY_tag:-pacemaker}
+
+#######################################################################
+
+meta_data() {
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+
+
+<resource-agent name="LVM-activate">
+<version>1.0</version>
+
+<longdesc lang="en">
+This agent manages LVM activation/deactivation work for a given volume group.
+It supports both new features: lvmlockd and system ID, and old features:
+clvmd and lvm tag. Please use the new features as possible as you can!
+</longdesc>
+<shortdesc lang="en">This agent activates/deactivates logical volumes.</shortdesc>
+
+<parameters>
+<parameter name="vgname" unique="1" required="1">
+<longdesc lang="en">
+The volume group name.
+</longdesc>
+<shortdesc lang="en">The volume group name</shortdesc>
+<content type="string" default="" />
+</parameter>
+
+<parameter name="lvname" unique="0" required="0">
+<longdesc lang="en">
+If set, only the specified LV will be activated.
+</longdesc>
+<shortdesc lang="en">Only activate the given LV</shortdesc>
+<content type="string" default="" />
+</parameter>
+
+<parameter name="vg_access_mode" unique="0" required="1">
+<longdesc lang="en">
+This option decides which solution will be used to protect the volume group in
+cluster environment. Optional solutions are: lvmlockd, clvmd, system_id and
+tagging.
+</longdesc>
+<shortdesc lang="en">The VG access mode</shortdesc>
+<content type="string" default="" />
+</parameter>
+
+<parameter name="activation_mode" unique="0" required="0">
+<longdesc lang="en">
+The activation mode decides the visibility of logical volumes in the cluster. There
+are two different modes: "shared" for cluster filesystem and "exclusive" for local
+filesystem. With "shared", an LV can be activated concurrently from multiple nodes.
+With "exclusive", an LV can be activated by one node at a time.
+
+This option only has effect on "lvmlockd"/"clvmd" vg_access_mode. For "system_id"
+and "tagging", they always mean exclusive activation.
+</longdesc>
+<shortdesc lang="en">Logical volume activation mode</shortdesc>
+<content type="string" default="exclusive" />
+</parameter>
+
+<parameter name="tag" unique="0" required="0">
+<longdesc lang="en">
+The tag used for tagging activation mode.
+</longdesc>
+<shortdesc lang="en">The tag used for tagging activation mode</shortdesc>
+<content type="string" default="pacemaker" />
+</parameter>
+
+</parameters>
+
+<actions>
+<action name="start"		timeout="90" />
+<action name="stop"		timeout="90" />
+<action name="monitor"		timeout="90" interval="30" depth="0" />
+<action name="meta-data"	timeout="10" />
+<action name="validate-all"	timeout="20" />
+</actions>
+</resource-agent>
+END
+}
+
+#######################################################################
+
+usage() {
+	cat <<END
+usage: $0 {start|stop|monitor|validate-all|meta-data}
+END
+}
+
+# VG access modes:
+# 0: unsafe to activate LV without proper protection in cluster
+# 1: vg is shared - lvmlockd (new)
+# 2: vg is clustered - clvmd (old)
+# 3: vg has system_id (new)
+# 4: vg has tagging (old)
+get_VG_access_mode() {
+	local access_mode
+	local vg_locktype
+	local vg_clustered
+	local vg_systemid
+	local vg_tags
+
+	# Use -o reporting fields to get multiple bits of info from a single command
+	kvs=$(vgs --foreign --nolocking --noheadings --nameprefixes \
+		--rows --config report/compact_output=0 \
+		-o name,clustered,lock_type,systemid,tags ${VG} 2>/dev/null | tr -d \')
+	export ${kvs}
+	vg_locktype=${LVM2_VG_LOCK_TYPE}
+	vg_clustered=${LVM2_VG_CLUSTERED}
+	vg_systemid=${LVM2_VG_SYSTEMID}
+	vg_tags=${LVM2_VG_TAGS}
+
+	# We know this VG is using lvmlockd if the lock type is dlm.
+	if [ "$vg_locktype" = "dlm" ]; then
+		access_mode=1
+	elif [ "$vg_clustered" = "clustered" ]; then
+		access_mode=2
+	elif [ -n "$vg_systemid" ]; then
+		SYSTEM_ID=$(lvm systemid 2>/dev/null | cut -d':' -f2 | tr -d '[:blank:]')
+		access_mode=3
+	elif [ -n "$vg_tags" ]; then
+		# TODO:
+		# We don't have reliable way to test if tagging activation is used.
+		access_mode=4
+	else
+		access_mode=0
+	fi
+
+	return $access_mode
+}
+
+# TODO: All tagging activation code is almost copied from LVM RA!!!
+# But, the old LVM RA just uses the ordinary tags, not the "hosttag" feature
+# which may be a better method for active-inactive cluster scenario.
+#
+# We have two choice:
+# 1. Continue to use the LVM way, which may work well on old system.
+# 2. Change to use the real hosttag feature, but it looks very same
+# to systemID.
+# Anyway, we can easily change this if anyone requests with good reasons.
+
+# does this vg have our tag
+check_tags()
+{
+	local owner=$(vgs -o tags --noheadings ${VG} | tr -d '[:blank:]')
+
+        if [ -z "$owner" ]; then
+                # No-one owns this VG yet
+                return 1
+        fi
+
+        if [ "$OUR_TAG" = "$owner" ]; then
+                # yep, this is ours
+                return 0
+        fi
+
+        # some other tag is set on this vg
+        return 2
+}
+
+strip_tags()
+{
+        local tag
+
+	for tag in $(vgs --noheadings -o tags $OCF_RESKEY_volgrpname | sed s/","/" "/g); do
+                ocf_log info "Stripping tag, $tag"
+
+                # LVM version 2.02.98 allows changing tags if PARTIAL
+                vgchange --deltag "$tag" ${VG}
+        done
+
+	if [ ! -z $(vgs -o tags --noheadings ${VG} | tr -d '[:blank:]') ]; then
+                ocf_exit_reason "Failed to remove ownership tags from ${VG}"
+                exit $OCF_ERR_GENERIC
+        fi
+
+        return $OCF_SUCCESS
+}
+
+set_tags()
+{
+        case check_tags in
+        0)
+                # we already own it.
+                return $OCF_SUCCESS
+                ;;
+        2)
+                # other tags are set, strip them before setting
+                if ! strip_tags; then
+                        return $OCF_ERR_GENERIC
+                fi
+                ;;
+        *)
+                : ;;
+        esac
+
+        if ! vgchange --addtag $OUR_TAG ${VG} ; then
+                ocf_log err "Failed to add ownership tag to ${VG}"
+                return $OCF_ERR_GENERIC
+        fi
+
+        ocf_log info "New tag \"${OUR_TAG}\" added to ${VG}"
+        return $OCF_SUCCESS
+}
+
+# Parameters:
+# 1st: config item name
+# 2nd: expected config item value
+config_verify()
+{
+	local name=$1
+	local expect=$2
+	local real=""
+
+	real=$(lvmconfig "$name" | cut -d'=' -f2)
+	if [ "$real" != "$expect" ]; then
+		ocf_exit_reason "config item $name: expect=$expect but real=$real"
+		exit $OCF_ERR_CONFIGURED
+
+	fi
+
+	return $OCF_SUCCESS
+}
+
+lvmlockd_check()
+{
+	config_verify "global/use_lvmlockd" "1"
+	config_verify "global/locking_type" "1"
+
+	# We recommend to activate one LV at a time so that this specific volume
+	# binds to a proper filesystem to protect the data
+	# TODO:
+	# Will this warn message be too noisy?
+	if [ -z "$LV" ]; then
+		ocf_log warn "You are recommended to activate one LV at a time or use exclusive activation mode."
+	fi
+
+	# Good: lvmlockd is running, and clvmd is not running
+	if ! pgrep lvmlockd >/dev/null 2>&1 ; then
+		ocf_exit_reason "lvmlockd daemon is not running!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	if pgrep clvmd >/dev/null 2>&1 ; then
+		ocf_exit_reason "clvmd daemon is running unexpectedly."
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	return $OCF_SUCCESS
+}
+
+clvmd_check()
+{
+	config_verify "global/use_lvmetad" "0"
+	config_verify "global/use_lvmlockd" "0"
+	config_verify "global/locking_type" "3"
+
+	# TODO:
+	# David asked a question: does lvchange -aey works well enough with clvmd?
+	#
+	# Corey said: I think it does work well enough. We do a fair amount of exclusive
+	# activation clvm testing, and my experience is you'll get the LV activated on
+	# the node you ran the command on. But, I think the specific scenario and issue
+	# that surprised us all was when the LV was *already* exclusively active on say
+	# nodeA, and nodeB then attempts to also exclusively activate it as well. Instead
+	# of failing, the activation succeeds even though nodeB activation didn't occur.
+	# This is documented in the following bug:
+	#	https://bugzilla.redhat.com/show_bug.cgi?id=1191724#c8
+	# Technically, you're not guaranteed to have it activated on the node you run
+	# the cmd on, but again, that's not been my experience.
+	#
+	# Eric: Put the interesting discussion here so that we can be more careful on this.
+
+	# Good: clvmd is running, and lvmlockd is not running
+	if ! pgrep clvmd >/dev/null 2>&1 ; then
+		ocf_exit_reason "clvmd daemon is not running!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	if pgrep lvmetad >/dev/null 2>&1 ; then
+		ocf_exit_reason "Please stop lvmetad daemon when clvmd is running."
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	if pgrep lvmlockd >/dev/null 2>&1 ; then
+		ocf_exit_reason "lvmlockd daemon is running unexpectedly."
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	return $OCF_SUCCESS
+}
+
+systemid_check()
+{
+	local source
+
+	# system_id_source is set in lvm.conf
+	source=$(lvmconfig 'global/system_id_source' 2>/dev/null | cut -d"=" -f2)
+	if [ "$source" = "" ] || [ "$source" = "none" ]; then
+		ocf_exit_reason "system_id_source in lvm.conf is not set correctly!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	if [ -z ${SYSTEM_ID} ]; then
+		ocf_exit_reason "local/system_id is not set!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	return $OCF_SUCCESS
+}
+
+# Verify tags setup
+tagging_check()
+{
+        # The volume_list must be initialized to something in order to
+        # guarantee our tag will be filtered on startup
+        if ! lvm dumpconfig activation/volume_list; then
+                ocf_log err  "LVM: Improper setup detected"
+                ocf_exit_reason "The volume_list filter must be initialized in lvm.conf for exclusive          activation without clvmd"
+		exit $OCF_ERR_CONFIGURED
+        fi
+
+        # Our tag must _NOT_ be in the volume_list.  This agent
+        # overrides the volume_list during activation using the
+        # special tag reserved for cluster activation
+        if lvm dumpconfig activation/volume_list | grep -e "\"@${OUR_TAG}\"" -e "\"${VG}\"";  then
+                ocf_log err "LVM:  Improper setup detected"
+                ocf_exit_reason "The volume_list in lvm.conf must not contain the cluster tag, \"${OUR_TAG}\",   or volume group, ${VG}"
+		exit $OCF_ERR_CONFIGURED
+        fi
+
+        return $OCF_SUCCESS
+}
+
+lvm_validate() {
+	local lv_count
+	local mode
+
+	check_binary pgrep
+	# Every LVM command is just symlink to lvm binary
+	check_binary lvm
+	check_binary dmsetup
+
+	if ! vgs --foreign ${VG} >/dev/null 2>&1 ; then
+		# stop action exits successfully if the VG cannot be accessed...
+		if [ $__OCF_ACTION = "stop" ]; then
+			ocf_log warn "VG [${VG}] cannot be accessed, stop action exits successfully."
+			exit $OCF_SUCCESS
+		fi
+
+		ocf_exit_reason "Volume group[${VG}] doesn't exist, or not visible on this node!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	# Get the access mode from VG metadata and check if it matches the input
+	# value. Skip to check "tagging" mode because there's no reliable way to
+	# automatically check if "tagging" mode is being used.
+	get_VG_access_mode
+	mode=$?
+	if [ $VG_access_mode -ne 4 ] && [ $mode -ne $VG_access_mode ]; then
+		ocf_exit_reason "The specified vg_access_mode doesn't match the mode on VG metadata!"
+		exit $OCF_ERR_ARGS
+	fi
+
+	# Nothing to do if the VG has no logical volume
+	lv_count=$(vgs --foreign -o lv_count --noheadings ${VG} 2>/dev/null)
+	if [ $lv_count -lt 1 ]; then
+		ocf_exit_reason "Volume group [$VG] doesn't contain any logical volume!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
+	# Check if the given $LV is in the $VG
+	if [ -n "$LV" ]; then
+		OUT=$(lvs --foreign --noheadings ${VG}/${LV} 2>&1)
+		if [ $? -ne 0 ]; then
+			ocf_log err "lvs: ${OUT}"
+			ocf_exit_reason "LV ($LV) is not in the given VG ($VG)."
+			exit $OCF_ERR_ARGS
+		fi
+	fi
+
+	# VG_access_mode specific checking goes here
+	case ${VG_access_mode} in
+	1)
+		lvmlockd_check
+		;;
+	2)
+		clvmd_check
+		;;
+	3)
+		systemid_check
+		;;
+
+	4)
+		tagging_check
+		;;
+	*)
+		ocf_exit_reason "Incorrect VG access mode detected!"
+		exit $OCF_ERR_CONFIGURED
+	esac
+
+	if [ $? -ne $OCF_SUCCESS ]; then
+		ocf_exit_reason "Improper configuration issue is detected!"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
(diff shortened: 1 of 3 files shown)


renzhengeek:
I don't want this PR to be merged very soon. I hope this PR is opened for some time to accept inputs and further enhancements.

heartbeat/LVM-activate
```diff
@@ -0,0 +1,830 @@
+#!/bin/bash
```
dmuhamedagic: Looks like no bash-specific features are used. Better go with POSIX sh (/bin/sh).

heartbeat/LVM-activate
```diff
@@ -0,0 +1,830 @@
...
+# to manage lvmlockd daemon. In addition, this new resource agent "LVM-activate"
+# is created to take care of LVM activation/deactivation work. This agent supports
+# the new features: lvmlockd and system ID, and also supports the old features:
+# clvmd and lvm tag.
+#
+# For LVM activation, we already have an old resource agent "LVM" that supports
+# LVM1, clvmd and lvm tag. Some reasons why we tend to create a new one rather
+# than changing the old:
+# 1. "LVM" supports clvmd and some legacy things. The code is already rather complex,
+# which make it difficult to change and ensure everything works correctly.
```
dmuhamedagic: Just took a look at LVM, it doesn't look so complex to me ;-)

heartbeat/LVM-activate
```diff
@@ -0,0 +1,830 @@
...
+#
+# 2. We want this new RA to do the activation work in a much more safe way by
+# leveraging the new features and activating at VG or LV basis depending on
+# your choice.
```
dmuhamedagic: But couldn't there still be a way to update the existing RA? The problem with having two RAs: a) maintenance effort, b) introduction of issues which were already fixed due to extensive field testing, c) users confused even more than necessary (i.e. which RA should I use?). The two points above don't seem to me to be arguments strong and clear enough for the introduction of a new RA.

heartbeat/LVM-activate
renzhengeek: Dejan,
Will have a try!

thanks very much for your review.

heartbeat/LVM-activate
renzhengeek: Good point. We (the redhat developers and me) initially discussed this concern. We think it would be almost complete rewrite of the old LVM RA if we want to support the new features and add some more parameters for the RA. So we decided to create a new one to see how much difficult to merge them.

Yes, it's time to discuss if we should keep this new one or merge it with the old one. My concerns are:
1. Is it acceptable to change the parameters of the old LVM?

2. Are there some principle rules to follow when I start to merge such a huge change to the old LVM RA?

teigland: maintenance effort:  the new RA is much simpler to maintain, and the old RA should not require  further maintenance.

introduction of issues: this is a reason to leave the old RA alone while the new one is phased in.  The vast majority of the junk in the old RA has no value and should be removed, not legitimate fixes.

users confused: changing the parameters on the same RA would be the most confusing it seems to me.

which RA should I use: the new one.

dmuhamedagic:
> maintenance effort:  the new RA is much simpler to maintain, and the old RA should not require  further maintenance.

How would you know that? Have a crystal ball? ;-) If you think
that running clusters will drop the old RA at once, on
developers' order, then you should definitely reconsider.

> introduction of issues: this is a reason to leave the old RA alone while the new one is phased in.  The vast majority of the junk in the old RA has no value and should be removed, not legitimate fixes.

Not an expert on LVM, neither the RA nor the underlying tools, so
cannot give a qualified opinion, but I'd sort of expect that the
code is there for a reason. Or at least once was. If there's
stuff without value that should be removed, maybe a pull request
or an issue so that someone can remove it.

I'm not against introducing a new RA in principle, but would like
to be convinced that it is definitely the better way forward.

dmuhamedagic:
> Yes, it's time to discuss if we should keep this new one or merge it with the old one. My concerns are:
> 1. Is it acceptable to change the parameters of the old LVM?

Unfortunately, there's no good procedure to retire or modify the
existing RA parameters.

But looking at the parameters you proposed, I don't see any
problems in that department. Or did I miss something?

> 2. Are there some principle rules to follow when I start to merge such a huge change to the old LVM RA?

Most important is not to introduce regressions. Other than that,
just follow good software engineering principles :) There are
most probably other things, but nothing occurs to me right now.

dmuhamedagic:
> I would be much happier with a new RA that I can migrate to on my own
> time when I can study the changes and behaviour than having the way the
> existing RA behaves change all of a sudden on a patch cycle.  

Introducing major changes in the default behaviour of the
existing RA is definitely not an option, so that won't happen. If
that's the only way forward, then we can only create a new RA.

> Especially if new parameters are suddenly expected. 

That'd be a regression, so that won't (shouldn't) happen either.

> I have 100s of LVs on a dozen clusters and being able to chose when I
> move to the new RA (and I love the fact that there is a new one, the old
> one is a mess) is important to me.

This is the third time somebody says the existing LVM is a mess,
or thereabouts, but it'd be great to see which part is a mess or
is it that all of it is a mess and we're doomed ;-)

teigland:
> Not an expert on LVM, neither the RA nor the underlying tools, so
> cannot give a qualified opinion, 

I think it should come as no surprise that some expertise in LVM is needed
to appreciate the new LVM mechanisms and implementation.

renzhengeek:
> Introducing major changes in the default behaviour of the existing RA is definitely not an option, so that won't happen. If that's the only way forward, then we can only create a new RA.

It's also my feeling. If we choose to update the existing LVM RA, it seems inevitable to bring a lot of changes, because the LVM RA code doesn't have a design/logic to add new functionality.

As David suggested, we tend to leave the LVM RA alone while the new one is phased in. As you maybe aware,  the new LVM-activate also support the functionality of the old. My thought is: let's focus on the new RA interface design (parameters, especially the "required" parameters), and the quality of supporting the new functionality. I'm not saying that the old functionality is not important because as said users can continue use the old.

Yes, the old functionality in the new LVM-activate may not consider various special cases that the old LVM RA does at this moment. "git blame LVM" can tell us there are many fixes for various issue historically. Some fixes are still valid, but I feel some maybe not. So, I think I can start to study more carefully about the old LVM by "git blame" to see if there is something useful fixes we should forward port to the new LVM-activate.

renzhengeek:
>Unfortunately, there's no good procedure to retire or modify the existing RA parameters. But looking at the parameters you proposed, I don't see any problems in that department. Or did I miss something?

Sorry, I didn't get the point. You mean you don't like the proposed parameters? If so, please tell which one? or the reason perhaps?

> Most important is not to introduce regressions. ...
Yeah, thanks for you comments. I'd say that's why we tend to add a new one, rather than updating the existing one :)

renzhengeek:
> This is the third time somebody says the existing LVM is a mess, or thereabouts, but it'd be great to see which part is a mess or is it that all of it is a mess and we're doomed ;-)

For example, in the existing LVM, the way it uses to check if all LVs in the given VG are active is apparently wrong:)

heartbeat/LVM-activate
renzhengeek: Hi Dejan,

After changing to /bin/sh, the checker gives some warnings:

```
In LVM-activate line 694:
	[[ -z ${LV} ]] && vol=${VG} || vol=${VG}/${LV}
        ^-- SC2039: In POSIX sh, [[ ]] is undefined.
...
In LVM-activate line 728:
	local vol
        ^-- SC2039: In POSIX sh, 'local' is undefined.
```

Sorry for my laziness, I'd still use /bin/bash, though I know it's easy to change. But, the previous new RA lvmlockd already uses /bin/bash. Someone can easily change to /bin/sh together if there's a real problem for some cases.

renzhengeek:
@krig

Hi kristoffer, this RA has been reviewed by David and others for several rounds. I'd like to ask you to give some further review especially on the RA specification like error handing and return code,etc.

If it's OK, please merge the PR so that we can have QA to test this, and advertise this in clusterlab list.

dmuhamedagic:
On Sun, Oct 22, 2017 at 11:30:57PM -0700, Eric Ren wrote:
> In LVM-activate line 694:
> 	[[ -z ${LV} ]] && vol=${VG} || vol=${VG}/${LV}
>         ^-- SC2039: In POSIX sh, [[ ]] is undefined.

That's easy to fix, just use [ instead of [[.

> In LVM-activate line 728:
> 	local vol
>         ^-- SC2039: In POSIX sh, 'local' is undefined.

We do tolerate local, you can ignore that warning.

> Sorry for my laziness, I'd still use /bin/bash, though I know
> it's easy to change. But, the previous new RA lvmlockd already
> uses /bin/bash.

LVM is one of the basic RAs. It'd be really good to have an
option not to use bash.

> Someone can easily change to /bin/sh together
> if there's a real problem for some cases.

How can they change to /bin/sh? By forking the RA?

dmuhamedagic:
On Mon, Oct 23, 2017 at 06:48:08AM +0000, Eric Ren wrote:
> @krig
> 
> Hi kristoffer, this RA has been reviewed by David and others for several rounds. I'd like to ask you to give some further review especially on the RA specification like error handing and return code,etc.
> 
> If it's OK, please merge the PR so that we can have QA to test this, and advertise this in clusterlab list.

Please not so fast. I didn't notice that we reached any
conclusion.

dmuhamedagic:
On Mon, Oct 16, 2017 at 10:51:31AM -0700, David Teigland wrote:
> 
> I think it should come as no surprise that some expertise in LVM is needed
> to appreciate the new LVM mechanisms and implementation.

There is another option: ship the new RA with lvm2 or wherever
the lvm tools live. There is no rule which says that every RA
needs to be part of the resource-agents. The OCF standard allows
different providers. That way you can implement the RA however
you feel like. Looks like that would be for the best.

dmuhamedagic:
On Wed, Oct 18, 2017 at 07:59:19AM +0000, Eric Ren wrote:
> Sorry, I didn't get the point. You mean you don't like the proposed parameters? If so, please tell which one? or the reason perhaps?

What I meant is that there is just volgrpname to vgname renamed.
Of course, that is not permissible, but I suppose that it is not
absolutely necessary to change the existing parameter names.

> > Most important is not to introduce regressions. ...
> Yeah, thanks for you comments. I'd say that's why we tend to add a new one, rather than updating the existing one :)

Or make an effort not to introduce regressions ;-)

dmuhamedagic:
On Wed, Oct 18, 2017 at 08:01:58AM +0000, Eric Ren wrote:
> > This is the third time somebody says the existing LVM is a mess, or thereabouts, but it'd be great to see which part is a mess or is it that all of it is a mess and we're doomed ;-)
> 
> For example, in the existing LVM, the way it uses to check if all LVs in the given VG are active is apparently wrong:) 

If it's wrong then it needs to be fixed, right? But that still
isn't really telling about the supposed "mess".

renzhengeek:
> How can they change to /bin/sh? By forking the RA?

OK, let me fix this.

renzhengeek:
> Please not so fast. I didn't notice that we reached any conclusion.

OK. I've been waiting for several days for response like this :) To be honest, I'm confused what you expect. I try to list my understanding below, please correct me if I'm wrong:

1. Updating the old LVM, rather than introduce the new RA?
As you said previously,
<Introducing major changes in the default behaviour of the existing RA is definitely not an option, so that won't happen. If that's the only way forward, then we can only create a new RA.

It's exactly the current situation.

2. Don't change the parameter unnecessarily.
I agree, so we don't choose to change exiting LVM. There is reason why we choose to change/add parameters for the new RA: 
1. "lvname" for individual LV activation, and you will see the real need in this thread [1];
2. "vg_access_mode": there're several solution to access VG/LV in clsuter, we want user to explicitly know which one they're using to decrease possible mis-configuration.
3. Because we already made the changes in 1) and 2), we think it doesn't hurt to change the "volgrpname" to "vgname". It's obvious which name looks better.

3. More maintanance efforts?
Yes, everything new will bring this kind of burden. The point is if it's worthy the effort. You never know the answer until you try the new thing. In LVM community, it has been decided that the new replacements will come in while the old will retire increasingly.


[1] [Shared VG, Separate LVs](https://www.redhat.com/archives/linux-lvm/2017-October/msg00001.html)

renzhengeek:
> There is another option: ship the new RA with lvm2 or wherever the lvm tools live. There is no rule which says that every RA needs to be part of the resource-agents. The OCF standard allows different providers. That way you can implement the RA however you feel like. Looks like that would be for the best.

Thanks for proposal like this, though I'm trying hard to avoid this, but it's good to have specific suggestion rather than just saying no :)

Before giving up pushing this RA into resource-agent, I'd ask, are there very good reasons suggesting the new RAs are bad for the resource-agent project?

As you know, there are both IPaddr and IPaddr2 in resource-agent. I think they may share the similar reason with this new RAs.

renzhengeek:
> If it's wrong then it needs to be fixed, right? But that still isn't really telling about the supposed "mess".

You're right. I can fix it in another PR. It's just one piece of mess, which has been there for years without fix and just proper TODO/FIXME comments in the code. I think I can list more mess, but I'm not interested in the old one at this moment. I need to focus on how to support the new LVM features now :)

krig:
As long as it's a new agent and not changes to the old one, I don't see how it could be introducing regressions?

I'm trying to catch up on the discussion, but it seems there are some comments missing from the github interface.. at least I'm finding it a little bit difficult to follow :) I'm assuming that the /bin/sh incompatibilities have been fixed, right?

krig:
For the maintenance effort comments: Given that a new agent is proposed and supported by the upstream developers due to not wanting to support the existing agent, perhaps keeping the old agent around and trying to support it without any help from upstream is an excessive maintenance burden. In that case I would be in favor of merging this new agent and marking the old agent as deprecated...

I would like it if we had some deprecation process where an agent can be removed after being deprecated for some time.

renzhengeek:
> In that case I would be in favor of merging this new agent and marking the old agent as deprecated...

We've done some kinda of this work by forward porting the functionalities of the old agent into the new one. I think it may help make the process of retiring the old smoothly.

However, I feel we have to keep the old around until most users move to use the new one.

dmuhamedagic:
On Mon, Oct 23, 2017 at 03:27:30AM -0700, Eric Ren wrote:
> > Please not so fast. I didn't notice that we reached any conclusion.
> 
> OK. I've been waiting for several days for response like this :) To be honest, I'm confused what you expect. I try to list my understanding below, please correct me if I'm wrong:

And I've been waiting to some information which was not forthcoming :)
In particular: what are the differences in how the two RA
implement the _existing_ functionaility. In other communication
I gave an example of the monitor action where I described how it
works in the existing RA and how you tried to implement it. That
is essential in deciding whether to merge or make a new RA.

> 1. Updating the old LVM, rather than introduce the new RA?
> As you said previously,
> <Introducing major changes in the default behaviour of the existing RA is definitely not an option, so that won't happen. If that's the only way forward, then we can only create a new RA.
> 
> It's exactly the current situation.

Well, not necessarily. That's something I couldn't yet divulge.

> 2. Don't change the parameter unnecessarily.
> I agree, so we don't choose to change exiting LVM. There is reason why we choose to change/add parameters for the new RA: 
> 1. "lvname" for individual LV activation, and you will see the real need in this thread [1];
> 2. "vg_access_mode": there're several solution to access VG/LV in clsuter, we want user to explicitly know which one they're using to decrease possible mis-configuration.

Introducing new parameters is not a problem.

> 3. Because we already made the changes in 1) and 2), we think it doesn't hurt to change the "volgrpname" to "vgname". It's obvious which name looks better.

I'd tend to agree, but it's just a name.

> 3. More maintanance efforts?
> Yes, everything new will bring this kind of burden. The point is if it's worthy the effort. You never know the answer until you try the new thing. In LVM community, it has been decided that the new replacements will come in while the old will retire increasingly.

That's fine, but note that the maintenance burden is essentially
with the resource-agents people. I can name only pgsql as the RA
which is maintained mostly by NTT. People come and go and the RA
remains and whoever is in charge in resource-agents have to deal
with it.

dmuhamedagic:
On Mon, Oct 23, 2017 at 10:41:00AM +0000, Eric Ren wrote:
> > There is another option: ship the new RA with lvm2 or wherever the lvm tools live. There is no rule which says that every RA needs to be part of the resource-agents. The OCF standard allows different providers. That way you can implement the RA however you feel like. Looks like that would be for the best.
> 
> Thanks for proposal like this, though I'm trying hard to avoid this, but it's good to have specific suggestion rather than just saying no :)

I am not saying no, I am trying to find out what would be for the
best for the upstream resource-agents project.

> Before giving up pushing this RA into resource-agent, I'd ask, are there very good reasons suggesting the new RAs are bad for the resource-agent project?

It makes additional work for the resource-agents project
maintainers. The maintainers of resource-agents don't necessarily
have the expertise to deal with it. I've had to look into
technologies ranging from mysql and oracle to iscsi and
filesystems. That's not always fun.

> As you know, there are both IPaddr and IPaddr2 in resource-agent. I think they may share the similar reason with this new RAs.

I was not around when IPaddr2 came to be, but I heard more than
once from people who were that introducing it was a veritable
disaster. The fact that it exists doesn't mean that it was a good
decision.

dmuhamedagic:
On Mon, Oct 23, 2017 at 03:48:06AM -0700, Eric Ren wrote:
> > If it's wrong then it needs to be fixed, right? But that still isn't really telling about the supposed "mess".
> 
> You're right. I can fix it in another PR. It's just one piece of mess, which has been there for years without fix and just proper TODO/FIXME comments in the code. I think I can list more mess, but I'm not interested in the old one at this moment. I need to focus on how to support the new LVM features now :)

I'm sure that there are parts where LVM is not pretty, but that's
what eventually most software ends like, reflecting reality which
is not as smooth as we'd like it to be. And, in particular, an
RA, trying to bring together two different worlds, is very prone
to ugliness. Note also that some of that ugliness happens to be
necessary and that we'll end up repeating it in whatever comes in
as a replacement. Having worked with resource agents for many
years, I believe to be in a position to claim that with some
authority.

dmuhamedagic:
On Mon, Oct 23, 2017 at 11:44:33AM +0000, Kristoffer Grönlund wrote:
> For the maintenance effort comments: Given that a new agent is proposed and supported by the upstream developers due to not wanting to support the existing agent, perhaps keeping the old agent around and trying to support it without any help from upstream is an excessive maintenance burden. In that case I would be in favor of merging this new agent and marking the old agent as deprecated...
> 
> I would like it if we had some deprecation process where an agent can be removed after being deprecated for some time.

Unfortunately we don't have any process in place. It would also
be rather difficult to devise one. Essentially, right now we can
only put a big warning in the meta-data and to be logged on
starts and then hope for the best ;->

dmuhamedagic:
On Mon, Oct 23, 2017 at 12:09:08PM +0000, Eric Ren wrote:
> > In that case I would be in favor of merging this new agent and marking the old agent as deprecated...
> 
> We've done some kinda of this work by forward porting the functionalities of the old agent into the new one. I think it may help make the process of retiring the old smoothly.
> 
> However, I feel we have to keep the old around until most users move to use the new one.

Right. And then you never know if "most users" moved to the new
one. Heck, there are still heartbeat v1 clusters around.

renzhengeek:
> In particular: what are the differences in how the two RA implement the _existing_ functionaility. In other communication I gave an example of the monitor action where I described how it works in the existing RA and how you tried to implement it. That is essential in deciding whether to merge or make a new RA.

Ah, if I understand correctly, that seems the main disagreements between us:

**You care more about the implementation of the existing functionalities in the new RA, while I focus more on the new functionalities.**

Let me explain the current status and the plan about this point. In the new RA, the existing functionalities are basically the same as in the old one.  According to my testing, the existing functionalities work well in the new RA. The main difference are:

1. The existing functionalities in the new RA can activate only one given LV now, of course it's optional.
2. The implementing code logical of the existing functionalities are basically copied from the old one,  but dropped some conner-case handling, overhead lvm commands like vgck and the lvm status method, etc.

There are two option we can pick:
1. Drop the existing functionalities support in the new RA. Users can use the LVM for old functionalities, while use the new RA for new features;
2. Keep the existing functionalities in the new RA. We can improve the quality of the existing functionalities while the new RA evolves. At a certain time, we can safely drop the old RA.

Personally, I prefer the second option.

renzhengeek:
> It makes additional work for the resource-agents project maintainers. The maintainers of resource-agents don't necessarily have the expertise to deal with it. I've had to look into technologies ranging from mysql and oracle to iscsi and filesystems. That's not always fun.

Totally understand the pain. It's a big nature of the resource-agent project - a set of service agents for various fields. It's _not_ easy to maintain such a project because the maintainer is ideally expected to be advanced user in many fields :)

dmuhamedagic:
On Tue, Oct 24, 2017 at 05:45:04AM +0000, Eric Ren wrote:
> > In particular: what are the differences in how the two RA implement the _existing_ functionaility. In other communication I gave an example of the monitor action where I described how it works in the existing RA and how you tried to implement it. That is essential in deciding whether to merge or make a new RA.
> 
> Ah, if I understand correctly, that seems the main disagreements between us:

I don't see any disagreement, what I mostly see is misunderstanding :)

> **You care more about the implementation of the existing functionalities in the new RA, while I focus more on the new functionalities.**

Yes! That's the point which is crucial.

> Let me explain the current status and the plan about this point. In the new RA, the existing functionalities are basically the same as in the old one.

But it's not! At least not in the status/monitor action. Maybe
your intention was the _same_, but the way they work is
_different_. I didn't look at the other actions.

However, that difference is really what at one point got removed
from the LVM RA because of the false positives.

> According to my testing, the existing functionalities work well in the new RA.

Please note that there are _many_ storage configurations out
there which differ in any number of ways. Many a bug was fixed
over the years of the field testing which was done by the users.

> The main difference are:
> 
> 1. The existing functionalities in the new RA can activate only one given LV now, of course it's optional.
> 2. The implementing code logical of the existing functionalities are basically copied from the old one,  but dropped some conner-case handling, overhead lvm commands like vgck and the lvm status method, etc.
> 
> There are two option we can pick:
> 1. Drop the existing functionalities support in the new RA. Users can use the LVM for old functionalities, while use the new RA for new features;
> 2. Keep the existing functionalities in the new RA. We can improve the quality of the existing functionalities while the new RA evolves. At a certain time, we can safely drop the old RA.

Going back to the IPaddr/IPaddr2 case: IPaddr2 is based on ip(8)
whereas IPaddr uses ifconfig(8). That was the main reason that
IPaddr2 got to be IPaddr2. Now, do we at least have some similar
change here?

> Personally, I prefer the second option.

A personal preference may not be to the benefit of the project ;-)

renzhengeek:
> Please note that there are _many_ storage configurations out there which differ in any number of ways. Many a bug was fixed over the years of the field testing which was done by the users.

You're right. That's also the reason why I ask to leave the old RA alone. It's _not_ a easy work to verify every commit/bug fix listed by git blame to validate if it's still needed or it's stale. I'd suggest the users to still use the old LVM for old functionalities. It's not a problem.

>But it's not! At least not in the status/monitor action. Maybe your intention was the _same_, but the way they work is _different_. I didn't look at the other actions.

Some difference is intentional, it's not hard to keep the nearly the same as the old one, but as said, I don't want forward port messy to the new one.

> However, that difference is really what at one point got removed from the LVM RA because of the false positives.

Pity, that guy then introduced a big obvious bug - if there's any LV active, it assumes the whole VG is active, which is a cheat to the HA manager. I think the first priority is doing it right, then try to do it better.

> Going back to the IPaddr/IPaddr2 case: IPaddr2 is based on ip(8) whereas IPaddr uses ifconfig(8). That was the main reason that IPaddr2 got to be IPaddr2. Now, do we at least have some similar change here?

I think, yes. Please read about the new features I provided some references in the previous mail :)

> A personal preference may not be to the benefit of the project ;-)

Right, so I said "personally".

krig:
> Pity, that guy then introduced a big obvious bug - if there's any LV active, it assumes the whole VG is active, which is a cheat to the HA manager. I think the first priority is doing it right, then try to do it better.

If there's a known bug in the existing RA, we should fix that as well (regardless of the status of the new RA).

oalbrigt:
I think it makes sense to keep this as a separate agent to avoid introducing what might cause new issues in our current agent, so it can get improved and become the preferred one later when it's proven to be better than the old agent.

dmuhamedagic:
On Tue, Oct 24, 2017 at 04:51:43AM -0700, Eric Ren wrote:
> > Please note that there are _many_ storage configurations out there which differ in any number of ways. Many a bug was fixed over the years of the field testing which was done by the users.
> 
> You're right. That's also the reason why I ask to leave the old RA alone. It's _not_ a easy work to verify every commit/bug fix listed by git blame to validate if it's still needed or it's stale. I'd suggest the users to still use the old LVM for old functionalities. It's not a problem.

I'm afraid that you're missing the point here. It's not about
pruning the existing RA, it's about having to eventually fix the
new one once users start using it. (Not that anything is wrong
with pruning if anybody has time for that.)

> >But it's not! At least not in the status/monitor action. Maybe your intention was the _same_, but the way they work is _different_. I didn't look at the other actions.
> 
> Some difference is intentional, it's not hard to keep the nearly the same as the old one, but as said, I don't want forward port messy to the new one.
> 
> > However, that difference is really what at one point got removed from the LVM RA because of the false positives.
> 
> Pity, that guy then introduced a big obvious bug - if there's any LV active, it assumes the whole VG is active, which is a cheat to the HA manager. I think the first priority is doing it right, then try to do it better.

It's not. The existing LVM operates on the VG level. One active
LV means that the VG is active. It could be that there was
already a discussion about that.

> > Going back to the IPaddr/IPaddr2 case: IPaddr2 is based on ip(8) whereas IPaddr uses ifconfig(8). That was the main reason that IPaddr2 got to be IPaddr2. Now, do we at least have some similar change here?
> 
> I think, yes. Please read about the new features I provided some references in the previous mail :)

I read about the new features, but, and saying it again, this is
not about new features. Adding new features/functionality is not
a problem because it cannot (or shouldn't, normally) introduce
such changes which would surprise existing configurations.

teigland:
> if there's any LV active, it assumes the whole VG is active

I can see reasons why someone may or may not want lvm_status to require
all LVs be active in the VG.  Even when you are activating an entire VG,
you can set lvm.conf volume_list to only activate selected LVs.  And we
have mentioned that running lvm commands in status can cause an unwanted
amount of disk scanning.  This may call for an option to configure one
behavior or the other.  The default should probably remain as it has been:
any number of active LVs in the VG passes, and for that we only need to
use dmsetup.

renzhengeek:
> I'm afraid that you're missing the point here. It's not about pruning the existing RA, it's about having to eventually fix the new one once users start using it. (Not that anything is wrong with pruning if anybody has time for that.)

I know, and I also see some common point of our comments:
1. We don't need to worry much about the regression when the old RA is kept around;
2. The old functionalities in the new RA need much improvements and bugfixes when people start to use the new RA. Yes, you're right, but I think you don't need to worry so much, because:
  - There will be a long time for the old RA retires while the new RA becomes mature. We should encourage user to new RA if they want to try new features, while we recommend people still use old RA for old functionalities. There is no reason for users have both at the same time.
 - Yes, it's still possible for users to use the old functionalities in the new RA especially when the old RA goes away. If there's bug, I think we can fix. I wish I can provide a perfect new RA at once, but...

renzhengeek:
> It's not. The existing LVM operates on the VG level. One active LV means that the VG is active. It could be that there was already a discussion about that.

Most time it's not a problem. But, if some of LVs in the VG are inactive, the monitor still reports OK. The pacemaker will not try to activate the inactive LVs.

renzhengeek:
> Even when you are activating an entire VG, you can set lvm.conf volume_list to only activate selected LVs.

Oh, I didn't notice that.

> we have mentioned that running lvm commands in status can cause an unwanted amount of disk scanning. 

Yes, it's a problem when lvmetad cannot be used with clvmd.

> This may call for an option to configure one behavior or the other. The default should probably remain as it has been: any number of active LVs in the VG passes, and for that we only need to use dmsetup.

Hmm, I hesitate to add more option. So, I think we should compromise to only use dmsetup in lvm_status(). @dmuhamedagic Sorry, you're right here :)

renzhengeek:
Another problem I see in the existing LVM RA is that people often neither use "clvmd" nor "tag-based exclusively" activation in their HA cluster configuration when they use "active-passive" model with local filesytem on top of the LV. It looks dangerous to me, doesn't it? The LV device will be activated on both nodes without cluster filesystem's protection. I've seen several such bug reports from users that their LVM metadata is corrupted.

But, I'm wondering why the LVM RA didn't try to verify if the configuration is safe? User often assume it's safe to use once the RA is started OK...

dmuhamedagic:
I'd assume that people read the documentation before deploying
the storage and are aware of the possible consequences of their
choices. If somebody wants to run LVM without protection and rely
on other mechanisms, we should let them.

dmuhamedagic:
On Wed, Oct 25, 2017 at 10:14:22AM +0000, Eric Ren wrote:
> > I'm afraid that you're missing the point here. It's not about pruning the existing RA, it's about having to eventually fix the new one once users start using it. (Not that anything is wrong with pruning if anybody has time for that.)
> 
> I know, and I also see some common point of our comments:
> 1. We don't need to worry much about the regression when the old RA is kept around;
> 2. The old functionalities in the new RA need much improvements and bugfixes when people start to use the new RA. Yes, you're right, but I think you don't need to worry so much, because:
>   - There will be a long time for the old RA retires while the new RA becomes mature. We should encourage user to new RA if they want to try new features, while we recommend people still use old RA for old functionalities. There is no reason for users have both at the same time.
>  - Yes, it's still possible for users to use the old functionalities in the new RA especially when the old RA goes away. If there's bug, I think we can fix. I wish I can provide a perfect new RA at once, but...

That's neither expected nor possible. But you just rephrased what
I said above ;-) The question remains why go again through the
process.

I'm refactoring the LVM RA, hope that we'll have a better picture
afterwards.

dmuhamedagic:
There's a PR #1044 which should put a bit more clarity into the existing LVM. I guess that the RA by @renzhengeek can be refactored in the same way. If not, then we can adjust/improve the interface between the RA and the various mechanisms. @renzhengeek please take a look at lvm-plain.sh for more details. I believe that this way we can keep the "messiness" from taking over our work ;-)

renzhengeek:
> I'm refactoring the LVM RA, hope that we'll have a better picture afterwards.

Appreciate your work, and have had some comments in your PR.

Actually in my PR, I want to focus on the new feature support, and don't bother much the old functionalities. On the other side, you cares much more about the old RA than the new features :-)

I think I can drop the old functionality support in the new RA...

teigland:
On Wed, Nov 01, 2017 at 09:56:03AM +0000, Eric Ren wrote:
> > I'm refactoring the LVM RA, hope that we'll have a better picture afterwards.
> 
> Appreciate your work, and have had some comments in your PR.
> 
> Actually in my PR, I want to focus on the new feature support, and don't bother much the old functionalities. On the other side, you cares much more about the old RA than the new features :-)
> 
> I think I can drop the old functionality support in the new RA...

We will be using the new RA and will want to keep all the capabilities.
I don't think anyone has any interest making unnecessary changes to the
existing RA.

dmuhamedagic:
On Wed, Nov 01, 2017 at 09:56:01AM +0000, Eric Ren wrote:
> Actually in my PR, I want to focus on the new feature support, and don't bother much the old functionalities. On the other side, you cares much more about the old RA than the new features :-)

This is not about the new or the old features, but about how do
we organize the LVM RA and whether it is preferable, for both
maintainers and users, to ship one or two RAs.

renzhengeek:
I have removed the LVM command in the lvm_status() because LVM command is too heavy to be used in status action, though it's a compromise - the VG is assumably  active if any LV of the VG is active, the RA won't response to a situation that some LVs are inactive.

renzhengeek:
Hi @dmuhamedagic,

I thought we are getting closer, but seems it is not unfortunately.  I spent much to write and test the new RA, also had pretty much discussion with David. I don't want to spend more time to adapt the new RA to your PR #1044. Neither think it deserves, nor want to push the efforts to mix the old and the new further. When there's big change in the old, it needs much testing, and I don't think you have enough testing on your change.

We are like a HA cluster. I think the cluster is brain-split now. We need to vote for the quorum. It seems to me that @teigland @oalbrigt and myself tend to agree the following points:

1. Leave the old RA alone - don't bring much change to it;
2. Let the new RA evolves to prove itself;
3. Try to support the old functionalities in the new as better as we can.

@dmuhamedagic If you still don't agree to accept this PR, please kindly list the specific TODO for me, I will try my best to fix them.

@krig Could you act as the stonith - the bad guy, hah?

dmuhamedagic:
Hi Eric,

Don't panic :) We don't need to shoot anybody and definitely not
in the head!

Let me try again to put things into perspective. You want to a)
add some new functionality and b) to retire the existing LVM.
I'm all for a) but still not convinced about the b).

You claim that a) is not possible with the existing LVM, but
whenever I asked why, there was nothing concrete forthcoming. At
any rate, with PR #1044 I'm absolutely sure that adding new
functionality would be a breeze. Again, please note that adding
new parameters, when needed, is not an issue.

b) is more interesting. After PR #1044, the core LVM code is
more than halved. More precisely, the start action relies on vgs
to check if the VG exists and vgchange to activate the VG. The
stop action utilizes only vgchange to deactivate the VG. Does the
proposed RA use other stuff? Is bringing that other stuff into
the existing RA impossible or very impractical? Can you say that
"vgchange is no good, we want to do xyz instead"?

Recall the IPaddr2 vs IPaddr: the former is based on ip(8) and
the latter on ifconfig(8). The two are completely different ways
of managing a network interface. Hence the new IPaddr2 and yet,
even though there was a very good reason for it, having two RAs
didn't really go well. Let's try to learn from past mistakes :)

krig:
@renzhengeek @dmuhamedagic If we can get the new functionality into a single RA, I am also in favor of that. I'd rather not shoot anyone in the head just yet! There is definitely a cost to introducing a new version of a resource agent, so if the proposed refactoring really does make it easier to include the new functionality into the old agent, I would also like to see that rather than adding another agent.

teigland:
> You claim that a) is not possible with the existing LVM, but
> whenever I asked why, there was nothing concrete forthcoming.

I don't believe anyone said it was not possible.  We said it was our
decision, and we have given clear reasons.  The main one: avoid risking
changes to LVM which would be needed to add the new features.  Other
reasons are to replace existing things with improved ways of doing them.
Also, I believe that some of the existing parameters are changing in terms
of being required and their defaults.

renzhengeek:
> Don't panic :) We don't need to shoot anybody and definitely not in the head!

Hah, you know it's analogy. STONITH is bad, but you have to admit sometime it's necessary.

> Let me try again to put things into perspective. You want to a) add some new functionality and b) to retire the existing LVM. I'm all for a) but still not convinced about the b). 

I like comments like this - say what you agree and disagree. 1) The exiting LVM will stay in master branch for a long time, will exit in old version of resource-agent forever, will be there in the old resource-agent RPM delivered by OS distributor. No one force user to replace the old. 2) As said, most of us tend to leave the old alone - there is an obvious reason - it's a way for the new to evolve and introducing regression into the old. 3) From the perspective of LVM project, the new features 

>You claim that a) is not possible with the existing LVM, but whenever I asked why, there was nothing concrete forthcoming. 

No one said it's impossible, it's just about choice, the *cost*, and the purpose of not introducing regression to the old.

> At any rate, with PR #1044 I'm absolutely sure that adding new functionality would be a breeze. 
Sorry, I don't want to comment much on this PR because it go against one key point above:
`leave the old alone`.

> Recall the IPaddr2 vs IPaddr: the former is based on ip(8) and the latter on ifconfig(8)

Similarly, clvmd and lvmlockd are also completely tools for managing lvm2 in cluster.

krig:
I don't think there will be consensus reached, so lets fall back to STONITH. ;)

My understanding:

@renzhengeek and @teigland want to introduce a new agent for functionality related to lvmlockd. @dmuhamedagic wants clarification around the reasons for doing this, and is skeptical about the arguments that it would not be possible to adapt the existing agent to support these changes. From what I can tell, neither side has changed their positions as a result of the (long) discussion that followed. It doesn't seem like there will be agreement.

There doesn't seem to be any technical objections to the proposed changes.

Votes in favor of a new agent: @renzhengeek @teigland @tparkercbn @oalbrigt 
Votes opposed: @dmuhamedagic 

I think the only way to move forward here is to merge this request, and to continue working on merging the two agents into a single implementation with different interfaces in future PRs like #1044.

krig:
There are some comments in the LVM-activate agent about how the old agent is "messy" which is think is unnecessary, but in the interest of moving forward I propose cleaning that up in another PR.

dmuhamedagic:
> I think the only way to move forward here is to merge this request, and to continue working on merging the two agents into a single implementation with different interfaces in future PRs like #1044.

It would be good the next time you just merge without creating a
pull request or mark the pull request as "not for discussion" or
something, so that I know not to waste my time. Thanks.

renzhengeek:
Hi Dejan,

> so that I know not to waste my time. Thanks.

Really sorry for that, but it would not be like this if you first proposed your idea before you change the code and then ask me to adapt to your change. In other words, it would take my time to adapt your change that I don't like.

Thanks for your input along the road. It really helps me to rethink something and do some changes.

krig:
> It would be good the next time you just merge without creating a
> pull request or mark the pull request as "not for discussion" or
> something, so that I know not to waste my time. Thanks.

Well, apologies, but I don't think that's fair at all: There's been a very long discussion, I've followed the whole thing. Merging a pull request is not the final word on anything. Anyone is free to open further pull requests. But I honestly don't think it's productive when a discussion reaches 60+ comments without agreement being reached and the discussion doesn't have any clear end goal. There should be a way to move forward.

My thought was that we could merge both this and your pull request, and adapt this new agent to the work you did in PR #1044 to unify the activation implementations. My understanding was that your main objection is against the proposal to retire the older LVM resource agent. Merging this PR doesn't do that. With #1044, we can even provide a common implementation for LVM and LVM-activate. That would seem to alleviate the maintenance concerns.

So far, my experience of maintaining this repository has been that waiting for consensus from all maintainers means PRs take a very long time (sometimes infinite) to get merged. Instead we do quite often revert merges if they turn out to have been mistaken. So please do revert this if you think that's the right thing to do here, but it would then be helpful to give @renzhengeek a clearer idea of what he would need to do to actually move forward.

dmuhamedagic:
I don't know what to say, I have already said a lot, perhaps too much ;-) I do hope that there will be an effort to merge the two RAs. Looking forward to further developments.

dmuhamedagic:
Hello,

I would like to say a few things regarding this matter, even though so much time has passed.

First, the LVM2 is a proverbial PoS. I am very sorry not to be able to use a less vulgar expression, for I've been carrying around a couple of SDD in a mirrored LVM2 configuration. Uncountable times I was unable to mount one or both disks. Isn't LVM supposed to be part of an HA solution? Well, for me at least, that doesn't hold. Thanks goes to the LVM2 developers (you know who you are).

Second, I would like to mention Mr. Alan Robertson. This time in a very public forum where so many people listen. I am proud to have been a pupil of his. Bar none, he is the best software engineer I have ever met.  Due to some unfortunate circumstances he was kicked out of the Linux HA and that in spite of him being its founder. It is a shame that that happened and we all carry some responsibility for that, some of us more, some of us less.  Again, the more responsible ones are going to recognize themselves. Please note that I don't want to get into the game of pointing fingers. We are all here grown up people and there is no place for that. However, it is important to say what is what and I am here to extend my thanks to Alan for all that he has done. Dear Alan, you are the best!

If you, anyone, has an issue with whatever I said, please step forward and speak. It is time for everyone to express themselves in clear terms.

Thank you all for your attention and please excuse the verbosity.

I wish everyone the very best in their lives.

Alan-R:
I certainly don't bear anyone any ill will about those things in the distant past. When I first started the project, I decided that my criteria for success was that it should grow to the point that it didn't need me.

Although at that time I was less than happy about how things transpired, this event demonstrated that it clearly met my success criteria, albeit a bit sooner than I anticipated, and in a somewhat awkward fashion.

PR #1041, opened 2017-10-20
aws*: generate man-pages

oalbrigt:


krig:
Oops, I always forget that's a separate listing. Thanks for updating it.

PR #1042, opened 2017-10-24
Medium: Filesystem: faster stop procedure when large number of proces…

dmuhamedagic:
…ses using the filesystem are found

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -483,13 +483,19 @@ get_pids()
 signal_processes() {
 	local dir=$1
 	local sig=$2
-	local pids pid
+	local pids pid pidcnt
 	# fuser returns a non-zero return code if none of the
 	# specified files is accessed or in case of a fatal 
 	# error.
 	pids=$(get_pids "$dir")
 	if [ -z "$pids" ]; then
 		ocf_log info "No processes on $dir were signalled. force_unmount is set to '$FORCE_UNMOUNT'"
+		return
+	fi
+	pidcnt=$(echo "$pids" | wc -w)
+	if [ $pidcnt -gt 10 ]; then
+		ocf_log info "$pidcnt processes found, signaling en masse"
+		echo "$pids" | xargs kill -s $sig
 		return
 	fi
 	for pid in $pids; do


krig:
Hm, why not do the mass-kill even with fewer processes? My guess is that it's to kill the processes in a predictable order. But if that is a real concern, it seems like it could be really difficult to debug problems caused by the switch to a different method at an arbitrary limit.

dmuhamedagic:
On Tue, Oct 24, 2017 at 03:47:33AM -0700, Kristoffer Grönlund wrote:
> Hm, why not do the mass-kill even with fewer processes?

I guess because we want to log the processes which are holding
the filesystem. It is assumed that such a situation is unexpected
and such information could be useful to the administrators.
Otherwise, I don't know :)

> My guess is that it's to kill the processes in a predictable order. But if that is a real concern, it seems like it could be really difficult to debug problems caused by the switch to a different method at an arbitrary limit.

No, we don't really care about the order. It's not possible to
establish one anyway (i.e. the pid generation can wrap).

krig:
If it's just for the logging, perhaps something like this would work?

```
ps -f $pids | tail -n +2 | while read line; do
  ocf_log info "sending signal $sig to: $line"
done
```
and then just killing them all in a single command all the time.

dmuhamedagic:
On Tue, Oct 24, 2017 at 04:35:12AM -0700, Kristoffer Grönlund wrote:
> If it's just for the logging, perhaps something like this would work?
> 
> ```
> ps -f $pids | tail -n +2 | while read line; do
>   ocf_log info "sending signal $sig to: $line"
> done
> ```
> and then just killing them all in a single command all the time.

Apparently, it is ocf_log that is the bottleneck, which is
probably true. Somebody recently complained about it hence this
change.

BTW, please don't merge this yet, it's untested and I asked the
poster to do some testing.

krig:
Ah, I see.. yeah, that does make sense that it would be.

dmuhamedagic:
Did you manage to do some further testing? This PR starting to rot a bit 😉

dmuhamedagic:
The latter seems to be the best way forward. Thank you for your work and thanks to the QE for providing the code!

oalbrigt:
Sadly it doesnt seem to work with our improved stop-action logic:
https://github.com/ClusterLabs/resource-agents/pull/1878

lge:
Finally obsolete with #2111 and #2114

PR #1043, opened 2017-10-24
Feature: adding docker-native healthcheck query to docker

edge-records:
Docker 1.12 introduced native health checks. This PR adds code to optionally let this resource agent query docker for the container's health status when monitoring the container.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -155,6 +155,27 @@ will persist after the container stops.
 <content type="boolean"/>
 </parameter>
 
+<parameter name="query_docker_health" required="0" unique="0">
+<longdesc lang="en">
+Query the builtin healthcheck of docker (v1.12+) to determine health of the
+container. If left empty or set to false it will not be used.
+
+The healthcheck itself has to be configured within docker, e.g. via
+HEALTHCHECK in Dockerfile. This option just queries in what condition
+docker considers the container to be and lets ocf do its thing accordingly.
+
+Note that the time a container is in "starting" state counts against the
+monitor timeout.
+
+This is an additional check besides the standard check for the container
+to be running, and the optional monitor_cmd check. It doesn't disable or
+override them, so all of them (if used) have to come back healthy for the
+container to be considered healthy.
+</longdesc>
+<shortdesc lang="en">use healthcheck</shortdesc>
+<content type="boolean"/>
+</parameter>
+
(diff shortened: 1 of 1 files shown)


krig:
Looks good to me.

dmuhamedagic:
On Tue, Oct 24, 2017 at 01:35:20PM +0000, Kristoffer Grönlund wrote:
> Looks good to me.

If it's only available with v1.12 and later then shouldn't that
be checked too?

krig:
I guess it might be possible to check if the feature is supported in `validate-all` and fail with a clear error message in case the option is enabled despite the docker version being <1.1.12.

edge-records:
> I guess it might be possible
Well I personally think that's over the top. Also it's not trivial in bash to compare version strings, particularly considering the version format change of docker from `Docker 1.13.0` to `Docker 17.03.0-ce` (see https://github.com/moby/moby/wiki). Also I don't have a test suite of all the old docker versions to determine whether the command I'm using to query the version or its output format hasn't changed between versions.

krig:
I agree, version checks are error-prone and it's probably better to avoid them. Unless there is some way to test for the feature directly, I'd agree with not doing that kind of validation and rely on the users ability to read the documentation for the feature before enabling it.

dmuhamedagic:
On Wed, Oct 25, 2017 at 09:27:34AM +0000, Kristoffer Grönlund wrote:
> I agree, version checks are error-prone and it's probably
> better to avoid them. Unless there is some way to test for the
> feature directly, I'd agree with not doing that kind of
> validation and rely on the users ability to read the
> documentation for the feature before enabling it.

We already do version comparison in several RAs and it usually
works. I don't know whether it is going to work for docker. I
still think it's better than _nothing_. Of course, as you
mentioned, another kind of validation is vastly preferable. At
any rate, in the case when the feature must be turned on by
setting a parameter, we can go either way :)

oalbrigt:
I guess we could use version comparison in addition to a parameter (e.g. auto default where it uses the comparison), and the parameter could be used to set it on if it's backported to earlier versions and you want to use it.

krig:
If it is OK with everyone, I'd merge this as it's an optional feature that has to be manually enabled. If we figure out a way to test for feature support that isn't looking at the version, we could add validation based on that later.

PR #1044, opened 2017-10-30
Dev: LVM: put VG protection mechanisms into separate files

dmuhamedagic:
Refactor LVM in order to reduce complexity.

clvm and tag specific stuff is moved into its own files which are
sourced depending on the resource configuration.

### DIFF
--- a/heartbeat/LVM
+++ b/heartbeat/LVM
@@ -51,9 +51,6 @@ usage() {
 EOF
 }
 
-# default for "tag"
-OUR_TAG="pacemaker"
-
 meta_data() {
 	cat <<EOF
 <?xml version="1.0"?>
@@ -93,7 +90,7 @@ can be as simple as setting 'volume_list = []' depending on your setup.
 If "exclusive" is set on a non clustered volume group, this overrides the tag to be used.
 </longdesc>
 <shortdesc lang="en">Exclusive activation tag</shortdesc>
-<content type="string" default="$OUR_TAG" />
+<content type="string" default="pacemaker" />
 </parameter>
 
 <parameter name="partial_activation" unique="0" required="0">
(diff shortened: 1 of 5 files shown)


dmuhamedagic:
The refactored RA passed the ocft tests. However, those tests don't exercise clvm or tag!

renzhengeek:
Hi Dejan,

I don't know how to react to your change. I think it's a good attempt to make LVM easy to add new feature support, but for now I still see a real challenge to reach the goal:

What if the new feature like lvmlockd want to add more parameters?

It's hard for me to take a preference right now. Let's see what the maintainer will comment on this. Looks I have to wait until the decision is made. 

BTW, there's an obvious logical error in your change below:

```
	if ocf_is_true "$OCF_RESKEY_exclusive"; then
		case $(vgs -o attr --noheadings $OCF_RESKEY_volgrpname | tr -d ' ') in
		?????c*)
			mode="clvm" ;;
		*)
			mode="tag" ;;
		esac
	else
		mode="plain"
fi
```

1. I don't think we should support "plain" (pure local activation) mode in RA which is designed for HA cluster.
2. "clvm" can support both "cluster-wide" (default) and "exclusive" activation, why it's under the condition of "exclusive" parameter?

dmuhamedagic:
On Wed, Nov 01, 2017 at 09:48:24AM +0000, Eric Ren wrote:
> What if the new feature like lvmlockd want to add more parameters?

Just add more parameters. It happens every now and then that new
parameters get introduced, in particular in concert with new
features.

> BTW, there's an obvious logical error in your change below:

I should stress again that this is _only_ refactoring and, by
definition, apart from code reorganization, there are no other
changes whatsoever. I even refrained from changing white space or
fixing typos. Hence, any existing issues in LVM are present here
too.

> 1. I don't think we should support "plain" (pure local activation) mode in RA which is designed for HA cluster.

We already discussed this a few days ago in the other pr. Apart
from what was said there, I can recall reports that people do
intentionally activate volume groups on more than one node in
order to use the data in the read only mode.

> 2. "clvm" can support both "cluster-wide" (default) and "exclusive" activation, why it's under the condition of "exclusive" parameter?

I really cannot say. That was how it was implemented.

This change separates the LVM code from clvmd and tags features.
Its purpose is to expose the LVM part, so that we can have a
better view of the possible pitfalls, differences, etc.

krig:
Merging this refactoring of the LVM agent. Sharing the various activation implementations with the LVM-activate agent should at least in theory be possible as a next step.

renzhengeek:
> Sharing the various activation implementations with the LVM-activate agent should at least in theory be possible as a next step.

OK, since Dejan has split the old functionality into files, I will see how to re-use the code in the new RA instead of keeping replicated code in the same source :)

PR #1047, opened 2017-11-03
Low: LVM-activate: Move notes from comment to long description

krig:
Edit and move comments regarding usage from the file header into the long description, so that they are visible in user tools.


krig:
I also removed some comments regarding the LVM agent. ;)

krig:
The CI failure seems to be completely unrelated (apt-get fails due to a broken package in the Ubuntu repositories), and these are just edits to comments.. so I'll merge despite the failure.

PR #1051, opened 2017-11-08
High: VirtualDomain: Rename utilization parameters and reverse default

krig:
Having the VirtualDomain agent set utilization limits by default
is confusing and (in my opinion) not the correct choice. Someone
setting the agent up for the first time would not look at the
autoset parameters and consider utilization an advanced feature,
and the resource wouldn't be able to start unless they had also
set matching utilization limits on the nodes, or used the
NodeUtilization agent.

One option would be to just change the default value. It seemed
more appropriate to rename the parameters, so that the change
becomes more visible to anyone upgrading.


gao-yan:
I think this change makes sense, or at least we should use the default placement-strategy instead of "balanced" in crmsh bootstrap.

oalbrigt:
add to whitelist

PR #1052, opened 2017-11-08
galera: recover from empty gvwstate.dat

dciabrin:
While running, a galera node keeps track of the last known state of
the cluster in a temporary file gvwstate.dat. This file is normally
deleted once a node is shutdown gracefully.

Some ungraceful shutdowns can leave an empty gvwstate.dat on
disk. This will prevent galera to join the cluster if it is
configured to attempt PC recovery. Removing that file makes the
node fall back to the normal, unoptimized joining process next
time it is restarted.


oalbrigt:
LGTM. Thanks.

PR #1053, opened 2017-11-09
Low: pgsql: improve validations for replication mode

ytakeshita:
* Fix incorrect judgment when include line is commented out in postgresql.conf
* Detects incorrect setting of synchronous_standby_names


kskmori:
Looks good to me. Merging.

PR #1058, opened 2017-11-17
Add resource agent for IPSEC tunnels

JAORMX:
This adds a simple resource agent that handles ipsec tunnels.

It will associate a given tunnel witha specific VIP, and will put it up
or down depending on the VIP being there or not. This is meant to have a
collocation rule with an IP resource.

### DIFF
--- a/heartbeat/ipsec
+++ b/heartbeat/ipsec
@@ -1,0 +1,172 @@
+#!/bin/sh
+#
+#
+#	IPSEC OCF RA. Handles IPSEC tunnels associated with a VIP
+#
+# Copyright (c) 2017 Red Hat Inc.
+#                    All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.  Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+
+#######################################################################
+# Initialization:
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+#######################################################################
+
+# Defaults
+
+OCF_RESKEY_confdir_default="/etc/ipsec.d/"
+: ${OCF_RESKEY_confdir=${OCF_RESKEY_confdir_default}}
+
+meta_data() {
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="ipsec">
+<version>1.0</version>
+
+<longdesc lang="en">
+This is a Resource Agent to manage IPSEC tunnels associated with a VIP.
+It's meant to be collocated with a specific VIP, and will manage
+setting up or down a specific tunnel.
+</longdesc>
+<shortdesc lang="en">Handles IPSEC tunnels for VIPs</shortdesc>
+
+<parameters>
+<parameter name="tunnel" unique="1" required="1">
+<longdesc lang="en">
+The name of the tunnel to be monitored.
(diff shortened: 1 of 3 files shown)


heartbeat/IPSEC
```diff
@@ -0,0 +1,163 @@
...
+	state=$?
+	if [ "$state" == "0" ]; then
+		return $OCF_SUCCESS
+	else
+		return $OCF_ERR_GENERIC
+	fi
+}
+
+: ${OCF_RESKEY_tunnel=${OCF_RESKEY_tunnel}}
+: ${OCF_RESKEY_vip=${OCF_RESKEY_vip}}
```
oalbrigt: You can skip these when there's no default.

heartbeat/IPSEC
```diff
@@ -0,0 +1,163 @@
...
+		fi
+	else
+		return $OCF_SUCCESS
+	fi
+}
+
+ipsec_validate() {
+	# The tunnel needs to be defined in the configuration
+	cat /etc/ipsec.d/*.conf | grep -q "conn $OCF_RESKEY_tunnel"
+	state=$?
```
oalbrigt: Maybe it'd be better to use a conf-parameter to specify the .conf-file and do 
"if [ -f ... ]" instead, and maybe a confdir parameter that defaults to "/etc/ipsec.d/" that could later be part of e.g. configure auto-detection based on distribution/OS.

oalbrigt:
Can you rename it to lowercase? As we prefer agents being named in lowercase.

You should also add it to Makefile.am in the heartbeat and doc/man directories.

heartbeat/IPSEC
JAORMX: I could do "/etc/ipsec.d/" in a confdir parameter. But do we really need to specify the parameter? I dynamically generate the names, where I use this in my ansible role, I dynamically generate the configurations (and their names), so I don't fully know which configuration contains which tunnel. I could just stick them all in one file, but then it gets harder to debug later. This just makes life way simpler :/

heartbeat/IPSEC
JAORMX: sure :)

heartbeat/IPSEC
oalbrigt: Ah. In that case I guess this is OK, so the confdir parameter would be nice to support all distros/OSs and other scenarios.

heartbeat/ipsec
```diff
@@ -0,0 +1,169 @@
...
+	cat ${OCF_RESKEY_confdir}/*.conf | grep -q "conn $OCF_RESKEY_tunnel"
+	state=$?
+	if [ "$state" == "0" ]; then
+		return $OCF_SUCCESS
+	else
+		return $OCF_ERR_GENERIC
+	fi
+}
+
+: ${OCF_RESKEY_confdir=${OCF_RESKEY_confdir}}
```
oalbrigt: This needs to be in the top and use pre-defined _default variable (which makes it set it if it isnt already set).

Example:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/awseip#L42-L49

You can then use the _default variable in the metadata as well instead of hard-coding it.

oalbrigt:
You need to chmod 755 the file, so it can be run during automatic man-page generation.

heartbeat/ipsec
```diff
@@ -0,0 +1,170 @@
...
+	state=$?
+	if [ "$state" == "0" ]; then
+		return $OCF_SUCCESS
+	else
+		return $OCF_ERR_GENERIC
+	fi
+}
+
+OCF_RESKEY_confdir_default="/etc/ipsec.d/"
+: ${OCF_RESKEY_confdir=${OCF_RESKEY_confdir_default}}
```
oalbrigt: Great. If you move this before metadata you can set default="${OCF_RESKEY_confdir_default}" in the parameter section.

JAORMX: it's already before meta-data. Or do you means something else?

oalbrigt: Oh. Before the meta_data() function.

heartbeat/ipsec
```diff
@@ -0,0 +1,172 @@
...
+</longdesc>
+<shortdesc lang="en">VIP</shortdesc>
+<content type="string" default="" />
+</parameter>
+<parameter name="confdir">
+<longdesc lang="en">
+The directory where the IPSEC tunnel configurations can be found.
+</longdesc>
+<shortdesc lang="en">Tunnel name</shortdesc>
+<content type="string" default="/etc/ipsec.d/" />
```
oalbrigt: You can change default here to default="${OCF_RESKEY_confdir_default}" so it's only set once in the agent.

oalbrigt:
LGTM.

@krig, @dmuhamedagic: You got anything to add? If not you're welcome to merge this.

JAORMX:
@oalbrigt thanks a lot for the reviews!

krig:
My only comment would be that the abbreviation VIP is never clarified anywhere in the metadata, which although not a problem just seems like it ought to be. Maybe in the long description for the `vip` parameter it could be spelled out as virtual IP address just for clarity?

Other than that it looks clean and simple, good to merge from my perspective!

JAORMX:
@krig thanks for the feedback. Done.

krig:
Great, thank you!

PR #1060, opened 2017-11-22
db2: fix HADR promote when master failed

oalbrigt:
... also fixed OCF_GENERIC_ERR (non-existing return code).


krig:
Looks good

PR #1064, opened 2017-11-30
High: jira: new resource agent

salehabbas:
This is a new ocf ra for managing JIRA software.

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -120,6 +120,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_ids.7 \
                           ocf_heartbeat_iscsi.7 \
                           ocf_heartbeat_jboss.7 \
+			  ocf_heartbeat_jira.7 \
                           ocf_heartbeat_kamailio.7 \
                           ocf_heartbeat_lvmlockd.7 \
                           ocf_heartbeat_lxc.7 \
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -120,7 +120,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_ids.7 \
                           ocf_heartbeat_iscsi.7 \
                           ocf_heartbeat_jboss.7 \
-			  ocf_heartbeat_jira.7 \
+                          ocf_heartbeat_jira.7 \
                           ocf_heartbeat_kamailio.7 \
                           ocf_heartbeat_lvmlockd.7 \
                           ocf_heartbeat_lxc.7 \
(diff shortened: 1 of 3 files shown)


doc/man/Makefile.am
```diff
@@ -120,6 +120,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_ids.7 \
                           ocf_heartbeat_iscsi.7 \
                           ocf_heartbeat_jboss.7 \
+			  ocf_heartbeat_jira.7 \
```
krig: Should be spaces, not tabs here

heartbeat/jira
```diff
@@ -0,0 +1,266 @@
...
+
+    return $OCF_SUCCESS
+}
+
+# Meta-data
+jira_meta_data(){
+    cat <<EOF
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="jira" version="0.1">
```
krig: There shouldn't be a version attribute here. This is why the CI check fails, so if you remove the version here I think it should pass.

heartbeat/jira
```diff
@@ -0,0 +1,266 @@
+#! /bin/bash
+# 
+####################################################################
+# Description: OCF Resource Agent to manage JIRA software.
+# Author     : Saleh A. (saleh.abbas.saber@gmail.com)
+#
+# License    : WTFPL 2
+#
+#        DO WHAT THE FUCK YOU WANT TO PUBLIC LICENSE
```
krig: Nice choice of license, haha :)

doc/man/Makefile.am
salehabbas: Fixed, thanks!

heartbeat/jira
salehabbas: :)

heartbeat/jira
salehabbas: Fixed, thanks!

heartbeat/jira
```diff
@@ -0,0 +1,281 @@
...
+
+    # if resource is already running, bail out early
+    if jira_monitor; then
+        ocf_log info "Resource is already running"
+        return $OCF_SUCCESS
+    fi
+
+    # Starting Jira
+    waittime=300
+    /bin/su -m $jira_user -c "$jira_installation/bin/startup.sh &> /dev/null"
```
oalbrigt: There's no reason to use full path for su, and it might cause issues for some users.

salehabbas: Removed. Thanks

oalbrigt:
LGTM.

krig:
I just noticed that the agent uses `curl`, this should probably be noted in the description so that a user knows what the dependencies for using the agent are.

But since it's just a minor detail I'll merge this as it is now, the agent itself looks good to me too.

PR #1066, opened 2017-12-05
Fix fallback name for ocf_attribute_target

dciabrin:
For bundles, various resource agents now use ocf_attribute_target to
get the name of the pacemaker node to store attributes on.

If a recent version of the resource agent is being run on a pacemaker
version which does not support bundles, ocf_attribute_target will
return an empty string as hostname.

Provide a fallback path so the resource agent gets a valid name when
the resource is not containerized.


dciabrin:
Please do not merge it right away, I'd like to discuss that so we all agree on when that can happen and whether that's the proper fix or not.

dciabrin:
OK, I've tested this PR by creating a galera resource with the latest resource-agents, on a cluster managed by an old pacemaker from RHEL 7.2.
Without this fix, the resource won't come up due to the agent not setting the attributes properly in the CIB.
With this fix, the resource goes to master state as expected.

PR #1071, opened 2017-12-12
CTDB: Cope with deprecated "idmap backend" smb.conf option

pablomh:
Keep it only if we find it in the original smb.conf file and change the default
to the non-deprecated "idmap config * : backend =" format.

### DIFF
--- a/heartbeat/CTDB
+++ b/heartbeat/CTDB
@@ -440,7 +440,7 @@ init_smb_conf() {
 
 	# replace these things in smb.conf
 	local repl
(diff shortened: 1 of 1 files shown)


pablomh:
If not, Samba logs get spammed with:

```
[2017/12/12 09:13:06.259809,  1] ../lib/param/loadparm.c:1770(lpcfg_do_global_parameter)
  WARNING: The "idmap backend" option is deprecated
```

krig:
Looks good, thanks

PR #1072, opened 2017-12-18
targetcli 3.x support

patrickdk77:
iSCSILogicalUnit fixups for lio-t to support targetcli 3.x, and remove target adjustments
iSCSITarget add the ipv6 support that was removed from iSCSILogicalUnit, with conditional test


krig:
At first glance, these changes don't look backwards compatible to me, is that correct? Unfortunately, I don't know iSCSI well enough to be comfortable reviewing this.

fpiecka:
I believe the changes are compatible with targetcli 2 which is included in the current Linux distributions.

Additionaly, I believe this PR resolves #1256.

@patrickdk77, would you mind resolving the conflicts? I can do it if you don't...

PR #1075, opened 2017-12-23
Container-monitoring resources

mgrzybek:
This PR adds two RA :
* LXDInfo : retrieves the number of running LXD containers  using `lxc list` and updates the CRM attributes
* MachineInfo : retrieves the number of running systemd machines using `machinectl` and updates the CRM attributes


krig:
`LXDInfo` doesn't have the executable bit set, that's why the CI check fails.

krig:
Ah, right: Please add the new agents to `doc/man/Makefile.am` as well, to generate the man pages.

krig:
Great, thanks!

krig:
Oh, I forgot about the naming policy we're trying to enforce. Would it be OK with you if we rename these agents `lxd-info` and `machine-info`?

mgrzybek:
That's all right.
Is there any document related to the naming convention?

krig:
There isn't yet. The convention-to-be is "lowercase with optional -'s".

PR #1080, opened 2018-01-04
Rename LXDInfo to lxd-info and MachineInfo to machine-info

krig:
@oalbrigt Hm, noticed that something isn't quite working with the CI: The agent name was typoed as LXCInfo in `doc/man/Makefile.am` but the CI check still passed. Will need to investigate.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ addons:
     #  - debian-sid
     packages:
       - libxml2-utils
+      - xsltproc
       - shellcheck
 script:
   - ./ci/build.sh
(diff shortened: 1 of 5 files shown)


oalbrigt:
Sounds like it doesnt catch errors during man-page creation for some reason.

krig:
Yeah, turns out it doesn't even do the manpage build on Travis.

```
checking for xsltproc... no
configure: WARNING: xsltproc not installed, unable to (re-)build manual pages
```

oalbrigt:
Great. It's now producing man-pages in the last CI-build.

krig:
Alright, I'll merge this :)

I noticed that the manpage generation for `redis` is broken though, so the error discovery isn't quite right. Will try to fix that in a separate PR.

PR #1085, opened 2018-01-08
Fixes for lvmlockd and LVM-activate RAs

renzhengeek:


krig:
Hmm, I'm not sure if I think automatically changing the configuration from the agent is a good idea. @oalbrigt, @dmuhamedagic, any opinion?

oalbrigt:
It might be a good idea to have a note in the metadata that it changes these settings.

renzhengeek:
> Hmm, I'm not sure if I think automatically changing the configuration from the agent is a good idea. @oalbrigt, @dmuhamedagic, any opinion?

In the clvm RA, it use lvmconf command to change the /etc/lvm/lvm.conf.

In Aug. 2017, I tried to push a patch for lvmconf to support enable/disable settings for lvmlock:

https://www.redhat.com/archives/lvm-devel/2017-August/msg00030.html

Unfortunately, LVM upstream seems not interested in that. We cannot just carry that patch in SUSE LVM package, if so, the resource agent wound not work for distro that has not this patch.

Anyway, lvmconf is just a script that use sed to change the lvm.conf directly. So, I think changing the lvm.conf directly in this RA doesn't make much difference with using lvmconf.

renzhengeek:
> It might be a good idea to have a note in the metadata that it changes these settings.

Good idea. I've updated as you suggest. Thanks.

krig:
> Unfortunately, LVM upstream seems not interested in that. 

Looking at the email, I don't see any explanation *why* you are adding the feature to enable/disable lvmlockd. Generally, it is difficult to argue for or against a chance without knowing the reason behind it. This might be why there was little response from upstream.

I would recommend submitting the patch again, and explaining why you want to make these changes. Then, upstream also has the option of describing an alternative way of achieving what you want. :)

renzhengeek:
> Looking at the email, I don't see any explanation why you are adding the feature to enable/disable lvmlockd. 

Hah, I think I had some bit explanation:

```
lvmconf is handy when you need to change lvm.conf in script,
like resource agent in HA cluster. This patch add support
for enabling/disabling lvmlockd
```
Maybe the wording is not good enough, but the reason is quite straightforward - allowing to change settings for lvmlockd with lvmconf so that it can be used in script like HA resource agent.

Well, let me try again. Thanks.

krig:
OK, I'll merge with the note in the meta data. Thanks!

PR #1086, opened 2018-01-10
Low: Remove "s" suffix from action default times

nrwahl2:
This patch introduces consistency by removing the trailing "s" character
from default time values for actions in resource agent metadata.
The default unit for time is seconds. Currently, some instances of time
values include the explicit "s" for the seconds unit, while others do
not.

### DIFF
--- a/heartbeat/IPaddr
+++ b/heartbeat/IPaddr
@@ -195,11 +195,11 @@ netmask for ARP - in nonstandard hexadecimal format.
 </parameters>
 
 <actions>
-<action name="start"   timeout="20s" />
-<action name="stop"    timeout="20s" />
-<action name="monitor" depth="0"  timeout="20s" interval="5s" />
-<action name="validate-all"  timeout="20s" />
-<action name="meta-data"  timeout="5s" />
+<action name="start"   timeout="20" />
+<action name="stop"    timeout="20" />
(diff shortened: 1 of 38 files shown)


nrwahl2:
If we implement this, we may also want to add a node to the dev guide so that we don't drift away from this again. Open to guidance on where to put it (e.g., Metadata or Conventions). The code examples in the dev guide already use this "s"-less form.

krig:
Makes sense to me.

marxsk:
I would prefer to go in the opposite direction. There is already support for even more complex XmYs (min:sec) in pcs/pacemaker for some cases. So making this a default might have sense. From the UI perspective, I would like to have different type 'time' so I can prepare a proper field.

[*] type='time' should be part of OCF this year. But we have to fix other issues before adding new ones.

nrwahl2:
I'm open to either format, as long as it's consistent. Some of us favor the bare numbers because of its arguably cleaner look and because of the prevalence of specifying timeouts/intervals in seconds.

oalbrigt:
I think it makes sense to add 's' where missing instead to unify with how it's done in fence-agents.

nrwahl2:
Created #1096 with the opposite approach (adding the "s" where it's missing). I'll leave both open for people to decide.

krig:
Closing since #1096 was merged instead. Thanks!

PR #1088, opened 2018-01-11
syslog-ng: add support for commercial syslog-ng PE

abalage:
Balabit's commercial variant of syslog-ng, a.k.a Premium Edition installs itself into /opt and uses a libexec wrapper to start the actual service, with original path assumption this will not work.
The changes I made strips off the path from the executable (basename). This way it works both with Syslog-NG OSE and PE.

ps. I work for Balabit


oalbrigt:
You can just set the syslog_ng_binary parameter to the location instead of this patch.

abalage:
I know, but that always means a local change on each system which might be overwritten by any pacemaker package upgrades in the future.

oalbrigt:
No, you set the parameter (for the resource not in the agent), and it'll keep on using it after upgrades.

abalage:
Oh I remember. I tried that. The problem with that is the resource executable which needs to be called is a wrapper script doing environment initialization and doing an exec at the end of the real binary whose path will be different, thus the check logic will not find it within the running processes.

ie. /opt/syslog-ng/sbin/syslog-ng is a symbolic link to /opt/syslog-ng/libexec/syslog-ng-wrapper.sh and the final executable will be /opt/syslog-ng/libexec/syslog-ng and the arguments of course.

What I did was just simply omitting the basedir and leaving syslog-ng <ARGS>.

oalbrigt:
Oh. That makes sense.

PR #1092, opened 2018-01-17
Drop python from build system since nothing uses it

krig:
There's no need to require python when it isn't actually used.


oalbrigt:
Nice catch.

PR #1093, opened 2018-01-23
Low: tomcat: Fix invalid stop option

ytakeshita:
"--force" is invalid option for stop. it should be "-force".


krig:
Thank you!

PR #1095, opened 2018-01-30
Fix raid1 stop detection

Werkov:
The following sequence of events was observed in the wild:
```
16:19:04.379377+01:00 Raid1(rsc_Raid1_md7)[25734]: INFO: Stopping array /dev/md7
16:19:04.389490+01:00 kernel: [12955.559495] md7: detected capacity change from 429496664064 to 0
16:19:04.389499+01:00 kernel: [12955.559513] md: md7 stopped.
16:19:04.511817+01:00 Raid1(rsc_Raid1_md7)[25734]: ERROR: mdadm failed on /dev/md7.
16:19:04.520194+01:00 Raid1(rsc_Raid1_md7)[25734]: ERROR: RAID /dev/md7 still active after stop command!
```
As pointed by @neilbrown, the `-W` and `--wait-clean` flags are ignored (when they come *after* device) and useless as `mdadm --stop` only returns when the device is stopped.
The false positive is a transient device that may be created for short periods when the removal uevent is being processed by other programs.

### DIFF
--- a/heartbeat/Raid1
+++ b/heartbeat/Raid1
@@ -384,8 +384,16 @@ raid1_monitor_one() {
 		2)	ocf_exit_reason "$mddev has failed."
 			return $OCF_ERR_GENERIC
 			;;
-		4)	ocf_exit_reason "mdadm failed on $mddev."
-			return $OCF_ERR_GENERIC
+		4)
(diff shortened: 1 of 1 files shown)


krig:
LGTM, thank you.

PR #1096, opened 2018-01-31
Low: Add "s" suffix to action default times

nrwahl2:
This patch introduces consistency by adding an explicit trailing "s"
character to default time values for actions in resource agent metadata.
The default unit for time is seconds, but other units are supported.
Currently, some instances of time values include the explicit "s", while
others do not.

Adding the "s" will also be more consistent with the approach in
fence agents.

### DIFF
--- a/heartbeat/AoEtarget
+++ b/heartbeat/AoEtarget
@@ -101,12 +101,12 @@ Location of the vblade binary.
     </parameter>
   </parameters>
   <actions>
-    <action name="start" timeout="15"/>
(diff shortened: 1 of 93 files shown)


nrwahl2:
See also #1086 with the opposite approach (removing the "s" where it's present). I'll leave both open for people to decide.

oalbrigt:
LGTM.

krig:
I think I would prefer this variant over #1086 as well.

Issue #1100, opened 2018-02-08
Filesystem heartbeat agent can't mount StorNext filesystem


davidmhewitt:
With a resource configured as follows:
```
# pcs resource show nbs_t1  
Resource: nbs_t1 (class=ocf provider=heartbeat type=Filesystem)   
Attributes: device=nbs_t1 directory=/stornext/nbs_t1 fstype=cvfs options=rw,diskproxy=client force_clones=true
```

The following error occurs:
```
# pcs resource debug-start nbs_t1 
Error performing operation: Input/output error Operation start for nbs_t1 (ocf:heartbeat:Filesystem) returned 5  >  
stderr: INFO: Running start for nbs_t1 on /stornext/nbs_t1  >  
stderr: ERROR: Couldn't find device [nbs_t1]. Expected /dev/??? to exist 
```

This is due to the way cvfs labels file systems and doesn't use block devices. The reference manual can be found here:
http://qsupport.quantum.com/kb/flare/Content/stornext/SN5_DocSite/Guide_Installation/Topics/Configuring_StorNext_RPM.htm

Here is a working line from our `/etc/fstab`:
`nbs_t1 /stornext/nbs_t1 cvfs rw,relatime,diskproxy=client,sparse=yes 0 0`

PR #1101, opened 2018-02-08
Add support for cvfs

davidmhewitt:
Fixes https://github.com/ClusterLabs/resource-agents/issues/1100


krig:
Seems straightforward enough to me, thanks!

PR #1102, opened 2018-02-09
pgsql: try to create stats_temp_directory

vvidic:
Database fails to start correctly if the stats_temp_directory
is specified in the configuration but does not exists.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -575,6 +575,8 @@ pgsql_real_start() {
         check_socket_dir
     fi
 
+    check_stat_temp_directory
+
     if [ "$OCF_RESKEY_rep_mode" = "slave" ]; then
         rm -f $RECOVERY_CONF
         make_recovery_conf || return $OCF_ERR_GENERIC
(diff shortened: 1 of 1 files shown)


heartbeat/pgsql
```diff
@@ -1977,6 +1979,43 @@ check_log_file() {
...
+        ocf_exit_reason "Can't create directory $stats_temp"
+        exit $OCF_ERR_PERM
+    fi
+
+    if ! chown $OCF_RESKEY_pgdba: "$stats_temp"; then
+        ocf_exit_reason "Can't change ownership for $stats_temp"
+        exit $OCF_ERR_PERM
+    fi
+
+    if ! chmod 750 "$stats_temp"; then
```
kskmori: Is there a reason that the permission is 750? All other directories in PGDATA have 700 permission by defaults so 700 seems like more appropriate for me.

kskmori:
Looks good to me for the rest. I will merge this PR after we make sure about the permissions.
Thank you for your contributions!

heartbeat/pgsql
vvidic: 750 is the perms used by the Debian startup scripts when creating this directory but it seems to work with 700 too so I'll update the patch now.

kskmori: I see. I've only checked on RHEL/CentOS so it seems distributions dependent.
Thank you for making sure that it always works with 700 and updating the patch!

PR #1103, opened 2018-02-09
pgsql: fix possible bashism

vvidic:
Quote grep regexp to prevent possible shell expansion and
fix bashism warning:

    possible bashism in pgsql line 1949 ([^] should be [!]):
      echo "$OCF_RESKEY_replication_slot_name" | grep -q -e [^a-z0-9_]


oalbrigt:
Thanks.

Issue #1087, opened 2018-01-10
VirtualDomain VMs aren't migrated during node shutdown.


Werkov:
### Steps to reproduce

* configure a two node cluster
* configure a VirtualDomain resource active on one node
* reboot the node hosting VM

### Expected behavior

The resource will be migrated to the other node using `migrate_from` and `migrate_to` RA actions.

### Actual behavior

The RA cannot query VM state and the `migrate_from` operation times out. (Alternatively, VM is shut off by systemd shutdown transaction before VirtualDomain RA can act on it.)

Technical details and proposal is in the referenced patch.

Werkov:
Possibly, `VirtualDomain` could call `systemctl start resource-agents-dep.target` to enforce state defined by the dependencies. (Transitively triggered `virt-guest-shutdown.target` is just an [ordering target](https://libvirt.org/git/?p=libvirt.git;a=blob;f=src/util/virsystemd.c;h=3be963d704ea7283ade55f873938ed2b17b4111e;hb=HEAD#l347), it doesn't activate any units.)

I'm not sure if this is the best solution there can be though.

Werkov:
FTR, this also needs that `libvirtd.service` specifies `After=systemd.machined`. (I'll add link to libvirtd ML thread once my initial e-mail goes through moderation.)

Werkov:
I pushed a modified version f65635c as suggested in [previous comment](https://github.com/ClusterLabs/resource-agents/issues/1087#issuecomment-356762222). It seems to work in my reproducer.

krig:
@oalbrigt Any opinion on this one? I'm not sure how portable it is.

oalbrigt:
Those systemd services are available on RHEL and Debian stable.

Werkov:
Thanks for the portability remark. I see that virt-guest-shutdown.target was only introduced in [libvirt v2.5.0-rc1~65](https://libvirt.org/git/?p=libvirt.git;a=commit;h=01079727fe29dfeafac751a07e564e0d6bb53389). I see that resource-agents.spec.in doesn't mention dependency on libvirt (as not all resource-agents users are probably using VirtualDomain RA). I guess it's alright if upstream resource-agents depend on upstream libvirt (or >= v2.5.0-rc1~65) to be more precise.
I've filed PR #1105 with (rebased) patch. If there were any other comments regarding the change, please let me know.

PR #1105, opened 2018-02-14
VirtualDomain: Properly migrate VMs on node shutdown (bsc#1074014)

Werkov:
There are three actors in sum controlling the VMs:
- pacemaker/VirtualDomain RA
  - issuing cluster resource operations,
- libvirtd
  - executing the given commands,
- systemd and systemd-machined
  - manages scope units of running VMs.

During shutdown both libvirtd and systemd-machined termination runs
concurrently with pacemaker so it cannot properly communicate with
libvirtd and terminate the machines (e.g. stopped scope unit would turn
the VM off instead of migration that's usually intended in a HA
cluster).

We add ordering dependency both on libvirtd.service and
virt-guest-shutdown.target so that pacemaker is terminated *before* the
other mechanisms would stop the VMs. Thus it can invoke appropriate
resource action on the running VM.

In order the ordering against virt-guest-shutdown.target to work its
stop job must be in the shutdown transaction thus we start the target
when activating a VM. The target is ordering-only, it doesn't activate
any other units.

Ref: #1087

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -518,6 +518,14 @@ VirtualDomain_start() {
 		return $OCF_SUCCESS
 	fi
 
+	# systemd drop-in to stop domain before libvirtd terminates services
+	# during shutdown/reboot
+	if systemd_is_running ; then
+		systemd_drop_in "99-VirtualDomain-libvirt" "After" "libvirtd.service"
+		systemd_drop_in "99-VirtualDomain-machines" "Wants" "virt-guest-shutdown.target"
+		systemctl start virt-guest-shutdown.target
+	fi
+
 	snapshotimage="$OCF_RESKEY_snapshot/${DOMAIN_NAME}.state"
 	if [ -n "$OCF_RESKEY_snapshot" -a -f "$snapshotimage" ]; then
 		virsh restore $snapshotimage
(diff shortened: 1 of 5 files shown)


heartbeat/VirtualDomain
```diff
@@ -518,6 +518,14 @@ VirtualDomain_start() {
 		return $OCF_SUCCESS
 	fi
 
+	# systemd drop-in to stop domain before libvirtd terminates services
+	# during shutdown/reboot
+	if ps -p 1 | grep -q systemd ; then
```
krig: I think a better way to detect systemd is to use

`[ "$(cat /proc/1/comm 2>/dev/null)" = "systemd" ]`

Werkov: I think that @beekhof once pointed it should be extracted to a function. I'll refactor the branch accordingly, then the actual detection can be easily modified.

Werkov:
I (force) pushed the updated branch. It's function should be identical, only systemd detection was changed following the comments.

oalbrigt:
I think we should call the function systemd_running or something similar, as that's what it's checking (systemd can be present and not running).

krig:
LGTM, @oalbrigt ?

oalbrigt:
LGTM too.

PR #1109, opened 2018-02-21
IPv6: rewrite :: to :0: for single-field 0 mistakes

oalbrigt:


kskmori:
Just a question... May I ask the reason why the rewriting is necessary? Any misbehavior without this? or only meant for complying with rfc5952 4.2.2?

oalbrigt:
Because ip accepts it and then shows it with the :0: when used later, which makes it fail.

kskmori:
Ah, now I see your point. Yes, that is definitly an issue. Thank you for the explanation!

But it turns out that there are more other cases that /sbin/ip will rewrite the address which would result the same failures. At least RFC5952 defines the following rules for sanitizing the IPv6 address representation.

* `2001:db8::0001` -> `2001:db8::1` (section 4.1)
* `2001:db8:0:0:0:0:2:1` -> `2001:db8::2:1` (section 4.2.1)
* `2001:db8::1:1:1:1:1` -> `2001:db8:0:1:1:1:1:1` (section 4.2.2, this is the case of this PR)
* `2001:db8:0:0:1::1` -> `2001:db8::1:0:0:1` (section 4.2.3)
* `2001:DB8::1` -> `2001:db8::1` (section 4.3)

It might be great if we could ask /sbin/ip only how it would rewrite the address and obtain the result. `/sbin/ip route get` might be usable but I have no confident if it always works reliably, and has no bad impacts (for the load, etc.). Hopefully there is a better way.
```
# /sbin/ip route get 2001:db8::0001 | awk '$1~/:/ {print $1}  $2~/:/ {print $2}'
2001:db8::1
# 
```

If you want to fix at least this particular case ASAP, then we can merge this PR for the moment, but probably we still need to ask users to specify the IPv6 address in the sanitized representation.

heartbeat/IPaddr2
```diff
@@ -101,6 +101,9 @@ VLDIR=$HA_RSCTMP
 SENDARPPIDDIR=$HA_RSCTMP
 CIP_lockfile=$HA_RSCTMP/IPaddr2-CIP-${OCF_RESKEY_ip}
 
+# IPv6: rewrite :: to :0: for single-field 0 mistakes
+OCF_RESKEY_ip=`echo $OCF_RESKEY_ip | awk -F: -vOFS=: '/::/ && NF==8 {for (i=1;i<=NF;i++) {$i=$i?$i:0}} {print}'`
```
kskmori: I would prefer to move this to somewhere close to the parameter checking in ip_init(), and also rewrite only when inet6 to minimize the impacts for the existing IPv4 configration.

oalbrigt:
I'm in no hurry to merge this, so I'll improve the patch as suggested (where it fails and tells the user to use correct representation based on how "ip" presents the address).

oalbrigt:
@kskmori can you create a PR for your fix?

kskmori:
@oalbrigt Sure. Do you agree on using `ip route get` for rewriting the address? It should just work but I have no confident if it is really flawless for every network configurations. Also I would need to think an error handling here.

oalbrigt:
@kskmori Yeah. Seems like that's the only reliable way without doing lots of trickery.

I think doing something like this would work:
```
SANITIZED_IP=$($IP2UTIL route get $OCF_RESKEY_ip 2> /dev/null | awk '$1~/:/ {print $1}  $2~/:/ {print $2}')
if [ -n "$SANITIZED_IP" ]; then
    OCF_RESKEY_ip="$SANITIZED_IP"
fi
```

kskmori:
I was thinking if it could return OCF_ERR_CONFIGURED here when an invalid address was specified and hence `ip` command failed, but I now reconsider that it would be too much complicated and more confusing if `ip route get` was failed with other reasons.

I'm going to create a PR exactly as you suggested. Thanks.

kskmori:
submitted as #1396

oalbrigt:
Closing in favor of @kskmori PR.

PR #1110, opened 2018-02-21
redirect standard output/error of selinuxenabled

ZeTopHat:
I noticed servers without selinux installed get spammed in the messages log by the monitor operation output of that command:
e.g.
lrmd[]:   notice: operation_finished: nfs-server_monitor_10000:6025:stderr [ /usr/lib/ocf/resource.d/heartbeat/nfsserver: line 183: selinuxenabled: command not found ]


oalbrigt:
Thanks.

PR #1111, opened 2018-02-22
sybaseASE: new resource agent

oalbrigt:
Based on https://github.com/ClusterLabs/resource-agents/pull/398 with fixes and improvements.


heartbeat/sybaseASE
```diff
@@ -0,0 +1,895 @@
...
+#                <!--Checks to see if we can read from the mountpoint -->
+#                <action name="status" depth="10" timeout="100" interval="120" />
+#                <action name="monitor" depth="10" timeout="100" interval="120" />
+#
+#                <action name="meta-data" timeout="5" />
+#                <action name="validate-all" timeout="5" />
+#     The timeout value is not supported by Redhat in RHCS5.0.
+#
+# (3) This script should be put under /usr/share/cluster. Its owner should be "root" with
+#     execution permission.
```
krig: Is this really true, still?

heartbeat/sybaseASE
```diff
@@ -0,0 +1,895 @@
...
+# Declare some variables we will use in the script. Please don't change their values. #
+#######################################################################################
+declare login_string=""
+declare RUNSERVER_SCRIPT=$OCF_RESKEY_sybase_home/$OCF_RESKEY_sybase_ase/install/RUN_$OCF_RESKEY_server_name
+declare CONSOLE_LOG=$OCF_RESKEY_sybase_home/$OCF_RESKEY_sybase_ase/install/$OCF_RESKEY_server_name.log
+
+##################################################################################################
+# This function will be called by rgmanager to get the meta data of resource agent "sybaseASE". #
+# NEVER CHANGE ANYTHING IN THIS FUNCTION.
+##################################################################################################
```
krig: How about dropping the "never change anything..." comments? :)

heartbeat/sybaseASE
oalbrigt: Oh. It was made for RGManager first, and then ported, so I'll remove that part :)

heartbeat/sybaseASE
oalbrigt: Good idea. Wont be able to fix bugs if we cant change anything.

PR #1114, opened 2018-02-23
new resource agent azure-phoenix

MSSedusch:
Add a new resource agent that is part of a new fencing mechanism for Azure virtual machines.
The fence operation is splitted into a fence agent (for off operation) and this resource agent (for on operation) to work around a limitation in pacemaker that makes the cluster wait for completion of the overall reboot operation.

Fence Agent PR: https://github.com/ClusterLabs/fence-agents/pull/176

### DIFF
--- a/heartbeat/azure-phoenix.py
+++ b/heartbeat/azure-phoenix.py
@@ -2,6 +2,7 @@
 
 import sys, re, os, subprocess, time, syslog
 import logging
+import azure_fence_lib
 
 OCF_SUCCESS = 0
 OCF_ERR_GENERIC = 1
@@ -40,19 +41,8 @@ class SyslogLibHandler(logging.StreamHandler):
 		syslog.syslog(syslog_level, msg.replace("\x00", "\n"))
 		return
 
-class AzureConfiguration:
-    RGName = None
-    VMName = None
-    SubscriptionId = None
-    Cloud = None
-    UseMSI = None
-    Tenantid = None
-    ApplicationId = None
-    ApplicationKey = None
-    Verbose = None
-
 def get_azure_config():
-    config = AzureConfiguration()
+    config = azure_fence_lib.AzureConfiguration()
 
     config.RGName = os.environ.get("OCF_RESKEY_resourceGroup")
     config.VMName = os.environ.get("OCF_RESKEY_vmName")
@@ -99,7 +89,7 @@ def check_azure_config(config):
     if not config.SubscriptionId:
         fail_usage("Parameter subscriptionId required.")
 
-    if not ocf_is_true(config.UseMSI):
+    if not azure_fence_lib.ocf_is_true(config.UseMSI):
         if not config.Tenantid:
             fail_usage("Parameter tenantId required if Service Principal should be used.")
         if not config.ApplicationId:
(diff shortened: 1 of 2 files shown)


heartbeat/azure-phoenix.py
```diff
@@ -0,0 +1,417 @@
...
+    print "                     tenantId, applicationId and applicationKey are ignored."
+    print "  tenantId (optional): Id of the Azure Active Directory tenant."
+    print "                       Only required if a Service Principal should be used"
+    print "  applicationId (optional): Application ID of the Service Principal."
+    print "                            Only required if a Service Principal should be used"
+    print "  applicationKey (optional): Authentication key of the Service Principal."
+    print "                             Only required if a Service Principal should be used"
+
+def print_metadata():
+    print "<?xml version=\"1.0\"?>"
```
krig: Why not use Python triple-quoted strings here, for clarity?

heartbeat/azure-phoenix.py
```diff
@@ -0,0 +1,417 @@
+#!/usr/bin/python
```
krig: Just a note on Python 2 vs. 3: Python 2 will be unmaintained by upstream by 2020, which is not that far away anymore :( I know at least two major Linux distributions that are moving towards not even providing Python 2 anymore. So for any new code written today, I would recommend at least making it Python 3-compatible, if not Python 3 only.

See: 
* https://pythonclock.org/
* https://docs.python.org/3/howto/pyporting.html

heartbeat/azure-phoenix.py
```diff
@@ -0,0 +1,417 @@
...
+    config.Cloud = os.environ.get("OCF_RESKEY_cloud")
+    config.UseMSI = os.environ.get("OCF_RESKEY_useMSI")
+    config.Tenantid = os.environ.get("OCF_RESKEY_tenantId")
+    config.ApplicationId = os.environ.get("OCF_RESKEY_applicationId")
+    config.ApplicationKey = os.environ.get("OCF_RESKEY_applicationKey")
+    config.Verbose = os.environ.get("OCF_RESKEY_verbose")
+
+    if len(sys.argv) > 2:
+        for x in range(2, len(sys.argv)):
+            argument = sys.argv[x]
```
anandramakrishna: make arguments such as resourcegroup, etc. be dynamically discovered through the metadata API instead of via configuration.

heartbeat/azure-phoenix.py
```diff
@@ -0,0 +1,417 @@
...
+    config.VMName = os.environ.get("OCF_RESKEY_vmName")
+    config.SubscriptionId = os.environ.get("OCF_RESKEY_subscriptionId")
+    config.Cloud = os.environ.get("OCF_RESKEY_cloud")
+    config.UseMSI = os.environ.get("OCF_RESKEY_useMSI")
+    config.Tenantid = os.environ.get("OCF_RESKEY_tenantId")
+    config.ApplicationId = os.environ.get("OCF_RESKEY_applicationId")
+    config.ApplicationKey = os.environ.get("OCF_RESKEY_applicationKey")
+    config.Verbose = os.environ.get("OCF_RESKEY_verbose")
+
+    if len(sys.argv) > 2:
```
oalbrigt: You can remove this entire section, as the above os.environ.get()'s will get them when you do e.g. 'OCF_RESKEY_resourceGroup="rg" azure-phoenix.py', which is how the resource agents are supposed to be run.

heartbeat/azure-phoenix.py
oalbrigt: I'll add some logic to be able to change this.

heartbeat/azure-phoenix.py
oalbrigt: Great suggestion.

heartbeat/azure-phoenix.py
MSSedusch: I was hoping that we could have the same build system as we do for the fence-agents. But I am by no means familiar with automake. @oalbrigt Would it be possible to use the same build system or to add python support for the resource-agent repo as well?

heartbeat/azure_fence_lib.py
```diff
@@ -0,0 +1,406 @@
...
+        logging.warning("Not able to use metadata service. Am I running in Azure?")
+
+    return None
+
+def fail_usage(message):
+    raise ValueError("%s Run with parameter help to get more information" % message)
+
+# https://stackoverflow.com/questions/715417/converting-from-a-string-to-boolean-in-python
+def ocf_is_true(strValue):
+    return strValue and strValue.lower() in ("yes", "true", "1", "YES", "TRUE", "ja", "on", "ON")
```
krig: There is no need to include upper-case words in the comparison list since the string is converted to lower-case before checking.

oalbrigt:
I'm working on integrating all the functionality into the fence agent, so hopefully we wont need this.

MSSedusch:
@oalbrigt what do you mean you want to integrate it into the fence agent?

oalbrigt:
I'm adding the network fencing to fence_azure_arm, so that it can report successful fencing when it has blocked the network.

MSSedusch:
But we still need the resource agent. Otherwise a stonith operation reboot will block untill off and on are executed which is too long

krig:
@oalbrigt As I understand it, the problem that this agent aims to solve is that the "reboot" fencing action in Pacemaker has to wait for the action to complete before restarting services. Which means that even though the node can be confirmed to be "off" and is coming back online, services remain stopped until the node has booted up and rejoined the cluster. To just add an early return from the reboot action is not sufficient, since part of the fencing of the network variant is to block the rebooting node from the network, and those blocks have to stay in place until the node really is back up. So the workaround is to have a resource agents which manages the network block, so that the fence agent can return early. @MSSedusch if I misunderstood and this is incorrect, I apologise :)

MSSedusch:
@krig that is correct. The fence agent takes care of the off operation and the resource agent does the on operation.

oalbrigt:
Yeah, but I found a better way to do it with the fence_azure_arm agent by adding a few more lines of code using the new library by @MSSedusch. I'm just going to do some more testing and improvements and I'll make a PR on fence-agents.

MSSedusch:
we found a major roadblock with the approach implemented in this fence agent. Hence closing the PR

oalbrigt:
New PR: https://github.com/ClusterLabs/fence-agents/pull/181

PR #1116, opened 2018-02-28
doc: RA dev guide: fix unnecessary quotations in examples

kskmori:
It fails with an error if an user wrote an RA as the examples verbatim.

I believe that it's pretty obvious so I'm going to merge it now.


oalbrigt:
Thanks.

PR #1118, opened 2018-03-03
syslog-ng resource agent improvements and extensions

jszigetvari:
* Added parameters to the syslog-ng resource agent in a backwards compatible way
* Added resource agent script for Syslog-ng Premium Edition 6
* Added resource agent script for Syslog-ng Premium Edition 7

(The Premium Edition is the non-free counterpart of the Syslog-ng open-source version. Their parameters, and install paths differ from that of the open-source edition. There are also slight differences between the PE 6 and PE 7 versions.)

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -156,6 +156,8 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
 			  ocf_heartbeat_mpathpersist.7 \
                           ocf_heartbeat_symlink.7 \
                           ocf_heartbeat_syslog-ng.7 \
+                          ocf_heartbeat_syslog-ng-pe-6.7 \
+                          ocf_heartbeat_syslog-ng-pe-7.7 \
                           ocf_heartbeat_tomcat.7 \
                           ocf_heartbeat_varnish.7 \
                           ocf_heartbeat_vmware.7 \
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -155,6 +155,8 @@ ocf_SCRIPTS	     =  AoEtarget		\
 			slapd			\
 			symlink			\
 			syslog-ng		\
+			syslog-ng-pe-6		\
+			syslog-ng-pe-7		\
 			tomcat			\
 			varnish			\
 			vmware			\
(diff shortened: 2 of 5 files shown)


krig:
Make sure to add the new resource agents to the makefiles for installation, testing and manpage generation to work correctly:

`heartbeat/Makefile.am`
`doc/man/Makefile.am`

jszigetvari:
@krig Thanks for the guidance! Please let me know if I have missed anything else.

oalbrigt:
Can we have these in one agent? There's not much difference between the original one and these, so it'd be better to have some auto-detection and/or parameter to use the additional functionality depending on the version.

jszigetvari:
@oalbrigt The Agent for the open-source edition of syslog-ng has been around for quite some time, and in its case I have put in extra effort to remain backwards compatible, so that users would be able to continue with their old settings without modification.

In case of the RAs for the Premium Edition, I could merge the two RAs, but that would necessitate version checks on nearly every invocation, which IMO would not be so elegant, plus PE-6 has a CLI option that PE-7 doesn't support, so in case of PE-7 having that option would be pointless.

Either way, I can merge the two PE RAs to one, if you'd like, but IMO it would be worth to keep them separate from the open source version, so that I wouldn't be limited by having to maintain backwards compatibility with it.

oalbrigt:
Thanks for your feedback. I think I'll use some time later this week to combine all 3 into one agent (where it defaults to current behaviour, and you can set a parameter to pe6 or pe7 when you're using it with those versions.

jszigetvari:
@oalbrigt Okay. Please let me know if you need any help in the effort. (eg. getting version strings of the PE versions in order to differentiate between them.)

oalbrigt:
I've made a new PR where I've merged your code with the main syslog-ng agent: https://github.com/ClusterLabs/resource-agents/pull/1135

There is an example of what to set the parameters to for Premium Edition in the metadata.

Tell me if it works as expected and whether the --qdisk-dir parameter is only available for specific Premium Edition version or not.

jszigetvari:
@oalbrigt I will try to test your new modifications in a few hours
The --qdisk-dir option is only available in PE6, and is absent in the OSE and PE7 versions.

oalbrigt:
Merged into main agent in: https://github.com/ClusterLabs/resource-agents/pull/1135

PR #1119, opened 2018-03-05
mpathpersist: Fixed problem that character string could not be correc…

yuusuke:
…tly parsed

In "device-mapper-multipath-0.4.9-111.el7.x86_64" I use, the output of the command is as follows.
A space appears between Key and Equals.
```
# mpathpersist --in --read-reservation /dev/mapper/lun4
  PR generation=0x7fb, Reservation follows:
   Key = 0xc0a86506
  scope = LU_SCOPE, type = Write Exclusive
```
In order to correctly grep you need to modify the pattern.

### DIFF
--- a/heartbeat/mpathpersist
+++ b/heartbeat/mpathpersist
@@ -271,7 +271,7 @@ mpathpersist_get_status() {
                         RESERVED_DEVS+=($dev)
                     fi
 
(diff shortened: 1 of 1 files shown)


krig:
Is it possible that this output has changed between versions of mpathpersist? Perhaps the pattern needs to take both outputs into consideration.

krig:
Regarding my previous comment: I checked the source, that's not the case. Thanks!

PR #1120, opened 2018-03-05
New RA that manages IP moving in GCP

dud225:
Hello

This Resource Agent uses the GCP API to move a floating IP addresse within a GCP VPC by changing an entry in the routing table, as discussed in the [documentation](https://cloud.google.com/solutions/best-practices-floating-ip-addresses#option_4_failover_using_routes_api_calls).

This agent also configures the floating IP locally on the instance OS.

### DIFF
--- a/doc/man/Makefile.am
+++ b/doc/man/Makefile.am
@@ -113,6 +113,7 @@ man_MANS	       = ocf_heartbeat_AoEtarget.7 \
                           ocf_heartbeat_fio.7 \
                           ocf_heartbeat_galera.7 \
                           ocf_heartbeat_garbd.7 \
+                          ocf_heartbeat_gcp-vpc-move-ip.7 \
                           ocf_heartbeat_iSCSILogicalUnit.7 \
                           ocf_heartbeat_iSCSITarget.7 \
                           ocf_heartbeat_iface-bridge.7 \
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -110,6 +110,7 @@ ocf_SCRIPTS	     =  AoEtarget		\
 			fio			\
 			galera			\
 			garbd			\
+			gcp-vpc-move-ip		\
 			iSCSILogicalUnit	\
 			iSCSITarget		\
 			ids			\
--- a/heartbeat/gcp-vpc-move-ip
+++ b/heartbeat/gcp-vpc-move-ip
@@ -1,0 +1,372 @@
+#!/bin/bash
+#
+#
+# OCF resource agent to move an IP address within a VPC in GCP
+#
+# License: GNU General Public License (GPL)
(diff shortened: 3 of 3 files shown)


oalbrigt:
Wouldnt it be better to use IPaddr2 for the ip-functionality, and only use the ip command where it's necessary for checks?

dud225:
Hello @oalbrigt 

that's a excellent question.
I've tought about it quite some time, here are my arguments : 
- IPaddr2 is more suited to conventional LAN networks where the admin manages the whole network. In particular  IPaddr2 regularly sends gratuitous ARPs to the others nodes but these frames are completely [ignored in GPG](https://cloud.google.com/solutions/best-practices-floating-ip-addresses#challenges_with_migrating_floating_ips_to_compute_engine)
- this code is based on the rival RA aws-vpc-move-ip that is also managing the IP itself. I've asked the original commiter @krig the same question, but I haven't received any feedback yet.

Regards
dud

krig:
@dud225 @oalbrigt Ah yes, I meant to reply to that, thanks for reminding me :) I'm not the original author so I can't speak to the motivation for why it doesn't rely on IPaddr/IPaddr2 for the actual IP management. I just assumed it was due to AWS-specific needs...

kskmori:
From my experience of AWS, though I'm not the original author of aws-vpc-move-ip but I've developed a similar RA on AWS before, it is necessary to use routing tables because a virtual IP cannot migrate across Availability Zones since they must have different subnets on AWS. (I heard that it is called the routing based HA cloud design pattern, but can not find a good reference right now)

The situation seems quite different on GCP according to the document above, but I feel like that it is a good thing to follow what the official documents tell you to do so.

Just my two cents.

oalbrigt:
Sounds good to me.

dud225:
> so I can't speak to the motivation for why it doesn't rely on IPaddr/IPaddr2 for the actual IP management. I just assumed it was due to AWS-specific needs...

Although I haven't tested it, I don't think there any constraints from either AWS or GCP preventing the use of IPaddr2, it's just that the ARP frames will be ignored.

AFAIU the others AWS RA work differently : 
- awseip : AWS automatically enforces NAT that translates the EIP to the private IP, so that this is transparent to the instance
- awsvip : the instance needs to have the secondary IP locally configured but fortunately as from the [doc](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/MultipleIP.html#StepTwoConfigOS)  there is a package that takes care all of it. On Debian this package doesn't exist but the Debian team that manages the official Debian AMI includes the [needed files inside each AMI](https://wiki.debian.org/Cloud/AmazonEC2Image/Jessie). So IPaddr2 may not be needed if there is another mechanism already present

So the whole point to my decision is to avoid unneeded stuff and to make things easier to configure ( I admit that this last point is kind of a weak argument).

PR #1124, opened 2018-03-14
Medium: exportfs: Fix square bracket stripping in clientspec

nrwahl2:
This patch fixes the issue of the `exportfs` RA failing its monitor
operation if the `clientspec` matches either of two conditions:
    
- It includes a character class wildcard at the beginning or end of
  the string.
    
    Example:
        `clientspec='[ab]-host.example.com'`
    
  In this case, the leading square bracket is incorrectly stripped,
  leaving `'ab]-host.example.com'`.
    
- It consists of an IPv6 network with a CIDR netmask.
    
    Example:
          `clientspec='[fe80::]/64'`
    
  In this case, the closing square bracket is incorrectly left intact,
  leaving `'fe80::]/64'`.
    
In either of these conditions, the `start` operation creates the export
successfully, but the `monitor` immediately fails when it does not find
the string it's looking for. Similarly, the `stop` operation succeeds but 
does not actually remove the export.
    
This patch replaces the `sed` operation that strips brackets so that:
  - It only matches IPv6 addresses (using a colon to distinguish).
  - It matches the closing bracket in an IPv6 network.

### DIFF
--- a/heartbeat/exportfs
+++ b/heartbeat/exportfs
@@ -272,7 +272,17 @@ exportfs_monitor ()
 		return $OCF_NOT_RUNNING
 	fi
 
-	spec="$(echo "$OCF_RESKEY_clientspec" | sed -e 's/^\[*//' -e 's/\]*$//')"
+	# IPv6 addresses and networks are encased in brackets that need
+	# to be removed
+        case "$OCF_RESKEY_clientspec" in
+		*:*:* )
+			spec="$(echo "$OCF_RESKEY_clientspec" | tr -d '[]')"
+			;;
+		*)
+			spec="$OCF_RESKEY_clientspec"
+			;;
+	esac
+
 	if forall is_exported "$spec"; then
 		if [ ${OCF_RESKEY_rmtab_backup} != "none" ]; then
 			forall backup_rmtab


oalbrigt:
Maybe we could do do an if with grep -q ":.*:" to detect IPv6 IPs and not remove the square brackets for those.

nrwahl2:
@oalbrigt Could you show me what you have in mind? In the current `monitor` logic, IPv6 addresses need to have their square brackets stripped.

Edit: Skip to the bottom for a tl;dr of the better idea you gave me.

Let me try to clarify my thought process to make sure we're on the same page (and in case my whole premise is dumb ;) )

Square brackets are specified in the `clientspec` and passed to the `exportfs` command, where they're used as a special designator to say "This is an IPv6 address." Without square brackets in `clientspec`, processing breaks on the colon character in an IPv6 address.

From `man 8 exportfs`:

> IPv6 presentation addresses contain colons, which are already used to separate the "host" and "path" command line arguments. When specifying a client using a raw IPv6 address, enclose  the  address in square brackets. For IPv6 network addresses, place the prefix just after the closing bracket.

When running `exportfs` with no arguments (as in `format_exports()`) to display the list of current exports, IPv6 addresses do NOT contain square brackets. This is the output that the `exportfs_monitor()` function searches through, and it is the reason that brackets get stripped in `spec` in `exportfs_monitor()`.

However, when an IPv6 _**network**_ is used, the closing bracket is left intact because it does not come at the end of the string. Additionally, in hostnames, a square bracket at the beginning or end of the string would be stripped, even though it's used to mark a character class wildcard.

```
# Scenario A
# exportfs fe80:::/mnt        <-- doesn't work without brackets
exportfs: Invalid export syntax: fe80
# exportfs [fe80::]:/mnt      <-- creates export
# exportfs -v                 <-- displayed export does NOT contain square brackets
/mnt          	fe80::(ro,sync,wdelay,hide,no_subtree_check,sec=sys,secure,root_squash,no_all_squash)

# Scenario B
# exportfs fe80::/64:/mnt      <-- doesn't work without brackets
exportfs: Invalid export syntax: fe80
# exportfs [fe80::]/64:/mnt    <-- creates export
# exportfs -v                  <-- displayed export does NOT contain square brackets
/mnt          	fe80::/64(ro,sync,wdelay,hide,no_subtree_check,sec=sys,secure,root_squash,no_all_squash)

# Scenario C
# exportfs [ab]-host.example.com:/mnt    <-- brackets used to indicate character list
# exportfs -v                            <-- displayed export DOES contain square brackets
/mnt          	[ab]-host.example.com(ro,sync,wdelay,hide,no_subtree_check,sec=sys,secure,root_squash,no_all_squash)
```

Results with the current logic, where `spec` is the pattern used to grep through the output of `exportfs`:
- Scenario A: OK
  `clientspec="[fe80::]"` --> `spec="fe80::"` 
- Scenario B: FAILS
  `clientspec="[fe80::]/64"` --> `spec="fe80::]/64"`
- Scenario C: FAILS
  `clientspec="[ab]-host.example.com"` --> `spec="ab]-host.example.com"`


Your comment leads to a perhaps better idea than my original one of changing the `sed` expression. We could look for ":.*:" to detect IPv6 IPs and DO remove square brackets from those. We could do away with `sed` altogether.

Before (current `master`):
```
        spec="$(echo "$OCF_RESKEY_clientspec" | sed -e 's/^\[*//' -e 's/\]*$//')"
```
My original proposition in this PR:
```
        spec="$(echo "$OCF_RESKEY_clientspec" | sed -r 's/^\[(.*:.*)\]/\1/')"
```
New alternative:
```
        # Remove square brackets from IPv6 addresses and networks
        if [[ $OCF_RESKEY_clientspec =~ :.*: ]]; then
                spec="$(echo "$OCF_RESKEY_clientspec" | tr -d '[]')"
        fi
```

oalbrigt:
Looks good to me. I would only update it to
> if [ $OCF_RESKEY_clientspec =~ :*: ]; then
to avoid introducing a bashism, and * should be good enough in shell scripts.

nrwahl2:
Single brackets with `=~` regex matching is a syntax error:
`[ $OCF_RESKEY_clientspec =~ :*: ]`

`:*:` as a regex would match a single colon -- we want two colons.
`[[ $OCF_RESKEY_clientspec =~ :*: ]]`

Single brackets with an `=` will expand `*` into a glob, expanding to filenames in the working directory.
`[ $OCF_RESKEY_clientspec = :*: ]`


If we want POSIX compliance and no bash double brackets, then we could go this route:
```
if ( echo $OCF_RESKEY_clientspec | grep -q ':.*:' ); then
```

It's a little messier but works. If you know of a way to make `:*:` act as two colons separated by zero or more characters in a single-bracket test, let me know and I'll use it.

oalbrigt:
Ah. That's right. You could use case with ":*:".

nrwahl2:
That's true, case would work with `*:*:*` (need asterisks on both sides). Could either go with the `if-else` approach above:
```
# Remove square brackets from IPv6 addresses and networks
if ( echo "$OCF_RESKEY_clientspec" | grep -q ":.*:" ); then
        spec="$(echo "$OCF_RESKEY_clientspec" | tr -d '[]')"
else
        spec="$OCF_RESKEY_clientspec"
fi
```
or the `case` statement:
```
# Remove square brackets from IPv6 addresses and networks
case "$OCF_RESKEY_clientspec" in
        *:*:* )
                spec="$(echo "$OCF_RESKEY_clientspec" | tr -d '[]')"
                ;;
        *)
                spec="$OCF_RESKEY_clientspec"
                ;;
esac
```

Up to you as the `resource-agents` boss man. I suppose the `if-else` comes off as redundant by comparison. Regardless, I don't see us ever expanding this to add more cases unless the `exportfs` command changes. Lemme know and I'll get the PR updated.

oalbrigt:
I think the case-statement would be more effective (no calling external commands).

nrwahl2:
Makes good sense. Updated, pending CI.

PR #1129, opened 2018-03-23
CI: fixes for bash path, strncpy in GCC 8 and missing docbook-style-xsl

oalbrigt:


configure.ac
```diff
@@ -919,9 +927,75 @@ doc/Makefile							\
...
+		rgmanager/src/resources/lvm_by_vg.sh		\
 	rgmanager/src/resources/utils/Makefile			\
+		rgmanager/src/resources/utils/fs-lib.sh		\
+		rgmanager/src/resources/utils/messages.sh	\
+		rgmanager/src/resources/utils/config-utils.sh	\
+		rgmanager/src/resources/utils/member_util.sh	\
+		rgmanager/src/resources/utils/ra-skelet.sh	\
 		)
 
+dnl Files we output that need to be executable
```
krig: Shouldn't all the RAs be executable? Seems like this should be achievable with a script that loops over all the agents rather than having to have another list that people adding new agents have to remember to modify.

oalbrigt: This is because we generate these from .in-files be able to update the path for bash.

PR #1132, opened 2018-04-04
enable chap authentication for lio-t iscsi target implementation. if …

stefanfriedel:
…attribute 'incoming_username' is set: enable attribute 'authentication', disable attribute 'generate_node_acls' and set chap 'userid' and 'password' for each 'allowed_initiator'

Tested on Debian stretch w/ targetcli.

Default values for attributes authentication=1 and generate_node_acls=0 are (re-)set to avoid problems if attributes are manually changed in targetcli.


krig:
Looks good to me.

oalbrigt:
add to whitelist

oalbrigt:
LGTM. CI failure unrelated to patch.

PR #1139, opened 2018-04-16
Correctly pickup old keys (improved v 0.0.2)

Splarv:
If a comp hardly poweroffed there are staled old registration keys on the scsi
device. On a poweron the module sg_persist recognizes them by the monitor-probe and
set itself state to the "slave" state without "start" action and thus
don't set ACT_MASTER_SCORE. And without it the comp can never be
"master". This patch is for fixing this. It add ACT_MASTER_SCORE or
remove ACT_MASTER_SCORE in the "monitor" if the "monitor" see that the
keys unexpectedly exists or absent, thus simulate normal behaviour of
the "start" or "stop".


oalbrigt:
Looks good to me.

PR #1141, opened 2018-04-19
Enhance galera to interact over multiple clusters

zzzeek:
This change adds a new parameter "remote_node_map"
to the Galera resource agent which allows it to consider
galera node names that are in other clusters as part of its
Galera quorum.  To achieve this, it launches pcs commands
over SSH to the remote clusters in order to view and modify
remote state variables.

WIP.


fabbione:
add to whitelist

heartbeat/galera
```diff
@@ -215,6 +218,7 @@ mapping in option cluster_host_map.
 <content type="string" default=""/>
 </parameter>
 
+
```
oalbrigt: Why are you adding empty lines to the longdesc?

heartbeat/galera
```diff
@@ -429,11 +466,56 @@ set_master_score()
...
+    local remote_ssh=$(get_remote_node $node)
+
+    if [ -z "$remote_ssh" ]; then
+        ${HA_SBIN_DIR}/crm_attribute -N $node $@
+    else
+        $SSH_CMD $remote_ssh ${HA_SBIN_DIR}/crm_attribute -N $node $@
     fi
 }
 
+
```
oalbrigt: Another new empty line with no reason.

heartbeat/galera
```diff
@@ -429,11 +466,56 @@ set_master_score()
...
+remote_crm_master()
+{
+
+    local node=$1
+    shift
+
+    local remote_ssh=$(get_remote_node $node)
+
+    if [ -z "$remote_ssh" ]; then
+        $CRM_MASTER -N $node $@
```
oalbrigt: CI fails due to missing quotes around $@ it seems:
heartbeat/galera:498:30: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
heartbeat/galera:500:76: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
heartbeat/galera:512:47: error: Double quote array expansions to avoid re-splitting elements. [SC2068]
heartbeat/galera:514:68: error: Double quote array expansions to avoid re-splitting elements. [SC2068]

heartbeat/galera
zzzeek: likely an editor artifact.    I'm also looking for reaction to the overall concept of embedding SSH commands into a resource agent and communicating with other clusters (hence WIP).

zzzeek:
cc @beekhof

heartbeat/galera
```diff
@@ -68,6 +68,24 @@
...
+# get at the "other" cluster, and to authenticate I just have id_rsa keys
+# copied to all the controllers under "root", since that's who can run
+# pcs commands.  So the RA on a docker container has root access to all
+# the controller nodes.
+#
+# So we would like to either:
+#
+# a. ssh as some other locked down account that can still run the
+# necessary pcs commands on the other cluster? or some setuid script that
+# does the things we need?
```
JAORMX: This sounds like the nicer solution.

heartbeat/galera
```diff
@@ -68,6 +68,24 @@
...
+#
+# b. totally other means of invoking pcs commands on remote cluster?  web
+# service or something?
+#
+SSH_CMD="${SSH} -oStrictHostKeyChecking=no"
```
JAORMX: can we make this configurable?

zzzeek: yeah, what other commands might we be using?

beekhof:
TBH, this scares the heck out of me.

At the minimum, can we put this in a new agent (source the original then add the new functionality)?

I can see the need for special handling (what to do if the remote isn't reachable for example) that I'd not want to complicate the non-stretch version with.

zzzeek:
@beekhof as far as sourcing the original, I will need to add a line to the case statement at the end that upon passing a command like "sourceonly" does a simple "return", to avoid the "exit" call.    am not finding any bash magic to source all the variables and having them survive past an "exit".

zzzeek:
@beekhof note this change no longer writes any data over SSH, and only reports status, and to that end we can also replace the usage of SSH with an xinetd service linked to a script.   however, in the xinetd case, authentication and encryption go out the window as far as I understand being able to create xinetd services.   the security audit will then say, if the xinetd script itself has a vulnerability, it's openly exposed.

another option is to use SSH but to limit the commands / scripts that the user can execute.  this can be done either with a custom shell in /etc/passwd or apparently you can limit commands for a specific key in authorized_keys, been googling that a bit.   I think having a "front end" that is reached via ssh but nonetheless is just a single script with an argument is the least we can do.

oalbrigt:
@zzzeek can you fix the quote-issues reported by Travis CI?

PR #1143, opened 2018-04-22
Low: Improve awsvip awseip

guessi:
Improved:

## awseip
- High: fix allocation_id not found error

## awsvip
- Low: get secondary-private-ip more precisely
- Low: get network-id from meta-data directly

### DIFF
--- a/heartbeat/awseip
+++ b/heartbeat/awseip
@@ -4,7 +4,7 @@
 #    Manage Elastic IP with Pacemaker
 #
 #
-# Copyright 2016 guessi <guessi@gmail.com>
+# Copyright 2016-2018 guessi <guessi@gmail.com>
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/heartbeat/awseip
+++ b/heartbeat/awseip
@@ -182,9 +182,8 @@ awseip_start() {
 awseip_stop() {
     awseip_monitor || return $OCF_SUCCESS
 
-    ASSOCIATION_ID=$($AWSCLI --profile $OCF_RESKEY_profile --output json ec2 describe-addresses \
-                         --allocation-id ${ALLOCATION_ID} | grep -m 1 "AssociationId" | awk -F'"' '{print$4}')
-    $AWSCLI --profile $OCF_RESKEY_profile ec2 disassociate-address  \
+    $AWSCLI --profile ${OCF_RESKEY_profile} \
+        ec2 disassociate-address \
         --association-id ${ASSOCIATION_ID}
     RET=$?
 
@@ -200,7 +199,9 @@ awseip_stop() {
 }
 
 awseip_monitor() {
-    $AWSCLI --profile $OCF_RESKEY_profile ec2 describe-instances --instance-id "${INSTANCE_ID}" | grep -q "${ELASTIC_IP}"
+    $AWSCLI --profile ${OCF_RESKEY_profile} \
+        ec2 describe-addresses \
+            --allocation-id ${ALLOCATION_ID}
     RET=$?
 
     if [ $RET -ne 0 ]; then
--- a/heartbeat/awseip
+++ b/heartbeat/awseip
@@ -182,8 +182,9 @@ awseip_start() {
 awseip_stop() {
     awseip_monitor || return $OCF_SUCCESS
 
-    $AWSCLI --profile ${OCF_RESKEY_profile} \
-        ec2 disassociate-address \
+    ASSOCIATION_ID=$($AWSCLI --profile $OCF_RESKEY_profile --output json ec2 describe-addresses \
+                         --allocation-id ${ALLOCATION_ID} | grep -m 1 "AssociationId" | awk -F'"' '{print$4}')
+    $AWSCLI --profile $OCF_RESKEY_profile ec2 disassociate-address  \
         --association-id ${ASSOCIATION_ID}
     RET=$?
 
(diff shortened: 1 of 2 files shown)


heartbeat/awseip
```diff
@@ -182,9 +182,8 @@ awseip_start() {
 awseip_stop() {
     awseip_monitor || return $OCF_SUCCESS
 
-    ASSOCIATION_ID=$($AWSCLI --profile $OCF_RESKEY_profile --output json ec2 describe-addresses \
-                         --allocation-id ${ALLOCATION_ID} | grep -m 1 "AssociationId" | awk -F'"' '{print$4}')
-    $AWSCLI --profile $OCF_RESKEY_profile ec2 disassociate-address  \
+    $AWSCLI --profile ${OCF_RESKEY_profile} \
+        ec2 disassociate-address \
         --association-id ${ASSOCIATION_ID}
```
guessi: @oalbrigt please consider to merge this commit to avoid `ASSOCIATION_ID` become `empty` while command failure, I think it would be better to move allocation_id check to `awseip_monitor()` instead.

oalbrigt: It's not a check. It's only used to disassociate the address, which is why it's defined in stop(). See line 187.

guessi: @oalbrigt 
imagine the Elastic IP have been disassociated with the node, it will lead to a empty string for the variable `ASSOCIATION_ID`, and `awscli` will complain that there's a no value error on ` --association-id ${ASSOCIATION_ID}` at line 187, isn't it?

guessi: I still insist to remove the existence check of `${ASSOCIATION_ID}` to make the code cleaner

line 201 ~ 204 in `awseip_monitor()` have already checked the existence of `${ASSOCIATION_ID}`
if it doesn't exist, `awseip_monitor()` would return a non-zero value, then trigger `... || return $OCF_SUCCESS` at line 183

oalbrigt: That is ALLOCATION_ID, which is defined by the allocation_id-parameter.

ASSOCIATION_ID changes each time the IP is associated with a node, which is why we get the information from AWS.

guessi: okay, got your point, reverting back.

guessi: @oalbrigt reverted, and re-commit. please review.

heartbeat/awsvip
```diff
@@ -203,7 +207,8 @@ esac
 AWSCLI="${OCF_RESKEY_awscli}"
 SECONDARY_PRIVATE_IP="${OCF_RESKEY_secondary_private_ip}"
 INSTANCE_ID="$(curl -s http://169.254.169.254/latest/meta-data/instance-id)"
-NETWORK_ID="$($AWSCLI --profile $OCF_RESKEY_profile --output json ec2 describe-instances --instance-id ${INSTANCE_ID} | grep -m 1 'eni' | awk -F'"' '{print$4}')"
+MAC_ADDRESS="$(curl -s http://169.254.169.254/latest/meta-data/mac)"
```
oalbrigt: Does this cause issues if there are multiple interfaces?

heartbeat/awseip
```diff
@@ -184,7 +184,14 @@ awseip_stop() {
 
     ASSOCIATION_ID=$($AWSCLI --profile $OCF_RESKEY_profile --output json ec2 describe-addresses \
                          --allocation-id ${ALLOCATION_ID} | grep -m 1 "AssociationId" | awk -F'"' '{print$4}')
-    $AWSCLI --profile $OCF_RESKEY_profile ec2 disassociate-address  \
+
+    if [ -z "${ASSOCIATION_ID}" ]; then
+        ocf_log info "ASSOCIATION_ID not found while stopping AWS Elastic IP"
+        return $OCF_NOT_RUNNING
+    fi
```
oalbrigt: Looks good to me.

oalbrigt:
add to whitelist

heartbeat/awsvip
guessi: I've tried to attach multiple eni, only one interface will return, the main eni.

The rest of others, will provided by another entry.

Should be fine here.

oalbrigt: Great. Thanks for checking.

guessi: Just created a new EC2 for double confirm, here's the operation log for your reference.

```
$ ip link show
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN mode DEFAULT group default qlen 1
    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
2: eth0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 9001 qdisc pfifo_fast state UP mode DEFAULT group default qlen 1000
    link/ether 06:a0:47:49:71:fa brd ff:ff:ff:ff:ff:ff
3: eth1: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
    link/ether 06:bd:7e:b3:cf:84 brd ff:ff:ff:ff:ff:ff

$ echo "$(curl -s http://169.254.169.254/latest/meta-data/mac)"
06:a0:47:49:71:fa

$ echo "$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/06:a0:47:49:71:fa/mac)"
06:a0:47:49:71:fa

$ echo "$(curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/06:bd:7e:b3:cf:84/mac)"
06:bd:7e:b3:cf:84
```

oalbrigt: Great. That's good to have here as documentation.

PR #1144, opened 2018-05-01
SAPInstance resource resilience improvements

OndrejHome:
Hi,

I have prepared some changes in this PR that are trying to improve the behaviour of `SAPInstance` resource agent (RA) when binaries needed by this RA gets missing.

- bd953cb move logging when we detect missing binaries earlier to see that this was issue even during 'stop' action
- f6fe133 - make the failure with missing binaries only 'hard' and not 'fatal' so this will only prevent resource from running on node where error was detected and not stop all resource instances on all nodes
- 64d9d49 - as we can use the binaries from both ASCS and ERS instances interchangeably for later operation make sure that we look for both of them. Without this only one location will be searched (except of common `/usr/sap/$SID/SYS/exe/run/`)
- 6069df2 - clear the `$DIR_EXECUTABLE` so that we properly detect when we are missing binaries when detection is done more than once during single operation. Currently the variable `$DIR_EXECUTABLE` may hold the path from previous run of `sapinstance_init` even if at the time of run some locations became unavailable. This tries to make shorter gap between the path detection and further actions. This is only effective when DIR_EXECUTABLE is not provided as resource attribute in which case there is only one path checked.

Changes were tested in RHEL 7.5 environment when trying to make binaries unavailable by removing execute bit on them (`chmod -x ...`) so `have_binary` function returns failure.

If anything is unclear/strange feel free to reach me out.

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -322,6 +322,8 @@ sapinstance_init() {
 	ASCSInstanceName=`echo $OCF_RESKEY_InstanceName| cut -d_ -f2`
   fi
 
+  # make sure that we don't care the content of variable from previous run of sapinstance_init
+  DIR_EXECUTABLE=""
   # optional OCF parameters, we try to guess which directories are correct
   if  [ -z "$OCF_RESKEY_DIR_EXECUTABLE" ]
   then
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -294,13 +294,13 @@ abnormal_end() {
     exit $?
   }
 
+  ocf_log err $err_msg
   if [ "$ACTION" = "stop" ]
   then
     cleanup_instance
     exit $OCF_SUCCESS
   fi
 
-  ocf_log err $err_msg
   exit $OCF_ERR_CONFIGURED
 }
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
add to whitelist

oalbrigt:
I have requested that some of the SAP maintainers evaluate it.

fdanapfel:
@krig: The SAPInstance and SAPDatabase resource agents are maintained by SUSE and Red Hat colleagues at the SAP LinuxLab together with people working outside of the SAP LinuxLab (Alexander Krauth used to work at the SAP LinuxLab but he now works somewhere else). 

SAP themselves have nothing to do with the maintenance of the resource agents.

Fabian Herschel and myself are currently reviewing these patches and will provide some feedback on them soon.

OndrejHome:
Communicating with @fdanapfel  I will be splitting the changes into separate PRs so they can be individually reviewed. Once done I will close this PR.

OndrejHome:
I have split the changes into separate PRs below.
https://github.com/ClusterLabs/resource-agents/pull/1154
https://github.com/ClusterLabs/resource-agents/pull/1155
https://github.com/ClusterLabs/resource-agents/pull/1156
https://github.com/ClusterLabs/resource-agents/pull/1157
We can discuss any concerns to given changes in those PRs. If there is any additional set of tests/proofs that needs to be run on them feel free to let me know.

Issue #1125, opened 2018-03-15
Openstack - Resource agent to manage floating IPs


mgrzybek:
### Description
In a bare-metal environment, resource agents such as IPAddr2 are used to move addresses. Moreover, an AWS EC2 agent has been written.

In order to use the same mechanism with Openstack, new resource agents should be provided. For example:

1. set node attributes like instance ID, ports ID + subnets ID…
2. use these attributes to attach a floating IP to a port

### Proposal

1. resource agent named `os-info` to populate the CIB
2. resource agent named `os-floating-ip` to manage the resource

That could be a good starting point for other agents such as `os-cinder-volume`.

### Questions

Do you think these names are OK? Any suggestion?

mbaldessari:
I am having a bit of trouble understanding the requirements here. Can you make some concrete example of the problems you're trying to solve?

krig:
I'm also not sure I understand the purpose of the proposed agents... as a side note, I would prefer `openstack-*` as prefix rather than `os-`, since `OS` can be interpreted as "operating system". At least, that would be my first association.

mgrzybek:
My IaaS team does not provide LBaaS service (using Neutron, Octavia…) or DBaaS.
As a consequence I need to deal with high-availability by myself. The corosync/pacemaker stack should be able to migrate an IP address or storage.

A "classic" resource group like:

```
Resource Group: g-lb
     haproxy		(systemd:haproxy):			Started node-1
     service-ip		(ocf::heartbeat:IPaddr2):		Started node-1

Clone Set: c-dlm-clvm [dlm-clvm]
     Started: [ node-1 node-2 node-3 ]

Resource Group: g-storage
     my-lvm-vg		(ocf::heartbeat:LVM):				Started node-2
     my-fs		(ocf::heartbeat:Filesystem):			Started node-2
```

Using the new RAs:

```
Clone Set: c-openstack-info [openstack-info]
     Started: [ node-1 node-2 node-3 ]

Resource Group: g-lb
     haproxy		(systemd:haproxy):			Started node-1
     service-ip		(ocf::heartbeat:openstack-floating-ip):	Started node-1

Resource Group: g-storage
     my-lun		(ocf::heartbeat:openstack-cinder-volume):	Started node-2
     my-lvm-vg		(ocf::heartbeat:LVM):				Started node-2
     my-fs		(ocf::heartbeat:Filesystem):			Started node-2


Node Attributes:
* Node node-1:
[...]

Issue #1089, opened 2018-01-11
[FreeBSD] service resource agent not supported


mgrzybek:
Configuring a resource, such as "service:hastd" does not work on BSD systems.
Pacemaker expects the RC script to be `/etc/init.d/hastd` instead of `/etc/rc.d/hastd`.

```
Jan 11 15:58:51 bsd1 lrmd[723]:  warning: Cannot execute '/etc/init.d/hastd': No such file or directory (2)
Jan 11 15:58:51 bsd1 crmd[726]:    error: Unable to retrieve meta-data for (null):heartbeat:hastd
Jan 11 15:58:51 bsd1 crmd[726]:    error: No metadata for service::(null):hastd
Jan 11 15:58:51 bsd1 crmd[726]:    error: Result of probe operation for hastd on bsd1: Not installed
```

The resource agent should use the `service` command or `/etc/rc.d`.

krig:
I’d recommend opening this issue against Pacemaker, the agents in this repository are OCF agents which use a different code path in Pacemaker :)

mgrzybek:
There is no Issue tab in https://github.com/ClusterLabs/pacemaker. I guess https://bugs.clusterlabs.org/ is the way to go. Isn't it ?

krig:
Yep, sadly ;)

oalbrigt:
We should move this to the openstack-resource-agents project (https://git.openstack.org/cgit/openstack/openstack-resource-agents/).

mgrzybek:
@oalbrigt : I think your reply should have been posted to #1147 instead :)

oalbrigt:
You're right. I have moved it there now :)

PR #1147, opened 2018-05-11
Openstack agents

mgrzybek:
Related to #1125 

Three agents are added:
* `openstack-info` - updates attributes (openstack_az, openstack_flavor, openstack_id, openstack_ports, openstack_region, openstack_tenant_id, openstack_tenant_name) ;
* `openstack-floating-ip` - manages a given floating IP (attach / detach from a node) ;
* `openstack-cinder-volume` - manages a given volume (attach / detach from a node)


vvidic:
Fix the title? :)

krig:
The agents also need to be added to `heartbeat/Makefile.am`.

mgrzybek:
@krig you were sceptical about #1125

Is it clearer for you now?

oalbrigt:
add to whitelist

oalbrigt:
The hardcoded bash-path makes it fail CI on FreeBSD.

If the agent requires bash you should rename it to .in and replace /bin/bash with @BASH_SHELL@ and add it in configure.ac (https://github.com/ClusterLabs/resource-agents/blob/master/configure.ac#L944)so it gets replaced on autogen/configure.

mgrzybek:
I am surprised that you mentioned FreeBSD because it looks like it is not supported, according to https://github.com/ClusterLabs/resource-agents/issues/1089.

krig:
> I am surprised that you mentioned FreeBSD because it looks like it is not supported, according to #1089.

Ah no, that was not the conclusion I wanted you to draw from that conversation :) It's just that this repository only contains the OCF standard resource agents, while you were attempting to use a non-resource agent based init script. That's a Pacemaker feature, not related to the OCF standard and the agents kept in this repository.

Regarding these agents, I'll need to take a closer look but I don't object to them if the others are OK with that.

oalbrigt:
We should move this to the openstack-resource-agents project (https://git.openstack.org/cgit/openstack/openstack-resource-agents/).

mgrzybek:
According to http://git.openstack.org/cgit/openstack/openstack-resource-agents/tree/README.rst this project aims to 

> provide highly available OpenStack services managed within Pacemaker clusters.

This repository is used to operate Openstack with high-availability in mind (target is IaaS admin)
The proposed resource agents aim to manage services provided by the API (target is IaaS consumer).

oalbrigt:
That's me.

mgrzybek:
Done !
https://blueprints.launchpad.net/openstack-resource-agents/+spec/resource-agents-consuming-openstack-api

beekhof:
Redirecting to openstack-resource-agents was my idea, but now that I see we have several Azure and AWS agents its hard to argue that the openstack ones shouldn't live here too.

Especially if this is about clusters running inside openstack VM instances.

A separate cloud section does sound like a good idea though

oalbrigt:
Thanks.

aspiers:
Only just discovered this (somehow I completely missed the blueprint - don't think launchpad ever notified me about it), but yes I agree with @beekhof and @mgrzybek that if these RAs stay here instead of [openstack-resource-agents](https://git.openstack.org/cgit/openstack/openstack-resource-agents/) (which [I'm not convinced is necessarily the best idea, but I'm willing to be convinced either way](http://lists.openstack.org/pipermail/openstack-dev/2018-October/135999.html)), then it does indeed make sense to put them under a separate `cloud` namespace instead.  Is it too late to change that?

oalbrigt:
If we put it under the `cloud` namespace I think we'll have to provide links from the `openstack` namespace, in which case we'd maybe have to add a `configure` parameter like `--without-openstack-links` for future major distribution releases that can ship without it.

What do you think @krig ?

krig:
@oalbrigt I’m just thinking that we want to move all the agents to different namespaces than heartbeat anyway, so maybe we should have a general solution that moves/renames the agents to various more appropriate locations, and have a switch to add legacy symlinks to the old locations?

oalbrigt:
That might be another way, yeah. So maybe we could change heartbeat -> generic and just have generic go to heartbeat and cloud (openstack agents) to openstack when we use `--legacy-location`.

krig:
I guess the problem is just to come up with a list of which agents should go where.

Something like

* net
* storage
* cloud
* database
* generic

mgrzybek:
There is something specific to some clouds or tools: some agents are used to make the tools high-available and others are designed to use the services provided by the tools.

This situation only applies to Openstack / OpenNebula / Proxmox… because you can build them by yourself (IaaS provider) or be the customer (IaaS user).

Maybe:
* cloud
* cloud-admin

oalbrigt:
@krig but it's actually the `provider` field, so it'll be confusing in that regard. I guess that should be cl, clusterlabs or similar sometime in the future.

krig:
@oalbrigt Well, sure. I'm not sure how much sense `cloud` makes as a provider, though? Maybe in that case it would make more sense to move the agents for each cloud provider to a unique provider, like `aws`, `gcp`, `azure`, `aliyun`, `openstack`, etc.?
For me, `clusterlabs` as a provider is kind of meaningless, I'd personally prefer to categorize the "standard" agents. But it's true that this is not the original intent of the provider field.

aspiers:
Thanks a lot guys for this discussion.  I just wanted to throw in my £0.02p and say that `cloud`, `cloud-admin`, and `openstack` all sound like good options to me - in fact anything along those lines would too :-)  Yes @mgrzybek I like your idea of separating out the two different cloud usage scenarios.  But I'll trust you guys to do a good job of the bike-shedding ;-)

oalbrigt:
@krig yeah. I think categories are a better way too. Maybe we could use `core` instead of `generic`.

kgaillot:
@oalbrigt @krig @aspiers What is the problem that reinterpreting provider is intended to solve?

If pulling in unnecessary dependencies is the problem, then we need to look at the packaging rather than the provider field. I.e. split monolithic resource-agents package into resource-agents-core, resource-agents-pgsql, etc. Anything with minimal depedencies would go in -core, anything with big dependencies would get its own package, and resource-agents (and/or resource-agents-all) would become a meta-package depending on the others. Users would have to learn to adapt their behavior (installs etc.) but I think it makes sense as the list of dependencies grows.

The initial rationale for the provider field was to allow different organizations or projects to deliver a commonly maintained set of agents, possibly overlapping. E.g. there could have been an ocf:apache:webserver and an ocf:nginx:webserver. That turned out not to be such a big deal, and the vast majority of development has stayed in the ClusterLabs resource-agents project. But it does still matter, particularly we tell people to use their own provider if they want to customize an existing agent, and that might not be as obvious with ocf:database:mysql. 

I completely agree "heartbeat" has become obsolete as a provider in any sense. I lean toward the "clusterlabs" camp (for the resource-agents project) to stick with the idea of the provider making it clear what organization/project supports the agent. 

In the particular case at hand, I think having overlapping openstack-related agents with different providers is actually in line with the original intent of providers. I agree it's confusing for users, but I still think it makes sense for the provider to indicate who maintains the agent. If the openstack-resource-agents project wants to maintain these agents too, that would be great all around.

krig:
@kgaillot My feeling is that the provider field was a failed experiment: There hasn't been many other providers, and having multiple providers provide the same agent name (like `redhat:webserver` and `suse:webserver`) only confuses things: We've already got a confusing situation with some agents that are both in `pacemaker` and `heartbeat` where the `heartbeat` agents have received bugfixes that the `pacemaker` versions haven't.

A separate issue that I have seen, that remains unsolved, is that of categorizing agents: Especially in the web interfaces it would be a huge help to have categories for the agents so that when someone wants to configure a database, they don't have to look through the whole list of hundreds of agents to find the one they need.

So my spontaneous reaction when the idea of a separate cloud provider was brought up was to hijack the provider field to provide some categorization for the agents in `resource-agents`. But I am fully aware that it's a hijack and not what the provider field was intended for.

kgaillot:
> My feeling is that the provider field was a failed experiment: There hasn't been many other providers, and having multiple providers provide the same agent name (like `redhat:webserver` and `suse:webserver`) only confuses things: We've already got a confusing situation with some agents that are both in `pacemaker` and `heartbeat` where the `heartbeat` agents have received bugfixes that the `pacemaker` versions haven't.
> 
> A separate issue that I have seen, that remains unsolved, is that of categorizing agents: Especially in the web interfaces it would be a huge help to have categories for the agents so that when someone wants to configure a database, they don't have to look through the whole list of hundreds of agents to find the one they need.

Ah, I didn't think of GUIs.
 
> So my spontaneous reaction when the idea of a separate cloud provider was brought up was to hijack the provider field to provide some categorization for the agents in `resource-agents`. But I am fully aware that it's a hijack and not what the provider field was intended for.

I just don't want to replace one source of confusion with another, e.g. if someone needs to modify the ocf:database:mysql agent locally, what do they name theirs and where does it go (without conflicting with stock packaging)?

Perhaps interestingly, the ocf:provider:agent naming scheme is pacemaker-specific. The OCF standard itself uses providers just to name file system directories.

I wonder if it would be useful to categorize agents via meta-data, e.g.

    <resource-agent name="IPaddr2" category="networking">

 GUIs would have to read all available agents' meta-data at once, not sure if that's problematic.

Another possibility might be going to ocf:provider:category/agent e.g. ocf:clusterlabs:networking/IPaddr2.

krig:
Having a category metadata attribute would be great, yeah.

I don't know for sure, but I'm a bit concerned that the last suggestion (ocf:clusterlabs:networking/IPaddr2) might make crmsh behave badly.. I have a feeling that the assumption that is baked into the tools is that the provider is the name of the directory that the agent is in. I'll need to take a look.

kgaillot:
> Having a category metadata attribute would be great, yeah.
> 
> I don't know for sure, but I'm a bit concerned that the last suggestion (ocf:clusterlabs:networking/IPaddr2) might make crmsh behave badly.. I have a feeling that the assumption that is baked into the tools is that the provider is the name of the directory that the agent is in. I'll need to take a look.

Definitely, that's part of the OCF standard. I was thinking that it would translate to /usr/lib/ocf/resource.d/clusterlabs/networking/IPaddr2 on disk (and the project would be modified to drop it there). If we're lucky, at least some applications would just tack "networking/IPaddr2" onto the end of the path and do the right thing without modification, but I'm sure some things would need changes.

jnpkrn:
Slightly orthogonal to the logical level structuring that the discussion
revolved to, as opposed to the original question of which repo is to
serve for brewing a set of agents, for which I think the answer can
become inferior once we are able to publicize tracked agents side by
side regardless of affiliations (in case developers ML is below radar):

https://lists.clusterlabs.org/pipermail/developers/2018-November/001319.html

If this is deemed a viable approach, logical structuring can, on the
contrary, only raise on importance, of course.

PR #1150, opened 2018-05-27
Medium: IPsrcaddr: Add "update_local_route" attribute

nrwahl2:
Currently, the IPsrcaddr resource agent updates the default route to
use a specified IP address (`OCF_RESKEY_ipaddress`) as the preferred
source IP. It does not make any changes to the local route table.

This patch adds an "update_local_route" boolean attribute
(default=false), which updates the local route for destination
`$OCF_RESKEY_ipaddress` to use `src $OCF_RESKEY_ipaddress`.

#### Rationale
`IPsrcaddr` is often paired with `IPaddr2` to provide a virtual IP
address and then use that virtual IP as the preferred source IP for
outgoing connections on the default route. Some applications make local
connections to the virtual IP address. A subset of those require
that the source IP for the connection be the virtual IP.

In other words, the local connection in these cases needs to be:
```
    virtual_ip -> virtual_ip
```

However, this is what is occurring:
```
    primary_ip -> virtual_ip
```

The local routing table determines the source IP for these connections.
For routes in the local routing table, the default src IP address has
always been the primary IP address of the interface. After adding an
`IPaddr2` resource, there will be a local route like this:

```
# ip route show table local to exact <virtual_ip>
local <virtual_ip> dev <iface> proto kernel scope host src <primary_ip>
```

The IPsrcaddr resource agent needs to be able to update the local route
to `OCF_RESKEY_ipaddress` in addition to updating the default route.

### DIFF
--- a/heartbeat/IPsrcaddr
+++ b/heartbeat/IPsrcaddr
@@ -58,8 +58,10 @@
 
 USAGE="usage: $0 {start|stop|status|monitor|validate-all|meta-data}";
 
-  CMDSHOW="$IP2UTIL route show   to exact 0.0.0.0/0"
-CMDCHANGE="$IP2UTIL route change to "
+        CMDSHOW="$IP2UTIL route show               to exact 0.0.0.0/0"
+      CMDCHANGE="$IP2UTIL route change             to"
+  CMDSHOW_LOCAL="$IP2UTIL route show   table local to exact $OCF_RESKEY_ipaddress"
+CMDCHANGE_LOCAL="$IP2UTIL route change table local local"
 
 SYSTYPE="`uname -s`"
 
@@ -96,6 +98,16 @@ dotted quad notation  255.255.255.0).
 </longdesc>
 <shortdesc lang="en">Netmask</shortdesc>
 <content type="string" default=""/>
+</parameter>
+
+<parameter name="update_local_route">
+<longdesc lang="en">
+If set, the route to destination \`ipaddress\` in the local route table
+will be updated to use \`ipaddress\` as the preferred source IP
+address.
+</longdesc>
+<shortdesc lang="en">Update local route</shortdesc>
+<content type="boolean" default="false"/>
 </parameter>
 </parameters>
 
@@ -159,6 +171,14 @@ srca_read() {
 	return 2
 }
 
+srca_read_local() {
+	local ROUTE="`$CMDSHOW_LOCAL`" || errorexit "command '$CMDSHOW' failed"
+	local SRCIP=`echo $ROUTE | sed -n "s/$MATCHROUTE/\3/p"`
+	[ -z "$SRCIP" ] && return 1
+	[ $SRCIP = $1 ] && return 0
+	return 2
+}
+
 #
 #	Add (or change if it already exists) the preferred source address
 #	The exit code should conform to LSB exit codes.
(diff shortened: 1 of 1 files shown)


nrwahl2:
I have added only that which is necessary for the local route update, plus replacing `ip` with `$IP2UTIL` in the `NETWORK` variable assignment. There are some other formatting, consistency, and safe parsing issues in the RA that should also be addressed, but I will keep it simple for now. I've stylistically aligned the new additions with the existing content as best I can.

oalbrigt:
add to whitelist

nrwahl2:
Related to RHBZ#1582685

heartbeat/IPsrcaddr
```diff
@@ -177,7 +197,14 @@ srca_start() {
 
 		$CMDCHANGE $ROUTE_WO_SRC src $1 || \
```
oalbrigt: Wouldnt it be either this command orthe one in the if-statement?

nrwahl2: The `if` statement only includes an `ocf_log_info`. Maybe you meant either this command or the `ip route replace` before it.

```
175                 ip route replace $NETWORK dev $INTERFACE src $1 || \
176                         errorexit "command 'ip route replace $NETWORK dev $INTERFACE src $1' failed"
177 
178                 $CMDCHANGE $ROUTE_WO_SRC src $1 || \
179                         errorexit "command '$CMDCHANGE $ROUTE_WO_SRC src $1' failed"
```

I'm not sure what the effective difference is between lines 175 and 178 in the original version. I've left the existing stuff alone and just added what's needed for `update_local_route=true` to work.

heartbeat/IPsrcaddr
```diff
@@ -177,7 +197,14 @@ srca_start() {
...
 			errorexit "command '$CMDCHANGE $ROUTE_WO_SRC src $1' failed"
+
 		rc=$?
```
oalbrigt: You might want to return early if it fails here.

heartbeat/IPsrcaddr
oalbrigt: You're welcome to update to $IP2UTIL.

heartbeat/IPsrcaddr
oalbrigt: You should return early if it fails here.

heartbeat/IPsrcaddr
oalbrigt: You're welcome to update it to $IP2UTIL.

heartbeat/IPsrcaddr
nrwahl2: The `errorexit` does that. Custom function that the creator of this RA wrote.

```
errorexit() {
        ocf_exit_reason "$*"
        exit $OCF_ERR_GENERIC
}
```

heartbeat/IPsrcaddr
nrwahl2: Will do!

nrwahl2:
This is going to need some more attention, to add a flag for `update_default_route` so that users can configure multiple local routes.

heartbeat/IPsrcaddr
oalbrigt: Ah. I didnt think of that :)

nrwahl2:
Closed in favor of https://github.com/ClusterLabs/resource-agents/pull/1212.

PR #1151, opened 2018-06-01
Volume groups and logical volumes with "-" in their name get mangled with …

paulmezz:
…double dashes

in dmsetup.  Switching to wc and just counting lines while depending on the vgname
+ lvname filter in the dmsetup call gets around the issue with dmsetup outputting
correctly but grep failing due to the name mangle.

Logic for both test cases and dmsetup calls changed so they match too.  No reason
to not have matching tests.

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -692,20 +692,27 @@ tagging_deactivate() {
 # lvs/vgs when the metadata is somehow inconsistent.
 #
 # So, we have to make compromise that the VG is assumably active if any LV of the VG is active.
+#
+# Paul:
+# VGS + LVS with "-" in their name get mangled with double dashes in dmsetup.
+# Switching to wc and just counting lines while depending on the vgname + lvname filter
+# in dmsetup gets around the issue with dmsetup reporting correctly but grep failing.
+#
+# Logic for both test cases and dmsetup calls changed so they match too.
+#
+# This is AllBad but there isn't a better way that I'm aware of yet.
 lvm_status() {
 	local dm_count
 
 	if [ -n "${LV}" ]; then
 		# dmsetup ls? It cannot accept device name. It's
 		# too heavy to list all DM devices.
-		dmsetup info --noheadings --noflush -c -S "vgname=${VG} && lvname=${LV}" \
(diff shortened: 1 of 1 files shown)


heartbeat/LVM-activate
```diff
@@ -692,19 +692,28 @@ tagging_deactivate() {
...
+		dm_count=$(dmsetup info --noheadings --noflush -c -S "vgname=${VG} && lvname=${LV}" |  wc -l )
 	else
-		dm_count=$(dmsetup --noheadings info -c -S "vgname=${VG}" 2>/dev/null | grep -c "${VG}-")
-		test $dm_count -gt 0
+		dm_count=$(dmsetup info --noheadings --noflush -c -S "vgname=${VG}" 2>/dev/null | wc -l )
 	fi
 
+	test $dm_count -gt 0
+
 	if [ $? -ne 0 ]; then
```
oalbrigt: Can you update this to "if [ $dm_count -gt 0 ]; then" and remove the test-line above?

heartbeat/LVM-activate
```diff
@@ -692,20 +692,27 @@ tagging_deactivate() {
...
-			| grep -Eq "${VG}-+${LV}"
+		dm_count=$(dmsetup info --noheadings --noflush -c -S "vgname=${VG} && lvname=${LV}" |  wc -l )
 	else
-		dm_count=$(dmsetup --noheadings info -c -S "vgname=${VG}" 2>/dev/null | grep -c "${VG}-")
-		test $dm_count -gt 0
+		dm_count=$(dmsetup info --noheadings --noflush -c -S "vgname=${VG}" 2>/dev/null | wc -l )
 	fi
-
-	if [ $? -ne 0 ]; then
+
```
oalbrigt: Can you remove this empty line too? Btw you can squash the commits, so we dont get all the minor mistakes in the log.

http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html

oalbrigt: No worries. We all do mistakes :)

oalbrigt:
Thanks.

PR #1155, opened 2018-06-10
SAPInstance improvements 2018/05 (2/4) - make missing binaries on Slave non-fatal issue

OndrejHome:
make exit in abnormal_end() function not 'fatal' but only 'hard' so this
affects only node on which the issue occurred and not all nodes.
This should allow the node running Master to survive failure of resource
on node running Slave when it reaches the abnormal_end() function.

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -301,7 +301,7 @@ abnormal_end() {
   fi
 
   ocf_log err $err_msg
-  exit $OCF_ERR_CONFIGURED
+  exit $OCF_ERR_INSTALLED
 }
 
 #


OndrejHome:
I would like to get into some technical objections to the proposed change here before turning to all possible deployments that are out there. I'm not familiar with what kind of change would trigger need for recertification of the RA so please let me know at least the scope of such change so I can see if it can be avoided or not. Same goes for any additional tests needed for this change as the upstream repository doesn't contain any hints on what else can be done to make it easier for proving that changes are doing what they are designed to do and not breaking old behaviour.

I have also checked further into this change and I noticed a possible place where its impact can be further limited to achieve the goal mentioned in the title of this PR and limit effect to other deployments.

There are technically 3 places where the `abnormal_end` function is called. Below is short description on their location in the code and explanation on what they affect (in sense if they affect the Master/Slave resource only, Standalone resource only or Both)

--------

### Scenario 1 
https://github.com/ClusterLabs/resource-agents/blob/41961bebac65c00e3417ce39f4e9b0c46ce30c25/heartbeat/SAPInstance#L347
- we have not detected the `$DIR_EXECUTABLE` in function `sapinstance_init()`
- `sapinstance_init()` is called only in functions with suffix `_clone`  so it affect **Master/Slave resource only**

### Scenario 2 
https://github.com/ClusterLabs/resource-agents/blob/41961bebac65c00e3417ce39f4e9b0c46ce30c25/heartbeat/SAPInstance#L442
- `/usr/sap/$SID/SYS/profile/` is not a directory
- this resides inside of the `check_sapstartsrv()` function that is within functions used by **both Master/Slave and Standalone** instances

### Scenario 3 
https://github.com/ClusterLabs/resource-agents/blob/41961bebac65c00e3417ce39f4e9b0c46ce30c25/heartbeat/SAPInstance#L445
- `$SAPSTARTPROFILE` doesn't look readable
- same impact as the 'Scenario 2' as this is inside of the `check_sapstartsrv()` function, so affects **both Master/Slave and Standalone** instances

----------

While I believe that there are benefits for getting this change into both, Standalone and Master/Slave, instances it is also possible to limit it. In overall the impact of the change proposed here should be similar to situation in which one of the nodes has failed and cluster has stopped resources running on that node (leaving other nodes operations if there are no other problems).

Below is example code that will restrict change in this PR to only Master/Slave instances of RA - basically only to 'Scenario 1'  mentioned above.
~~~
if [ ${CLONE} -eq 1 ]
then
  exit $OCF_ERR_INSTALLED
else
  exit $OCF_ERR_CONFIGURED
fi
~~~

fmherschel:
We in the LinuxLab like to reject this change till we have the chance to test that tiny change against the
SAP certification test procedures. Please stay patient. If the tests will be successful we could re-add such change later and make it available with future major releases of SAP specific Linux enterprise products.

fdanapfel:
After reviewing the patch again and the comments here I think this should be rejected for now for the following reasons:

- the original intention of the patch was to fix situations where the SAP instance binaries are missing, however as the comments above show the patch will also affect situations where the SAP instance profiles (which decribe how the instance should be configured) go missing; since the sap instance profiles are normally symlinked from the central /sapmnt directory, it indicates that there is a larger issue with the setup and we absolutely want to make sure that the system is brought down in such situations and the admins are alerted

- the change is done in a function called 'abnormal_end' which as the name implies should only be called if something happened that should never happen during normal operations and therefore stopping the whole environment and raising alarms to the admins makes sense here

Please also be aware that since SAP is moving to a new setup for running the Enqueue and Enqueue Replication servers the master/slave setup for manging the (A)SCS/ERS instances in a pacemaker cluster will no longer be needed in the future, therefore the current way the return code is set will most likely not have as much of an impact in the future anymore.

OndrejHome:

PR #1156, opened 2018-06-10
SAPInstance improvements 2018/05 (3/4) - detect binaries in M/S resource from both ASCS/ERS instance

OndrejHome:
When we have $OCF_RESKEY_ERS_InstanceName defined then we can use the
binaries from both ASCS and ERS instance.
This gives us one more place where we can autodetect the needed
binaries before giving up.

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -316,6 +316,11 @@ sapinstance_init() {
   InstanceName=`echo "$myInstanceName" | cut -d_ -f2`
   InstanceNr=`echo "$InstanceName" | sed 's/.*\([0-9][0-9]\)$/\1/'`
   SAPVIRHOST=`echo "$myInstanceName" | cut -d_ -f3`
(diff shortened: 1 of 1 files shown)


OndrejHome:
Isn't then attempt in code trying two locations considered a risk here? (I mean one from `/usr/sap/$SID/$InstanceName/exe` and another from `/usr/sap/$SID/SYS/exe/run` or it is safe to use these two locations interchangeably)

If we should not use binaries from ERS/ASCS interchangeably then this PR can be dropped.

fmherschel:
Using first /usr/sap/$SID/$InstanceName/exe is the best location (best effort). Using /usr/sap/$SID/SYS/exe/run is allowed, because it might match a situation where the kernel (again) needs to be copied. This directory contains the 'designated' binaries. Once a SAP instance is started the contents (if different) is copied to the instance specific directory.

Using instance specific directories crosswise is not acceptable, because there no guarantee that these binaries are matching the running SAP instance.

PR #1159, opened 2018-06-14
Medium: pgsql: Fix incorrect SQL is selected with PostgreSQL 11 or later

ytakeshita:
When  the return value of `ocf_version_cmp()` is `2` the right `test` command does not match because the  `$?` is replaced to the result of left test  command.
As a result incorrect SQL statement (i.e. pg_last_xlog*) is selected and fail to promote with PostgreSQL 11 or later.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1954,7 +1954,8 @@ pgsql_validate_all() {
 
     if use_replication_slot; then
         ocf_version_cmp "$version" "9.4"
-        if [ $? -eq 0 -o $? -eq 3 ]; then
+        rc=$?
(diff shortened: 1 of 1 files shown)


oalbrigt:
-o and -a should be avoided (http://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o), so I think assigning $? to a variable and check that instead is a better way to do this.

oalbrigt:
add to whitelist

oalbrigt:
retest this please

ytakeshita:
@oalbrigt 
update PR to avoid -o, and modify other part of using `-o` as well. What should I do for retest ?
This patch is working as expected in my test.

oalbrigt:
Thanks.

PR #1160, opened 2018-06-14
Freebsd and container agents

mgrzybek:
This PR adds 5 agents:
* jail: manages jails
* jail-info: adds node attributes (number of running jails)
* jailed-service: manages a service within a container (jail, lxc, lxd)
* IPalias: manages IP addresses like IPaddr2
* IPjail: manages IP addresses using iocage


krig:
Please add the new agents to the documentation Makefile as well `doc/man/Makefile.am`.

Other than that, it looks good to me.

krig:
Oh, one thing: `IPalias` and `IPjail` should be named `ip-alias` and `ip-jail`: We've set a naming standard for new agents, we just haven't forced all the existing agents to conform yet. There's a standard proposal for name aliasing which will allow us to actually rename the existing agents, but it's not there yet unfortunately.

krig:
Heh, another thought: What's the difference between the existing `IPaddr` and this `IPalias` agent? I mean, `IPaddr` is meant to be the BSD-compatible variant of `IPaddr2`.

mgrzybek:
The agents are renamed. Moreover I'v never read `IPaddr` agent before. `IPalias` is a duplicate indeed.

heartbeat/ip-jail
```diff
@@ -0,0 +1,209 @@
...
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+#######################################################################
+
+meta_data()
+{
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="IPjail">
```
oalbrigt: The name needs to be updated in the metadata.

heartbeat/jailed-service
```diff
@@ -0,0 +1,223 @@
...
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+#######################################################################
+
+meta_data()
+{
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="jailed_service">
```
oalbrigt: The name has to match the filename here too.

heartbeat/ip-jail
```diff
@@ -0,0 +1,209 @@
+#!/usr/bin/env bash
```
oalbrigt: We prefer /bin/sh. If you only need bash for =~ below you could replace it with case and *:* (example: https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/findif.sh#L234-L241).

If not there's the "@BASH_SHELL@" variable you could use, and rename it to ip-jail.in and add it to configure.ac so it gets updated when you run ./configure.

oalbrigt:
Can you undo the merge commit and do a rebase instead? We prefer not having additional merge commits in the git history.

PR #1164, opened 2018-07-02
SAPDatabase: Add support for LiveCache instances with SAPHostAgent. (1/2)

OndrejHome:
Hi,
This PR adds ability to specify attribute `LIVECACHE_INSTANCE` that causes the `StopDatabase` function to not use the `-service` option which causes the 'x_server is not active' fail when attempting to stop it with it when LiveCache instance is used. Additionally it adds documentation about requirements for this setup that have further explanation in SAP Note 1927600. The LIVECACHE_INSTANCE is by default set to `false` which means that there is no impact on current deployments and can be activated only when `DBTYPE` is `ADA`.

### DIFF
--- a/heartbeat/SAPDatabase
+++ b/heartbeat/SAPDatabase
@@ -83,6 +83,7 @@ The resource agent supports the following databases:
 - SAP-DB / MaxDB 7.x
 - Sybase ASE 15.7
 - SAP HANA Database since 1.00 - with SAP note 1625203 (http://sdn.sap.com)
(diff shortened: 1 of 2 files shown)


OndrejHome:

OndrejHome:
I have no longer access to testing systems to do any further evaluation on this. If this is of a concern then feel free to take over or close this PR without merging.
If there is no response here I will close this next month (September) to not have here long-time hanging PR without progress.

OndrejHome:
Closing due to no response from SAPDatabase maintainers for extended period of time.

PR #1165, opened 2018-07-02
SAPDatabase: Add ability to specify extra arguments for SAPHostAgent functions (2/2)

OndrejHome:
Hi, 
Currently for functions StartDatabase/StopDatabase/GetDatabaseStatus
so this can support LiveCache instances that require additional
arguments to be passed to SAPHostAgent in order to connect properly
to LiveCache instance as described in the SAP Note 1927600.

Additional note: I would like to also have ability to hide some of the arguments  - for example by providing them via file/script instead of directly specifying them in resource agent. This is not part of this PR but a questions that I would like to incorporate here for future changes. Would that make sense?

### DIFF
--- a/heartbeat/SAPDatabase
+++ b/heartbeat/SAPDatabase
@@ -186,6 +186,11 @@ Usually you can leave this empty. Then the default: /usr/sap/hostctrl/exe is use
   <shortdesc lang="en">path to a post-start script</shortdesc>
   <content type="string" default="" />
  </parameter>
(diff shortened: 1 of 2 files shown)


OndrejHome:

OndrejHome:
I have no longer access to testing systems to do any further evaluation on this. If this is of a concern then feel free to take over or close this PR without merging.
If there is no response here I will close this next month (September) to not have here long-time hanging PR without progress.

OndrejHome:
Closing due to no response from SAPDatabase maintainers for extended period of time.

PR #1173, opened 2018-07-17
Do not call ocf_attribute_target in the meta-data function

mbaldessari:
Starting with pacemaker-1.1.19 a "crm_node -n" call triggers
a CRM_OP_NODE_INFO cluster operation. If this is called
from a bundle with 1.1.19 code (or later) running on a 1.1.18
cluster, during a meta-data call we will get the following error in the
cluster logs:
Jul 14 11:35:27 [20041] controller-0 crmd: error: handle_request: Unexpected request (node-info) sent to non-DC node
Jul 14 11:35:27 [20041] controller-0 crmd: error: handle_request: Unexpected <create_request_adv origin="send_node_info_request" t="crmd" version="3.0.14" subt="request" reference="node-info-crm_node-1531568127-1" crm_task="node-info" crm_sys_to="crmd" crm_sys_from="e67698a2-6f50-45fc-b8de-423c94e11c99" acl_target="root" crm_user="root" src="controller-0"/>

By not calling ocf_attribute_target (which triggers a crm_node -n
call) when polling for meta-data we do not run into this issue.

This can easily get triggered when creating a resource invoking
crm_node -n inside a 1.1.19 container with pcs, as that invokes
the 'meta-data' action explicitely.

Suggested-By: Ken Gaillot <kgaillot@redhat.com>

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -66,9 +66,11 @@
 
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
-. ${OCF_FUNCTIONS_DIR}/mysql-common.sh
 
-NODENAME=$(ocf_attribute_target)
+if [ "$__OCF_ACTION" != "meta-data" ]; then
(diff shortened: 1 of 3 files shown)


oalbrigt:
LGTM. Thanks.

PR #1176, opened 2018-07-19
gcp-vpc-move-vip.in: filter call to aggregatedList

helen-fornazier:
Don't list all the instances in the project, filter only the one we are
interested in.


oalbrigt:
LGTM. Thanks.

PR #1186, opened 2018-08-02
SAPInstance: Improved SAP instance profile detection

fmherschel:
SAPInstance has been initially written for NetWeaver 6.x and 7.0. In these times SAP instances have had a START_xx profile and a <SID>_xx parameter profile. Beginning with SAP NetWeaver 7.1 or 7.2 SAP removed the START_xx profile and added the content to the <SID>_xx parameter profile.
Since that change the automated detection of the start-profile failed and each cluster needed to use
the parameter setting for START_PROFILE. 
The given patch re-implements the automated profile detection. While the START_xx profile is still preferred to be backward compatible, SAP installations without an outdated START_xx profile could
now be detected via the <SID>_xx profile.

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -371,7 +371,11 @@ sapinstance_init() {
 
   if [ -z "$currentSTART_PROFILE" ]
   then
-    SAPSTARTPROFILE="$DIR_PROFILE/START_${InstanceName}_${SAPVIRHOST}"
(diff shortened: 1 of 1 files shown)


pirat013:
@branch master,

I've implemented the patch and verified them extensively. I see a great improvement in this patch to simply fie the cluster implementation.  
+1 to make the setup as simple as possible.

Bernd

krig:
Looks good to me. Thanks!

PR #1187, opened 2018-08-06
RFC: aws-vpc-move-ip: Use ip utility to check address

Werkov:
When pinging the assigned address during initial monitor (probe) on one
node we may actually ping the reachable address when the resource is
running on another node. This yields false positive monitor result on
the pinging node. Avoid this by merely checking the assignment of the
address to an interface.

The patch is not tested, this is a RFC.


krig:
Make sense to me.

oalbrigt:
Looks good to me.

Werkov:
Despite it makes sense and looks good, there was a bug. See #1203.

PR #1188, opened 2018-08-09
rabbitmq-cluster: retry start when cluster join fails

dciabrin:
When a node tries to join an existing cluster, it fetches a node
list to try to connect from any of those running nodes.

If the nodes from this list become unavailable while we're joining
the cluster, the rabbitmq server will fail to get clustered and
make the start operation fail.

Give the resource a chance to start anyway by retrying the entire
start actions until it succeeds or until the start timeout is
reached and pacemaker stops the start operation.

Co-Authored-by: <michele@acksyn.org>
Suggested-by: <abeekhof@redhat.com>

### DIFF
--- a/heartbeat/rabbitmq-cluster
+++ b/heartbeat/rabbitmq-cluster
@@ -31,6 +31,12 @@
 
 #######################################################################
 
+# This arbitrary value here is used by the rmq_start action to
+# signify that the resource agent must retry the start process
+# It might potentially conflict with OCF assigned error code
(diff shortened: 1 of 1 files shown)


dciabrin:
@jeckersb @lemenkov @mbaldessari @beekhof Let's discuss that PR here, I can't add reviewers in the PR directly.

beekhof:
Ship it!

krig:
Looks good to me too.

PR #1189, opened 2018-08-13
ipsec: Add tunnel fallback option

JAORMX:
This option makes brings up a fallback tunnel if the main tracked tunnel
goes down. This is useful for when there is a failover in the VIP (which
is our main tunnel), and then we need to establish a tunnel to the new
node that has the VIP.

The fallback mechanism is a little bit hacky, but it's the only way I
could get it to work... basically it's put up in a background process,
which will keep trying retry putting up the fallback node. Meanwhile,
the resource agent will keep doing its job and put up the main node on
the host that received the VIP. eventually, a retry on the original node
will bring up the fallback tunnel.

### DIFF
--- a/heartbeat/ipsec
+++ b/heartbeat/ipsec
@@ -109,7 +109,7 @@ END
 
 ipsec_start() {
 	ipsec auto --add "${OCF_RESKEY_tunnel}"
-	ipsec whack --listen &>> /tmp/ipsec-agent.log
+	ipsec whack --listen
 	local return_code=$?
 	if [ $return_code -eq  1 -o $return_code -eq 10 ]; then
 		ocf_log warn "${OCF_RESOURCE_INSTANCE} : Unable to add tunnel ${OCF_RESKEY_tunnel} with return code ${return_code}"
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

heartbeat/ipsec
```diff
@@ -116,6 +123,16 @@ ipsec_stop() {
 	ipsec auto --down "${OCF_RESKEY_tunnel}"
 	local return_code=$?
 	ocf_log info "${OCF_RESOURCE_INSTANCE} : Put down tunnel ${OCF_RESKEY_tunnel} with return code ${return_code}"
+	ipsec whack --listen &>> /tmp/ipsec-agent.log
```
krig: Writing to a random file in `/tmp` doesn't seem very clean to me... I'm not sure if it ought to be a file in `$HA_RSCTMP` or what, maybe it ought to be a configurable parameter?

oalbrigt: Seems like it's the same in the start-action, so I guess we should improve it there as well.

JAORMX: I was using that for debugging. Let me remove that.

JAORMX:
@krig @oalbrigt thanks for the review. Fixed.

oalbrigt:
Thanks.

PR #1198, opened 2018-08-27
Opentack agents updates

mgrzybek:
This PR:
* provides a better layout
* uses `crm_node -n` result instead of `$HOSTNAME`
* deals with API v3 (tenant / project)

### DIFF
--- a/heartbeat/openstack-cinder-volume
+++ b/heartbeat/openstack-cinder-volume
@@ -37,6 +37,8 @@
 # Defaults
 OCF_RESKEY_openstackcli_default="/usr/bin/openstack"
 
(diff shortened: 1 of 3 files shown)


diegoakechi:
Hi @oalbrigt,
I would start to review this PR, but I notice that it is failing on the Kronosnet CI. As I don't have access to check why is it failing, I could not do much to help to fix it. Can you, please, take a look at the reasons for the failure?

oalbrigt:
It fails on Debian testing with newer gcc, which has been fixed already, so it can be ignored.

@mgrzybek could fix it by doing a "git rebase upstream/master" followed by a "git push -f".

PR #1199, opened 2018-08-27
Consul agents

mgrzybek:
Theses agents get data from Consul and create services:
* consul-info: number of datacenters (useful for split-brain detection), list of passing and failing services, domain name, datacenter name
* consul-catalog: creates and removes a service from Consul catalog. This can be used instead of a floating IP.


krig:
Hey @mgrzybek, I see that you are using Python inside your agent. As of recently, we have added a Python library intended to make it easier to write the whole agent in Python, making sure that it's OCF compatible, that the Python interpreter is correctly located on the current platform, etc. So I would recommend writing the whole agent in Python in this case, since you want to use Python for certain things anyway.

You can look at `doc/dev-guides/writing-python-agents.md` or one of the existing Python agents like `heartbeat/gcp-pd-move.in` for more details.

PR #1200, opened 2018-08-27
Xen: Set utilization for cpu and hv_memory

vvidic:
Set resource utilization attributes in the monitor
action like it is done for KVM in VirtualDomain agent.

### DIFF
--- a/heartbeat/Xen
+++ b/heartbeat/Xen
@@ -275,16 +275,15 @@ set_util_attr() {
 
 	cval=$(crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>/dev/null)
 	if [ $? -ne 0 ] && [ -z "$cval" ]; then
-		crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>&1 | grep -e "not connected" > /dev/null 2>&1
-		if [ $? -eq 0 ]; then
-			ocf_log debug "Unable to set utilization attribute, cib is not available"
+		if crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>&1 | grep -q "not connected"; then
+			ocf_log debug "Unable to get utilization attribute $attr: cib is not available"
 			return
 		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


heartbeat/Xen
```diff
@@ -248,6 +269,45 @@ Xen_Status_with_Retry() {
...
 }
 
+set_util_attr() {
+	local attr=$1 val=$2
+	local cval outp
+
+	cval=$(crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>/dev/null)
+	if [ $? -ne 0 ] && [ -z "$cval" ]; then
+		crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>&1 | grep -e "not connected" > /dev/null 2>&1
+		if [ $? -eq 0 ]; then
```
diegoakechi: Just a side note, I think that using a different construction here could improve the readability and maintainability of the code, as the return code 0 is usually used to handle success and in that case is being used to check a failure.

vvidic: ``set_util_attr`` function is taken directly from ``VirtualDomain`` agent but I can try to make it simpler.

vvidic: Ah, it seems the test is made this way for the case when the attribute is not set: exit code will be 1 and ``$cval`` empty. The second call to ``crm_resource`` than checks for possible not connected error.

Don't know if there is a better way to check for unset utilization attribute?

diegoakechi: As you test the grep RC, you can inverse it to give you an RC 1 in case it matches the "not connected" string, for example:

``` bash
crm_resource -Q -r $OCF_RESOURCE_INSTANCE -z -g $attr 2>&1 | grep -v -e "not connected" > /dev/null 2>&1
		if [ $? -eq 1 ]; then
```
So it makes easier for another developer understand that in that case, we had an error on the call.

diegoakechi: Ok, looks better for me.

vvidic: Updated the ``set_util_attr`` with this change now.

oalbrigt:
Looks good to me. Thanks.

PR #1202, opened 2018-08-29
rabbitmq-cluster: get cluster status from mnesia during monitor

jeckersb:
If mnesia is not running (for example if `rabbitmqctl stop_app` has
been called, or the service has paused during partition due to the
pause_minority strategy) then the cluster_status command to
rabbitmqctl will read the cached cluster status from disk and the
command returns 0 even though the service isn't really running at all.

Instead, force the cluster status to be read from mnesia.  If mnesia
is not running due to the above or similar circumstances, the command
will catch that and properly fail the monitor action.

Resolves: RHBZ#1595753

### DIFF
--- a/heartbeat/rabbitmq-cluster
+++ b/heartbeat/rabbitmq-cluster
@@ -181,26 +181,16 @@ remove_pid () {
 rmq_monitor() {
 	local rc
 
-	$RMQ_CTL cluster_status > /dev/null 2>&1
-	rc=$?
-	case "$rc" in
-	0)
+	if $RMQ_CTL eval 'rabbit_mnesia:cluster_status_from_mnesia().' | grep -q '^{ok'; then
 		ocf_log debug "RabbitMQ server is running normally"
 		rmq_write_nodename
-		
+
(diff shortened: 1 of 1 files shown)


jeckersb:
@lemenkov @mbaldessari

heartbeat/rabbitmq-cluster
```diff
@@ -181,26 +181,16 @@ remove_pid () {
 rmq_monitor() {
 	local rc
 
-	$RMQ_CTL cluster_status > /dev/null 2>&1
-	rc=$?
-	case "$rc" in
-	0)
+	if $RMQ_CTL eval 'rabbit_mnesia:cluster_status_from_mnesia().' | grep -q '^{ok'; then
```
mbaldessari: This should be fine I think. Another alternative seems to be to call 'rabbit_misc:which_applications().' and then check for the rabbit app, but from my small understanding this seems more solid.

oalbrigt:
add to whitelist

oalbrigt:
LGTM. Thanks.

PR #1203, opened 2018-08-30
aws-vpc-move-ip: Fix broken shell quoting

Werkov:
The argument 4th to `ip` is passed with single quotes around which
cannot be parsed as valid IP address. Furthermore, we need to expand the
$OCF_RESKEY_ip for grep. This breaks correct detection of the assigned
address.

Fixes 7632a85bcf642b484df52a25dbffbfa0031421bc.


Werkov:
This is tested to successfully start the resource against a dummy AWS API.

krig:
Thank you!

PR #1206, opened 2018-08-31
aws-vpc-move-ip: Fixes

diegoakechi:
1. Check the routing table also during the monitor probe action
2. Force text output during the awscli call
3. Includes support for the address parameter with backward compatibility purposes.


oalbrigt:
ok to test

PR #1207, opened 2018-08-31
pgsql: Avoid the change of /dev/null to postgres owner/group

diegoakechi:
The check_log_file performs a -f test on the log file which is set to
/dev/null by default and it returns 1 for non-regular files.


oalbrigt:
Makes sense to me.

oalbrigt:
add to whitelist

PR #1208, opened 2018-09-03
Filesystem: Canonicalize mountpoint symlinks

nrwahl2:
Commit 2aa8015 added support to `Filesystem_status()` for mountpoints
that are symlinks. However, it missed two other places where `readlink`
calls should have been added to canonicalize symlinks.


krig:
Looks good to me.

oalbrigt:
Thanks.

PR #1211, opened 2018-09-05
Try to enable use of ss wherever netstat is used

krig:
So netstat has been deprecated on Linux for a long time, but we still have a lot of agents that need it.

It should be possible to replace any use of netstat with ss.

These commits don't fix all of the uses, but a few.


oalbrigt:
LGTM.

krig:
Thanks.

PR #1212, opened 2018-09-06
Medium: IPsrcaddr: Add "update_local_route" attribute

nrwahl2:
Currently, the IPsrcaddr resource agent updates the default route to
use a specified IP address (`OCF_RESKEY_ipaddress`) as the preferred
source IP. It does not make any changes to the local route table.

This patch adds an "update_local_route" boolean attribute
(default=false), which updates the local route for destination
`$OCF_RESKEY_ipaddress` to use `src $OCF_RESKEY_ipaddress`.

It also adds an "update_default_route" boolean attribute
(default=true), which updates the default route to use `src
$OCF_RESKEY_ipaddress`. By disabling this attribute, the user can
create multiple `IPsrcaddr` resources that update local routes without
updating the default route. Only one `IPsrcaddr` resource at a time
can manage the default route.

#### Rationale:
`IPsrcaddr` is often paired with `IPaddr2` to provide a virtual IP
address and then use that virtual IP as the preferred source IP for
outgoing connections on the default route. Some applications make local
connections to the virtual IP address. A subset of those require
that the source IP for the connection be the virtual IP.

In other words, the local connection in these cases needs to be:
    virtual_ip -> virtual_ip

However, this is what is occurring:
    primary_ip -> virtual_ip

The local routing table determines the source IP for these connections.
For routes in the local routing table, the default src IP address has
always been the primary IP address of the interface. After adding an
`IPaddr2` resource, there will be a local route like this:

```
local <virtual_ip> dev <iface> proto kernel scope host src <primary_ip>
```

The IPsrcaddr resource agent needs to be able to update the local route
to `OCF_RESKEY_ipaddress` in addition to updating the default route.

### DIFF
--- a/heartbeat/IPsrcaddr
+++ b/heartbeat/IPsrcaddr
@@ -17,7 +17,11 @@
 #	that such packets carry the same IP irrespective of which host in
 #	a redundant cluster they actually originate from.
 #
-#	It can add a preferred source address, or remove one.
+#	Optionally, with the update_local_route attribute enabled, this
+#	script can update the local route table to use `ipaddress` as the
+#	preferred source address when `ipaddress` is also the destination.
+#
+#	This script can add a preferred source address, or remove one.
 #
 #	usage: IPsrcaddr {start|stop|status|monitor|validate-all|meta-data}
 #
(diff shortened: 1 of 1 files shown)


nrwahl2:
I closed PR#1150 in favor of this one due to some git branch issues.

As a wise man once said,  /*You are not expected to understand this.*/ This resource agent, written long ago, does not adhere to any of our current conventions. You may scratch your head at some of it. 

I tried to change only that which is necessary to make the new features work, and leave the rest of the implementation alone. This still required numerous additions and changes.

Test cases:
https://gist.github.com/nrwahl2/161e4e09c535aec04e0dbecfa0a050af

nrwahl2:
Closed in favor of #1212.

PR #1214, opened 2018-09-14
Add gcp-pd-move python script

gut:


gut:
I'm not really sure how to test the CSEK key. It looks to me that the file given should simply be read and handed over to google API but if somebody has some directions. I can gladly test.

A refactor of the python scripts scripts is something I'm willing to do next, as there is some common code copy-pasted.

gut:
I'm not really sure how to test the CSEK key. It looks to me that the file given should simply be read and handed over to google API but if somebody has some directions. I can gladly test.

A refactor of the python scripts scripts is something I'm willing to do next, as there is some common code copy-pasted.

krig:
Looks good to me.

oalbrigt:
HOSTNAME and hostname should be changed to instacename equivalents.

gut:
done

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1215, opened 2018-09-17
Use ${HA_SBIN_DIR} on all the command invokation

mbaldessari:
It is present on all commands except a single 'crm_mon' invokation.


oalbrigt:
Thanks.

PR #1221, opened 2018-09-21
Xen: disable utilization autoset by default

vvidic:
If node utilization is also not set, resource might not
be able to start anywhere.


krig:
LGTM too.

PR #1223, opened 2018-09-26
LVM-activate: fixes

oalbrigt:
- read parameters for start/stop/monitor-actions
- fail during monitor-action when run with incorrect access_mode


krig:
Looks good to me.

PR #1224, opened 2018-09-27
nfsserver: mount rpc_pipefs

oalbrigt:
Without this rpc_pipefs gets unmounted during the stop-action, causing the next start to fail.


krig:
That's a really odd problem, I don't think we've seen any issue with restarting the nfsserver resource as far as I know.

oalbrigt:
Closing in favor of updated PR:
https://github.com/ClusterLabs/resource-agents/pull/810

oalbrigt:
Yeah. It's not happening on initial start because then it's already available through the systemd service unless you disabled that, so it only happens after you've stopped the service and started it again.

PR #1229, opened 2018-10-01
Dev: ocf.py: Flesh out python library for resource agents

krig:
This PR adds various helper functions (like `is_probe` and `ocf_exit_reason`) as well as a metadata generator and run loop.

It also adds a document describing how to use the library.

Usage example:

``` python
import ocf

def start_action(argument):
    print("The start action receives the argument as a parameter: {}".format(argument))


def main():
    metadata = ocf.Metadata("example-agent",
                            shortdesc="This is an example agent",
                            longdesc="An example of how to " +
                            "write an agent in Python using the ocf " +
                            "Python library.")
    metadata.parameter("argument",
                       shortdesc="Example argument",
                       longdesc="This argument is just an example.",
                       content_type="string",
                       default="foobar")
    metadata.action("start", timeout=60)
    ocf.run(metadata,
            handlers={
                "start": start_action
            })

if __name__ == "__main__":
    main()
```


oalbrigt:
No need to rewrite the indentation, and it just makes it harder to track the changes.

krig:
Leaving the indentation as tabs confuses my editor, and it's not according to PEP 8 :(
But I can revert that part.

krig:
Alright, removed the indentation and whitespace changes.

heartbeat/ocf.py
```diff
@@ -135,3 +135,256 @@ def emit(self, record):
...
+		action = os.environ.get("__OCF_ACTION")
+	if action is None:
+		ocf_exit_reason("No action argument set")
+		sys.exit(OCF_ERR_UNIMPLEMENTED)
+	if action in ('meta-data', 'usage', 'methods'):
+		sys.stdout.write(str(metadata) + "\n")
+		sys.exit(OCF_SUCCESS)
+
+	check_required_params()
+	for handler, func in handlers.items():
```
oalbrigt: I think it'd be smarter to do "if action in handlers:" instead of the for-loop.

krig: yeah, I'll change it.

krig:
I updated the commit and also added a `OCF_ACTION` variable set to `$__OCF_ACTION` if set or else if only one command line argument is provided, to that command line argument.

I'm going to dig into how `ocft` works to see if there is anything needed to make python agents compatible with it as well.

oalbrigt:
Sounds good.

krig:
Everything seems ocft-compatible and OCF-compliant as far as I can tell, yeah.

krig:
Yeah, I considered that. I’ll change that.

oalbrigt:
Perfect. Thanks.

krig:
Thanks!

doc/writing-python-agents.md
```diff
@@ -0,0 +1,90 @@
+# Resource Agent guide for Python
```
oalbrigt: @krig I think we should move this under dev-guides.

krig: Sure, I'll make a PR.

krig:
Done: #1242

PR #1233, opened 2018-10-04
Dev: ocf.py: Improved exception handling

krig:
Catch exceptions thrown from the handler functions and convert them into a reasonable exit code.
Also set the exit reason if it hasn't been set already, else just log the error.

Also update the doc example to include the imports needed for the ocf import.

edit: Added some more cleanups / niceties as well. Printing usage on `--help`, passing the handler to `add_action()`...

edit again: OK, after doing the above, renaming the Metadata class to Agent made more sense. Feedback on that is welcome :)


krig:
As a test I converted the Dummy agent to Python, see this gist:

https://gist.github.com/krig/6676d0ae065fd852fac8b445410e1c95

heartbeat/ocf.py
```diff
@@ -338,21 +368,33 @@ def call_handler(func):
 		else:
 			params = inspect.getargspec(func).args
 		def default_for_parameter(paramname):
-			for meta in metadata.parameters:
+			for meta in agent.parameters:
```
oalbrigt: I think this should be for p

heartbeat/ocf.py
```diff
@@ -141,17 +141,21 @@ def emit(self, record):
 logger = logging.LoggerAdapter(log, {'OCF_RESOURCE_INSTANCE': OCF_RESOURCE_INSTANCE})
 
 
+_exit_reason_set = False
```
oalbrigt: Wouldnt it be better to have this as a parameter in the Agent class?

heartbeat/ocf.py
krig: sure, will fix

heartbeat/ocf.py
krig: yeah, that's cleaner. I'll do that.

krig: Ah... actually, that won't work. The idea is to catch the case where the agent code already set the exit reason, so that catching the exception doesn't overwrite that. But since `ocf_exit_reason` is a free function, it needs to record having been called itself.

Issue #1026, opened 2017-08-25
iSCSITarget always fails to start (LIO-T mode)


johnkeates:
The iSCSITarget RA never succeeds in automatically starting a target. It seems to first create the target and then tries to create it again (which obviously fails), and then exit with error 1. Checking targetcli shows the target, so it's not exiting gracefully either.

Manually starting it twice with pcs resource debug-start does make it work, but then it still fails in a different way: it never adds the target portal so no connections from initiators can be made.

The target primitive is about as simple as it gets:

```
primitive iscsi0-target iSCSITarget \
        params implementation=lio-t iqn="iqn.2017-08.access.net:prod-1-ha"  \
        op monitor interval=30s \
        meta target-role=Started
```

johnkeates:
Interestingly, on line 335 in the iSCSITarget RA, it states:

`                # lio distinguishes between targets and target portal
                # groups (TPGs). We will always create one TPG, with the
                # number 1. In lio, creating a network portal
                # automatically creates the corresponding target if it
                # doesn't already exist.`

johnkeates:
Digging around some more: on line 334 there seems to be a case where a loop containing an if/else block might create a default target for the tpg if a default portal was found:

```
                for portal in ${OCF_RESKEY_portals}; do
                        if [ $portal != ${OCF_RESKEY_portals_default} ] ; then
                                IFS=':' read -a sep_portal <<< "$portal"
                                ocf_run targetcli /iscsi/${OCF_RESKEY_iqn}/tpg1/portals create "${sep_porta$
                        else
                                ocf_run targetcli /iscsi create ${OCF_RESKEY_iqn} || exit $OCF_ERR_GENERIC
                        fi
                done
```

This is later on not checked before the actual function that is supposed to create the target and this causes the issue.

So specifying a portal="0.0.0.0.0:3260" or no portal at all wil cause iSCSITarget to fail since LIO-T will have created a target automatically before the RA reaches the point where it wants to create the target itself.

dmuhamedagic:
On Fri, Aug 25, 2017 at 07:45:22AM -0700, John Keates wrote:
> The iSCSITarget RA never succeeds in automatically starting a target. It seems to first create the target and then tries to create it again (which obviously fails),

The start action must be idempotent, so this is already a problem.
The relevant error is this:

ERROR: This Target already exists in configFS

> and then exit with
> error 1. Checking targetcli shows the target, so it's not
> exiting gracefully either.
> 
[...]

PR #1239, opened 2018-10-12
possible fix for #1026

colttt:
add an if-condition and remove an useless 'targetcli create'


oalbrigt:
ok to test

oalbrigt:
LGTM. What do you think @krig ?

krig:
Yeah, looks right to me.

PR #1243, opened 2018-10-16
Python "make release" fixes

oalbrigt:


krig:
LGTM.

krig:
Hmm.

Since we are already passing the python agents through a preprocessor for @PYTHON@, maybe it would make sense to move the path manipulations into a macro as well so that changes to it only need to be done in one place?

oalbrigt:
Let's consider that for the next release. I'd prefer if we can detect which ones to do from configure.ac, as that has all the magic for replacing any available parameters already.

I'm thinking of moving the libs out of the heartbeat directory so we can make some automation to avoid having to add anything manually in the future.

PR #1246, opened 2018-10-17
dev: ocf.py: depth is not a time parameter and shouldn't a suffix

krig:
I'm not 100% sure this breaks Pacemaker, but it causes crmsh to warn about missing monitor action.

### DIFF
--- a/heartbeat/ocf.py
+++ b/heartbeat/ocf.py
@@ -250,7 +250,7 @@ class Action(object):
 	def to_xml(self):
 		def opt(s, name, var):
 			if var is not None:
-				if type(var) == int:
+				if type(var) == int and name in ("timeout", "interval"):
 					var = "{}s".format(var)
 				return s + ' {}="{}"'.format(name, var)
 			return s


heartbeat/ocf.py
```diff
@@ -248,16 +248,16 @@ def __str__(self):
...
 				if type(var) == int:
-					var = "{}s".format(var)
+					var = "{}{}".format(var, "s" if time else "")
 				return s + ' {}="{}"'.format(name, var)
 			return s
 		ret = '<action name="{}"'.format(self.name)
 		ret = opt(ret, "timeout", self.timeout)
 		ret = opt(ret, "interval", self.interval)
-		ret = opt(ret, "depth", self.depth)
+		ret = opt(ret, "depth", self.depth, time=False)
```
oalbrigt: Shouldnt these be += ?

heartbeat/ocf.py
```diff
@@ -248,16 +248,16 @@ def __str__(self):
 		return self.to_xml()
 
 	def to_xml(self):
-		def opt(s, name, var):
+		def opt(s, name, var, time=True):
 			if var is not None:
 				if type(var) == int:
-					var = "{}s".format(var)
+					var = "{}{}".format(var, "s" if time else "")
```
oalbrigt: Maybe we could just do `if name in ("timeout", "interval")` instead? Seems less hacky.

heartbeat/ocf.py
krig: no, it takes the previous version of the string as the first parameter and returns the updated version.. so the += is done inside the opt function, if that makes sense.

heartbeat/ocf.py
krig: yeah, actually that is probably clearer... I'll change it.

heartbeat/ocf.py
oalbrigt: Ah. I didnt notice that :)

oalbrigt:
Perfect. Thanks.

PR #1249, opened 2018-10-17
rabbitmq-cluster: fail monitor when node is in minority partition

jeckersb:
It's possible for mnesia to still be running, but for mnesia to be
partitioned.  And it's also possible to get into this state without
pacemaker seeing the node go down so no corrective action is taken.

When monitoring, check the number of nodes that pacemaker thinks is
running, and compare to the number of nodes that mnesia thinks is
running.  If mnesia only sees a minority of the total nodes, fail it
so corrective action can be taken to rejoin the cluster.

This also adds a new function, rmq_app_running, which simply checks
whether the app is running or not and does not care about the
partition status.  This is now used instead of the full monitor in a
few places where we don't care about partition state.

Resolves: RHBZ#1639826

### DIFF
--- a/heartbeat/rabbitmq-cluster
+++ b/heartbeat/rabbitmq-cluster
@@ -178,10 +178,31 @@ remove_pid () {
 	rm -f ${RMQ_PID_FILE} > /dev/null 2>&1
 }
 
+rmq_app_running() {
+	if $RMQ_CTL eval 'application:which_applications().' | grep -q '{rabbit,'; then
+		ocf_log debug "RabbitMQ application is running"
+		return $OCF_SUCCESS
+	else
+		ocf_log debug "RabbitMQ application is stopped"
+		return $OCF_NOT_RUNNING
+	fi
+}
+
 rmq_monitor() {
 	local rc
 
 	if $RMQ_CTL eval 'rabbit_mnesia:cluster_status_from_mnesia().' | grep -q '^{ok'; then
+		pcs_running=$(rmq_join_list | wc -w)
+		ocf_log debug "Pacemaker thinks ${pcs_running} RabbitMQ nodes are running"
+		rmq_running=$($RMQ_CTL eval 'length(mnesia:system_info(running_db_nodes)).')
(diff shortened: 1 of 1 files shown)


krig:
Spontaneously to me, this feels like something that might cripple an already damaged cluster in certain cases, like if one mnesia node fails but Pacemaker on that node has yet to notice, and the monitor on another node then decides to kill its instance of mnesia since the nodecount isn't matching.

jeckersb:
> Spontaneously to me, this feels like something that might cripple an already damaged cluster in certain cases, like if one mnesia node fails but Pacemaker on that node has yet to notice, and the monitor on another node then decides to kill its instance of mnesia since the nodecount isn't matching.

I think this should be ok still, because it won't fail if there is a mismatch, only if the mismatch is a minority.

Consider a cluster of nodes A,B,C with the following sequence:

- All mnesia nodes are up, the nodename attrs are set for each so pacemaker thinks all three are up.
- mnesia on A becomes partitioned by itself
- Monitor action runs on B: pacemaker still thinks there are three nodes up (all node attrs are still present because failure has not been detected by monitor yet), but querying mnesia on B will show only two members.  Because two out of three is not a minority, the monitor still succeeds for B.
- If monitor on C runs, same thing as above.
- Monitor action runs on A: pacemaker still thinks there are three nodes up (all node attrs are still present because failure has not been detected by monitor yet), but querying mnesia shows only one.  Minority is detected and the monitor fails.
- Node attr for A is removed.
- A is stopped and restarted to attempt to rejoin.

At this point if the partition was brief and has been corrected, A should be able to rejoin the cluster and everything is fine.  If not, the resource will remain stopped until it is manually cleaned up once the environment has recovered.

While A is down, the monitor actions for B and C will show two nodes both in the pacemaker node attrs as well as from mnesia, so they will continue to function with quorum.

krig:
Ahh I see, so it's not just that the node count mismatches, but that only a minority is visible. That makes sense.

Thanks for the explanation!

krig:
The code looks good to me as well.

oalbrigt:
LGTM. Thanks.

PR #1253, opened 2018-10-19
ZFS: cleanup different indentations

gmelikov:
Signed-off-by: George Melikov <mail@gmelikov.ru>

Looks like many agent files have different indentations (tabs and spaces) per file and just wrong indentations. This PR is for ZFS only,
but I think it would be great to cleanup other files too (first example is https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/IPaddr2).

### DIFF
--- a/heartbeat/ZFS
+++ b/heartbeat/ZFS
@@ -31,7 +31,7 @@ USAGE="usage: $0 {start|stop|status|monitor|validate-all|meta-data}";
 #######################################################################
 
 meta_data() {
-        cat <<END
+	cat <<END
 <?xml version="1.0"?>
 <!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
 <resource-agent name="ZFS">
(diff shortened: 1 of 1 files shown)


oalbrigt:
We only fix indentation as part of fixing code with mixed indentation, as that makes it easier to see why the code was added/changed via blame, and it has the additional benefit of seeing only the fixed lines instead of a lot of lines that arent related if you diff a distro's version of a file against upstream.

gmelikov:
@oalbrigt so it's the intended situation. Should I close this PR then, or we can merge it for ZFS (it doesn't have many commits though)?

oalbrigt:
Yeah, it's quite new. What do you think @krig ?

krig:
To me, the big problem is commits that mix indentation fixes with other fixes, so I think this one is OK. On the other hand it does make backporting patches a bit messier, but in this case since the agent is so new and AFAIK no distribution that ships ZFS also support this agent.

oalbrigt:
Thanks.

PR #1255, opened 2018-10-23
spec: fix broken reliance on systemd_requires defined in SRPM build time

jnpkrn:
Actually orthogonal to how buildroot (usually used also for building
source packages alone) looks like/which packages are always present.
In a container, for instance, may not be any systemd (and hence its
RPM macros) at all.

Still rather fragile but working "is systemd a native platform where
you build this SRPM" detecting macro borrowed from upstream pacemaker
spec (I happen to author there, for completeness and as a reminder,
please keep these snippets in sync if at all possible).

Signed-off-by: Jan Pokorný <jpokorny@redhat.com>

### DIFF
--- a/resource-agents.spec.in
+++ b/resource-agents.spec.in
@@ -14,6 +14,22 @@
 %global numcomm @numcomm@
 %global dirty @dirty@
 
+# Whether this platform defaults to using systemd as an init system
+# (needs to be evaluated prior to BuildRequires being enumerated and
+# installed as it's intended to conditionally select some of these, and
+# for that there are only few indicators with varying reliability:
+# - presence of systemd-defined macros (when building in a full-fledged
+#   environment, which is not the case with ordinary mock-based builds)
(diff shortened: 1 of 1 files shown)


oalbrigt:
These issues are easily fixed by adding: `BuildRequires: systemd`

jnpkrn:
On 24/10/18 01:00 -0700, Oyvind Albrigtsen wrote:
> These issues are easily fixed by adding: `BuildRequires: systemd`

Sure, but this way, you oversimplify the landscape, which is not
appropriate for an upstream spec file -- you would cut people off
from the possibility to build the package for EL 6 based distros,
for instance.

With my change, it would keep working just fine.  Definitely it
wouldn't choke on the lack of systemd packages/macros.

-- 
Nazdar,
Jan (Poki)

oalbrigt:
Ah. Makes sense.

oalbrigt:
Thanks.

PR #1258, opened 2018-10-26
Fix demote example log line.

canardleteer:
This tiny typo confused me for a few minutes. Thought I'd throw up a patch to prevent others from getting confused by the same.


krig:
Good catch, thanks!

PR #1261, opened 2018-11-01
Dev: ocf.py: Add support for role argument to actions

krig:
role=Master / role=Slave is commonly used to define
different monitor actions when agent is promoted or
demoted.


oalbrigt:
LGTM.

krig:
Thanks.

PR #1264, opened 2018-11-06
tomcat: use systemd on newer RHEL where catalina.sh is unavailable

oalbrigt:
without this patch it falls back to running the command manually which fails due to missing PID-file.

The -e check might be a bit confusing, but it'll avoid issues if it's backported to older RHEL where it's setup with catalina.sh available.


krig:
Looks OK to me.

PR #1265, opened 2018-11-11
lxc: Add Support for lxc-stop

Lukey3332:
This adds an "use_lxc_stop" switch which enables stopping
the container using lxc-stop instead of kill -PWR.

This was tested on a two node pacemaker cluster with drbd.

I wonder why it isn't already used and if we could change it to the default stop action.

### DIFF
--- a/heartbeat/lxc.in
+++ b/heartbeat/lxc.in
@@ -179,8 +179,7 @@ LXC_start() {
 
 
 LXC_stop() {
-	local shutdown_timeout
-	local now
+	
 	LXC_status
 	if [ $? -eq $OCF_NOT_RUNNING ]; then
 		ocf_log debug "Resource $OCF_RESOURCE_INSTANCE is already stopped"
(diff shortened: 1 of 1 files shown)


krig:
Perhaps this agent was written before lxc-stop existed? Either way, according to the man page for `lxc-stop` it does kill -PWR as well, followed by SIGKILL after 60 seconds. Maybe it would be OK to avoid the parameter and use `lxc-stop` if it is found to be installed?

Lukey3332:
For me the Previous method didn't work, it always ran into the timeout and then hard killed the Container, with lxc-stop it does an ordered stop as expected.

krig:
Hm, well. Actually I think it would be fine to assume that lxc-stop is present, and just use that all the time. @oalbrigt ?

Also, you might want to pass `--timeout=T` to lxc-stop and pass it the configured timeout - 5 similar to the `shutdown_timeout` variable. I have to say that the default stop timeout of 30s in the metadata is extremely short, but that's a separate issue...

krig:
Yeah, that's odd. Personally I would prefer to clean it up, make it use lxc-stop and avoid adding the parameter.

krig:
Ah, no, it's the return codes from `ocf_version_cmp` that are confusing. From the doc:

```
#     0: VER1 smaller (older) than VER2
#     1: versions equal
#     2: VER1 greater (newer) than VER2
#     3: bad format
```

So the version check does what you'd expect: on pre-1.0, it runs without `-k`.

krig:
So, I think it's safe to say that the complicated logic in stop here was created before lxc-stop implemented that stuff. I would prefer to remove the added parameter, and just make the stop operation call lxc-stop.

oalbrigt:
We should probably try to be backwards compatible. Maybe you could use case to mach 1|2 for equal or great.

Lukey3332:
So if lxc version is smaller than 1.0.0 we use the legacy Code, and lxc-stop otherwise?

krig:
Yeah, that would work for me at least.

krig:
Looks good to me.

PR #1269, opened 2018-11-28
Enable --query flag in DescribeRouteTable API 

gguifelixamz:
This is a proposed fix to avoid a race condition with "grep" while parsing the results from AWS CLI. 

Currently, "grep" will match for IPs that don't belong tho the cluster resource. For example, if your cluster resource IP address is "192.168.1.1", "grep" will match "192.168.1.1*" (192.168.1.1,10,11,...,111).

This could have been fixed by a simple "grep -w" but we (AWS) thought it is more elegant to directly use the API to query for the IP.

I also included a debug log line.

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -167,7 +167,7 @@ ec2ip_validate() {
 ec2ip_monitor() {
 	if ocf_is_true ${OCF_RESKEY_monapi} || [ "$__OCF_ACTION" = "start" ] || ocf_is_probe; then
 		ocf_log info "monitor: check routing table (API call)"
-		cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile --output text ec2 describe-route-tables --route-table-ids $OCF_RESKEY_routing_table --query RouteTables[*].Routes[?DestinationCidrBlock==\`$OCF_RESKEY_address/32\`].InstanceId"
+		cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile --output text ec2 describe-route-tables --route-table-ids $OCF_RESKEY_routing_table --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].InstanceId"
 		ocf_log debug "executing command: $cmd"
 		ROUTE_TO_INSTANCE=$($cmd)
 		ocf_log debug "Overlay IP is currently routed to ${ROUTE_TO_INSTANCE}"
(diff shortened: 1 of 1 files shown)


gguifelixamz:
Changed the outer quotes to " and removed inner quotes as they're no longer needed

oalbrigt:
The indentation is is replaced with spaces in the updated lines. Can you update it to tabs (for consistency with the rest of the code)?

You can see the "Files changed" tab to see the lines not aligning with the rest of the code.

gguifelixamz:
Oops! My bad! :-) Fixed now

heartbeat/aws-vpc-move-ip
```diff
@@ -167,9 +167,10 @@ ec2ip_validate() {
 ec2ip_monitor() {
 	if ocf_is_true ${OCF_RESKEY_monapi} || [ "$__OCF_ACTION" = "start" ] || ocf_is_probe; then
 		ocf_log info "monitor: check routing table (API call)"
-		cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile --output text ec2 describe-route-tables --route-table-ids $OCF_RESKEY_routing_table"
+		cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile --output text ec2 describe-route-tables --route-table-ids $OCF_RESKEY_routing_table --query RouteTables[*].Routes[?DestinationCidrBlock==\`$OCF_RESKEY_address/32\`].InstanceId"
```
oalbrigt: The quotes here should be `'` the backslash followed by reverse quote (`).

oalbrigt: Oh, and I just noticed it has to end with `_ip` instead of `_address`, as the backwards-compatible to avoid issues for users using the `ip` parameter instead of the deprecated `address` parameter

gguifelixamz: The \` cannot be replaced by `\' ` as it will break the JMSPath Query:

> Bad value for --query RouteTables[*].Routes[?DestinationCidrBlock==\'192.168.1.9/32\'].InstanceId: Bad jmespath expression: Unknown token \:
> RouteTables[*].Routes[?DestinationCidrBlock==\'192.168.1.9/32\'].InstanceId

gguifelixamz: > Sorry, I meant ' without the backspace.

That seems to work well, I will give it a go

oalbrigt:
ok to test

gguifelixamz:
@oalbrigt sorry for first-timer silly mistakes. I was originally testing this (since some months ago) with a version that was still using _address to when I merged with the latest version I never changed to _ip.

The quotation part was a bit of pain as I was trying to use a combination of quotes that we usually use in when issuing "aws cli" commands from a shell.

Thanks for your patience

oalbrigt:
No problem. And thank you for the patch and patience in improving it.

diegoakechi:
@gguifelixamz  Thansk for the patch.

@oalbrigt @krig  Can you squash the commits when merging the PR, as it is not a big change? It helps backports if needed.

oalbrigt:
@diegoakechi Done.

diegoakechi:
@oalbrigt  Thanks!  ;)

PR #1270, opened 2018-11-29
docker: Fix issues with stop operation

nrwahl2:
The docker RA's stop operation doesn't behave properly in some cases.

1. It returns a false success code in case of an error response from
the daemon.
2. It fails at `remove_container()` if the container does not exist
but another docker object of the same name does exist.

In case 1, the `container_exists()` function returns the same exit code
(1) if the container is not found (an expected error) or if there is an
error response from the docker daemon (an unexpected error). These types
of errors should be handled differently.

In case 2, the `docker inspect` calls do not limit their search to
containers. So if a non-container object is found with a matching name,
the RA attempts to remove a container by that name. Such a container may
not exist.

This patch fixes these issues as follows:
1. Match an error response in `container_exists()` against the string
"No such container".
2. Add `--type=container` to the `docker inspect` calls to restrict
the match.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -215,7 +215,7 @@ monitor_cmd_exec()
 		out=$(docker exec ${CONTAINER} $OCF_RESKEY_monitor_cmd 2>&1)
 		rc=$?
 	else
-		out=$(echo "$OCF_RESKEY_monitor_cmd" | nsenter --target $(docker inspect --format {{.State.Pid}} ${CONTAINER}) --mount --uts --ipc --net --pid 2>&1)
+		out=$(echo "$OCF_RESKEY_monitor_cmd" | nsenter --target $(docker inspect --type=container --format {{.State.Pid}} ${CONTAINER}) --mount --uts --ipc --net --pid 2>&1)
 		rc=$?
 	fi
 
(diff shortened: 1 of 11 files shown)


nrwahl2:
RHBZ#1629357
RHBZ#1631225

heartbeat/docker
```diff
@@ -236,7 +236,25 @@ monitor_cmd_exec()
...
 {
-	docker inspect --format {{.State.Running}} $CONTAINER | egrep '(true|false)' >/dev/null 2>&1
+	local err
+
+	err=$(docker inspect --type=container $CONTAINER 2>&1 >/dev/null)
+
+	if [ $? -ne $OCF_SUCCESS ]; then
+		case $err in
+			*"No such container"*)
+				# Container does not exist
```
oalbrigt: I think we should add e.g. `ocf_exit_reason "Container doesnt exist"` or similar here, or just skip the entire case if the output from `docker inspect` is clear enough to output directly as in `*)`.

krig:
Other than @oalbrigt 's comment, this looks good to me as well

heartbeat/docker
nrwahl2: > I think we should add e.g.` ocf_exit_reason "Container doesnt exist"` or similar here

Is that still relevant since we're not exiting if the container doesn't exist?

"No such container" => Return 1 from the function
Any other error => Exit with code 1 from the whole script

oalbrigt: It seems like you're right. Maybe just `return 1` instead of the error-code and update to something along the lines of "return instead of exit when container doesnt exist", as this wasnt easy to think was a mistake.

nrwahl2: Sorry I dropped the ball on this one till now. I initially had `return 1` and changed it to `OCF_ERR_GENERIC` in response to feedback. I think there was some misunderstanding between us at the time on what it was being used for. I'll push an updated version in a minute.

nrwahl2: Done.

nrwahl2: Botched commit history now... will try to fix.

nrwahl2:
Opened new PR #1280.

PR #1272, opened 2018-12-03
Do not log at debug log level when HA_debug is unset

mbaldessari:
There might be situations (e.g. bundles) where the HA_debug variable
is unset. It makes little sense to enable debug logging when the HA_debug env
variable is unset.
So let's skip debug logs when HA_debug is set to 0 or is unset.

Tested inside a bundle and observed that previously seen 'ocf_log debug'
calls are now correctly suppressed (w/ HA_debug being unset inside the
container)

### DIFF
--- a/heartbeat/ocf-shellfuncs.in
+++ b/heartbeat/ocf-shellfuncs.in
@@ -257,7 +257,7 @@ ha_log()
 
 ha_debug() {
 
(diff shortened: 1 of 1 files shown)


krig:
Makes sense, thanks!

oalbrigt:
Thanks.

PR #1274, opened 2018-12-05
rabbitmq-cluster: better ensure node attributes are removed

jeckersb:
Ensure that the attribute is removed at the end of the stop action.
Also if rmq_app_running or rmq_node_alive shows the service as down,
ensure the attribute is deleted as well.

Resolves: RHBZ#1656368


krig:
Alright, looks good to me

mbaldessari:
Nice catch, Eck. Thanks!

PR #1276, opened 2018-12-13
rabbitmq-cluster: always use quiet flag for eval calls

jeckersb:
On older rabbitmq versions, rabbitmqctl appends "...done." at the end
of the output.  However we expect eval without this extra output so it
can be used for further processing.  The -q option to rabbitmqctl
suppresses the extra output, so ensure we always pass that when
calling eval.

Resolves: RHBZ#1659072


jeckersb:
@mbaldessari

oalbrigt:
LGTM.

PR #1277, opened 2018-12-18
Medium: pgsql: Set an initial score for primary and hot standby in the prove.

knakahira:
Currently, pgsql RA does not set an initial score for primary and hot standby in the prove operation.
This behavior cause following problems.

1. If primary PostgreSQL instance detected in the probe operation, then pgsql RA have no chance to set a master score "1000" because pgsql_promote() is not executed.
   It cause unnecessary fail over in some situations. (E.g. restart Pacemaker in the maintenance mode)

2. If hot standby PostgreSQL instance detected in the probe operation, then control_slave_status() can't update master score for hot standby because master score for hot standby is null.
   It cause lack of master score for hot standby in the "crm_mon -A". 

This patch aim to avoid above problems by setting an initial score for primary and hot standby in the prove operation.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -974,11 +974,19 @@ pgsql_real_monitor() {
         case "$output" in
             f)  ocf_log debug "PostgreSQL is running as a primary."
                 if [ "$OCF_RESKEY_monitor_sql" = "$OCF_RESKEY_monitor_sql_default" ]; then
+                    if ocf_is_probe; then
+                        # Set initial score for primary.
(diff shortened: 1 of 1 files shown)


oalbrigt:
test this please

oalbrigt:
@krig What do you think about this? Do you have a better suggestion for how to improve this?

krig:
hmm, master score handling is always fiddly.. I don't know if I can think of a better solution either.

oalbrigt:
Thanks.

PR #1279, opened 2019-01-02
vdo-vol: new resource agent

oalbrigt:


oalbrigt:
Happy new year.

Yeah, we ended up with vdo-vol as it's shorter, and vdo is what the systemd service is called.

PR #1280, opened 2019-01-03
docker: Fix issues with stop operation

nrwahl2:
The docker RA's stop operation doesn't behave properly in some cases.
  1. It returns a false success code in case of an error response from
     the daemon.
  2. It fails at `remove_container()` if the container does not exist
     but another docker object of the same name does exist.

In case #1, the `container_exists()` function returns the same exit code
(1) if the container is not found (an expected error) or if there is an
error response from the docker daemon (an unexpected error). These types
of errors should be handled differently.

In case #2, the `docker inspect` calls do not limit their search to
containers. So if a non-container object is found with a matching name,
the RA attempts to remove a container by that name. Such a container may
not exist.

This patch fixes these issues as follows:
  1. Match an error response in `container_exists()` against the string
     "No such container".
  2. Add `--type=container` to the `docker inspect` calls to restrict
     the match.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -215,7 +215,7 @@ monitor_cmd_exec()
 		out=$(docker exec ${CONTAINER} $OCF_RESKEY_monitor_cmd 2>&1)
 		rc=$?
 	else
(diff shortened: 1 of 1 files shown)


nrwahl2:
RHBZ#1629357
RHBZ#1631225

nrwahl2:
Same as #1270. Opened new PR with fixed commit history and closed the old one.

oalbrigt:
Thanks.

PR #1281, opened 2019-01-09
LVM-activate: dont fail initial probe

oalbrigt:


vvidic:
Yep, resource does not start properly without this one.

PR #1285, opened 2019-01-15
LVM-activate: fix dmsetup check

vvidic:
When there are no devices in the system dmsetup outputs one line:

  # dmsetup info -c
  No devices found


oalbrigt:
LGTM. Thanks.

PR #1286, opened 2019-01-15
LVM-activate: make vgname not unique

vvidic:
If activating one lvname at a time, vgname will not be unique.

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -102,7 +102,7 @@ because some DLM lockspaces might be in use and cannot be closed automatically.
 <shortdesc lang="en">This agent activates/deactivates logical volumes.</shortdesc>
 
 <parameters>
-<parameter name="vgname" unique="1" required="1">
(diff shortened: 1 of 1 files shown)


oalbrigt:
Nice catch. Thanks.

jnpkrn:
Just FYI, while this change is sound from the perspective of
https://github.com/ClusterLabs/resource-agents/blob/master/doc/dev-guides/ra-dev-guide.asc
(and RGmanager, for that matter), it's not so much when it comes
to pacemaker, since it only implements this understanding of `unique:

- https://github.com/ClusterLabs/pacemaker/blob/Pacemaker-2.0.1-rc4/daemons/controld/controld_execd.c#L582
- https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/2.0/html-single/Pacemaker_Explained/index.html#_reloading_services_after_a_definition_change

Just that there are no unrealistic expectations.

PR #1291, opened 2019-01-17
 Added support for multiple slaves to pgsql.

mitaka:
When you have a single postgresql cluster managed by corosync
and pacemaker you cannot have any further slaves either with
streaming or logical replication. This patch enables connecting
other postgresql servers as slaves to the cluster without breaking
or interfering with the running cluster.

Signed-off-by: Dimitar Ianakiev <dimitar.q@siteground.com>

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -73,6 +73,8 @@ OCF_RESKEY_xlog_check_count_default="3"
 OCF_RESKEY_crm_attr_timeout_default="5"
 OCF_RESKEY_stop_escalate_in_slave_default=90
 OCF_RESKEY_replication_slot_name_default=""
+OCF_RESKEY_replication_interface_default="eth0"
+OCF_RESKEY_replication_multiple_slaves_default="false"
 
 : ${OCF_RESKEY_pgctl=${OCF_RESKEY_pgctl_default}}
 : ${OCF_RESKEY_psql=${OCF_RESKEY_psql_default}}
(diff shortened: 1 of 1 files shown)


heartbeat/pgsql
```diff
@@ -107,6 +108,7 @@ OCF_RESKEY_replication_slot_name_default=""
 : ${OCF_RESKEY_crm_attr_timeout=${OCF_RESKEY_crm_attr_timeout_default}}
 : ${OCF_RESKEY_stop_escalate_in_slave=${OCF_RESKEY_stop_escalate_in_slave_default}}
 : ${OCF_RESKEY_replication_slot_name=${OCF_RESKEY_replication_slot_name_default}}
+: ${OCF_RESKEY_replication_interface=${OCF_RESKEY_replication_slot_interface}}
```
oalbrigt: Seems like you got `slot_` in here by accident, and forgot to add  `_default`.

oalbrigt:
Can you update the indentation to follow the tab-style of the agent?

I think we should default to not having the interface and just `if [ -n "$...interface" ]; then` for this additional code to avoid issues for users upgrading.

heartbeat/pgsql
mitaka: This is now fixed.

mitaka:
I believe i fixed the identation and made the new functionality optional by activating it explicitly. The other comment for the _default variable is addressed too.

heartbeat/pgsql
```diff
@@ -1915,14 +1946,44 @@ pgsql_validate_all() {
         else
             CHECK_XLOG_LOC_SQL="select pg_last_xlog_replay_location(),pg_last_xlog_receive_location()"
         fi
-        CHECK_REPLICATION_STATE_SQL="select application_name,upper(state),upper(sync_state) from pg_stat_replication"
+
+        NODE_LIST=`echo $OCF_RESKEY_node_list | tr '[A-Z]' '[a-z]'`
+
+        if [ -n "$OCF_RESKEY_replication_multiple_slaves" ];then
```
oalbrigt: You could do a `if ocf_is_true "$OCF_RESKEY_replication_multiple_slaves"; then` instead, and set it to e.g. no or off in the _default on top for clarity.

mitaka: This is a nice one. Thanks! Fixed it with ocf_is_true which is definitely more readable.

oalbrigt:
LGTM. Thanks.

PR #1296, opened 2019-01-24
Revert " Added support for multiple slaves to pgsql."

mitaka:
Reverts ClusterLabs/resource-agents#1291

Upon further extensive testing we found a corner case where the replication is broken when a failed node joins the back cluster. Please revert until we find the root cause and provide a fix.


oalbrigt:
Thanks.

PR #1305, opened 2019-03-18
Squid: fix pid file issue due to new Squid version saving the PID of the parent process instead of the listener child process

oalbrigt:
New versions of Squid (e.g. 4.2) saves the PID of the parent process to the pidfile instead of the PID of the listener child process as it used.


krig:
Looks good.

PR #1309, opened 2019-04-01
LVM-activate: return OCF_NOT_RUNNING on initial probe

zzhou1:
@In the use case of lvm on top of cluster md/raid. When the fenced node
rejoins to the cluster, Pacemaker will run the monitor action for the
probe operation. At that time, LVM PV and VG won't exist before cluster md/raid
get assembled, and the probe should return $OCF_NOT_RUNNING instead of
$OCF_ERR_CONFIGURED.

Another cosmetic fix is to allign dmsetup terminology to its latest
document `dmsetup info -c -S help`, hence to change 'vgname/lvname' to
'vg_name/lv_name'.

Signed-off-by: Roger Zhou <zzhou@suse.com>


oalbrigt:
ok to test

heartbeat/LVM-activate
```diff
@@ -701,7 +707,7 @@ tagging_deactivate() {
 # method:
 #
 # lv_count=$(vgs --foreign -o lv_count --noheadings ${VG} 2>/dev/null | tr -d '[:blank:]')
-# dm_count=$(dmsetup --noheadings info -c -S "vgname=${VG}" 2>/dev/null | grep -c "${VG}-")
```
krig: This and the other `vgname` to `vg_name` / `lvname` to `lv_name` should probably be a separate commit, and make sure to write a commit message to explain this change: Is it a simple typo, or is it a change between versions? If it is the latter, there should be a version check to use the old name with older versions.

PR #1310, opened 2019-04-01
LVM-activate: fix initial probe and align dmsetup command to standard

zzhou1:


heartbeat/LVM-activate
```diff
@@ -481,6 +482,11 @@ lvm_validate() {
 			exit $OCF_SUCCESS
 		fi
 
+		if ocf_is_probe; then
+			ocf_log info "initial probe: VG [${VG}] doesn't exist on any block device yet."
```
renzhengeek: It's been long time for me looking at this code. May I ask a question: what's the difference between $OCF_NOT_RUNNING and $OCF_ERR_CONFIGURED? I mean the difference for pacemaker behavior, not just the log msg.

renzhengeek: "VG [${VG}] doesn't exist on any block device yet"

Maybe the VG is already on the disks, but the raid not assembled up yet? If so, a better msg could be like "... not found yet ..."?

oalbrigt: CONFIGURED is when you might've set some settings incorrectly, so the resource wont try to restart/relocate if this is returned, while NOT_RUNNING is the expected state of a stopped resource and for a started resource will make it try to start again and relocate if that fails.

krig:
Looks good to me now.

heartbeat/LVM-activate
renzhengeek: > CONFIGURED is when you might've set some settings incorrectly, so the resource wont try to restart/relocate if this is returned, while NOT_RUNNING is the expected state of a stopped resource and for a started resource will make it try to start again and relocate if that fails.

clear enough, thanks!

With the info msg tweaked, this change LGTM then.

zzhou1: > Maybe the VG is already on the disks, but the raid not assembled up yet? If so, a better msg could be like "... not found yet ..."

En, sounds better ;)

zzhou1:
Force update my "for-next"  branch to improve based on Eric's comments.
Thanks all !

oalbrigt:
ok to test

oalbrigt:
Thanks.

oalbrigt:
Link to original PR for future reference: https://github.com/ClusterLabs/resource-agents/pull/1309

PR #1314, opened 2019-04-05
[rabbitmq] Stop redirecting all output to /dev/null in set_policy

mbaldessari:
The command only runs on the bootstrap node and we actually are very
interested in the output of it, because if it failed we want to know
why. It produces zero extra output when successful so let's get some
debuggability for free here.
f


oalbrigt:
LGTM. Thanks.

PR #1316, opened 2019-04-09
Iscsi target stop in failed resource state

jiri-lunacek:
When the LUN resource is in failed state e.g. the backing store file or device is not available, this prevented targetcli cleanup.

When the backing store file or device comes back online and admin issues resource cleanup this resulted in targetcli config being partially present and start operation failing.

I'm not sure why this block is there in the first place because subsequent code is prepared for partial targetcli config status.

Is there a reason for stop operation to return success every time the resource is not-running?


krig:
ok to test

oalbrigt:
ok to test

PR #1320, opened 2019-04-16
aws-vpc-move-ip - fix for defect making update of route table fail in case VM has multiple network interfaces

krojzl:
In case that AWS VM is having more that one network interface the following command with fail:
```
hostname01:/usr/lib/ocf/resource.d/suse # /usr/bin/aws --profile cluster --output text ec2 replace-route --route-table-id rtb-1234a567812a34567 --destination-cidr-block 192.168.1.1/32 --instance-id i-1234567890b012345

An error occurred (InvalidInstanceID) when calling the ReplaceRoute operation: There are multiple interfaces attached to instance 'i-1234567890b012345'. Please specify an interface ID for the operation instead.
```

In such case the command should use `--network-interface-id` parameter.

The adjustment is using value `$OCF_RESKEY_interface` that is already passed down and used in the code to determine MAC address of given network interface and this value is subsequently used to query AWS CLI to get associated network interface id.

This fix does not need any additional security policy privileges and should be non-disruptive to existing implementations.

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -222,25 +222,18 @@ ec2ip_get_and_configure() {
 	MAC_FILE="/sys/class/net/${OCF_RESKEY_interface}/address"
 	if [ -f $MAC_FILE ]; then
 		cmd="cat ${MAC_FILE}"
-		ocf_log debug "executing command: $cmd"
-		MAC_ADDR="$(eval $cmd)"
-		rc=$?
-		if [ $rc != 0 ]; then
-			ocf_log warn "command failed, rc: $rc"
-			return $OCF_ERR_GENERIC
-		fi
-		ocf_log debug "MAC address associated with interface ${OCF_RESKEY_interface}: ${MAC_ADDR}"
 	else
 		cmd="ip -br link show dev ${OCF_RESKEY_interface} | tr -s ' ' | cut -d' ' -f3"
-		ocf_log debug "executing command: $cmd"
-		MAC_ADDR="$(eval $cmd)"
-		rc=$?
-		if [ $rc != 0 ]; then
-			ocf_log warn "command failed, rc: $rc"
-			return $OCF_ERR_GENERIC
-		fi
-		ocf_log debug "MAC address associated with interface ${OCF_RESKEY_interface}: ${MAC_ADDR}"
 	fi
+
+	ocf_log debug "executing command: $cmd"
+	MAC_ADDR="$(eval $cmd)"
+	rc=$?
+	if [ $rc != 0 ]; then
+		ocf_log warn "command failed, rc: $rc"
+		return $OCF_ERR_GENERIC
+	fi
+	ocf_log debug "MAC address associated with interface ${OCF_RESKEY_interface}: ${MAC_ADDR}"
 
 	cmd="curl -s http://169.254.169.254/latest/meta-data/network/interfaces/macs/${MAC_ADDR}/interface-id"
 	ocf_log debug "executing command: $cmd"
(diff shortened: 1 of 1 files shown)


krig:
ok to test

krojzl:
Thank you - what is the usual time frame to get the change accepted and merged?

gguifelixamz:
Hello @krojzl,

Thank you for submitting this change to aws-vpc-move-ip agent.

The limitation of a single ENI per cluster node in the agent is not a bug but it is by design, so I would call this as a new feature instead of a fix.

As for your implementation, I would like to suggest a new method to retrieve the interface id as we've been trying to avoid more API calls to be made from inside the agent itself as API calls can be throttled by AWS and cause cluster resources to miss behave and failover unexpectedly.

I would suggest that instead of doing an API call to retrieve the interface id to fetch it directly from the EC2 metadata server by basically replacing:

`cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile --output text ec2 describe-instances --instance-ids $EC2_INSTANCE_ID --query 'Reservations[*].Instances[*].NetworkInterfaces[*].[NetworkInterfaceId,MacAddress]' | grep ${MAC_ADDR} | cut -f1"`

With:

`cmd="curl http://169.254.169.254/latest/metadata/network/interfaces/macs/${MAC_ADDR}/interface-id`

The above is only an example so feel free to adapt as required to your code.

krojzl:
Requested change implemented...

Also spent some time reviewing first command:
```
cmd="ip -br link show dev $OCF_RESKEY_interface | tr -s ' ' | cut -d' ' -f3"
```

and found that MAC address can be also obtained in following way:
```
cmd="cat /sys/class/net/${OCF_RESKEY_interface}/address"
```

However not sure if this path exists on all Linux distributions. Please advice.

gguifelixamz:
Hello @krojzl,

I have mixed feelings about reading it from /sys. However, I would expect it to be "standard" across modern distributions and kernel versions.

If you want to be safe add a strategy to fallback to "ip" in your code, or just use "ip".

Checking other agents I see that iface-bridge resource agent uses /sys to get a lot of info.

Here is the output from SUSE 12 SP[2|3] and Red Hat 7.5:

Red Hat 7.5:
```
# uname -r && cat /sys/class/net/eth0/address
3.10.0-862.29.1.el7.x86_64
12:b3:3f:8f:4b:ac
```

SUSE 12 SP2:
```
# uname -r && cat /sys/class/net/eth0/address
4.4.121-92.85-default
0a:12:ea:b9:94:76
```

SUSE 12 SP3:
```
# uname -r && cat /sys/class/net/eth0/address
4.4.143-94.47-default
12:28:6d:f3:1a:3a
```

SUSE 15
```
# uname -r && cat /sys/class/net/eth0/address
4.12.14-25.6-default
0a:74:8c:0d:f5:4a
```

@oalbrigt and @diegoakechi - would you like to provide your opinion in this matter?

krojzl:
Code was adjusted as proposed. It checks if file `/sys/class/net/${OCF_RESKEY_interface}/address` exists - if yes then it gets MAC from there - otherwise it falls back on `ip link` command.

I think this takes the best from both options.

diegoakechi:
@gguifelixamz I agree with you. AFAIK it will work on most of the Linux distributions, but having a fallback seems the best way to go.

krojzl:
All requested changes were implemented, seems everyone agreed with the approach. Any chance to get the code merged in? Or what will be the next steps?

krig:
> All requested changes were implemented, seems everyone agreed with the approach. Any chance to get the code merged in? Or what will be the next steps?

Looks good to me. Thank you everyone!

PR #1322, opened 2019-04-17
ethmonitor: fix is_interface() regression

hrasiq:
Commit 40d05029ce0b changed is_interface() to list by link instead of address. We need to adjust the grep filter for the new output format, else we'll miss interfaces.

Apologies for missing this previously. This commit should fix is_interface() to behave as expected for all interfaces:
$ ip -o -f link a | grep " lo "

$ ip -o -f link a | grep " lo:"
1: lo: <LOOPBACK,UP,LOWER_UP> ...


krig:
Thank you!

PR #1323, opened 2019-04-19
galera: Ignore safe_to_bootstrap in grastate.dat in some cases

vuntz:
We've seen a case where one node in the cluster had safe_to_bootstrap as
1 in grastate.dat, but apparently the UUID and the last commit (or
seqno) were 00000000-0000-0000-0000-000000000000 and -1 -- which matches
a fresh state.

In that case, we do not want to use that node as a bootstrap node, as it
causes a risk of losing everything.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -564,10 +564,21 @@ detect_first_master()
 detect_safe_to_bootstrap()
 {
     local safe_to_bootstrap=""
+    local uuid=""
+    local seqno=""
 
     if [ -f ${OCF_RESKEY_datadir}/grastate.dat ]; then
         ocf_log info "attempting to read safe_to_bootstrap flag from ${OCF_RESKEY_datadir}/grastate.dat"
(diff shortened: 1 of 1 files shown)


vuntz:
I'm submitting to get some reviews, but I'd like @rhafer to give a +1 before this gets merged.

krig:
I can't imagine a case where you actually would want to bootstrap from a node with zero uuid/seqno, so disallowing that seems sensible to me at least.

rhafer:
Makes sense to me. So +1 for merging this.

It's still unclear however how that node ended up with "safe_to_bootstrap" set to 1 after a failed SST (IIRC). But that's a different problem to figure out. I might be safer to always check for that lateste seqno regardless of the "safe_to_bootstrap" flag. Even though the current approach pretty much reflects the recommended steps from the documentation.

krig:
Thank you! Merging.

PR #1330, opened 2019-05-13
galera: Add "check_passwd_use_empty" parameter

toabctl:
When using a custom "check_user" without a "check_password" *and* the
user has a .my.cnf file in its home directory, the monitoring fails
like:

Operation promote for galera:0 (ocf:heartbeat:galera) returned: 'unknown error' (1)
 >  stderr: INFO: Node <d52-54-77-77-01-02> is bootstrapping the cluster
 >  stderr: ERROR: MySQL is not running
 >  stderr: INFO: Waiting on node <d52-54-77-77-01-01> to report database status before Master instances can start.
 >  stderr: INFO: MySQL is not running
 >  stderr: INFO: MySQL is not running
 >  stderr: ERROR 1045 (28000): Access denied for user 'monitoring'@'localhost' (using password: YES)
 >  stderr: ocf-exit-reason:Unable to retrieve wsrep_cluster_status, verify check_user 'monitoring' has permissions to view status
 >  stderr: ocf-exit-reason:local node <d52-54-77-77-01-02> is started, but not in primary mode. Unknown state.
 >  stderr: ocf-exit-reason:Failed initial monitor action

This happens because the credentials from ~/.my.cnf are used. Adding
the "--no-defaults" parameter to the mysql client call solves this.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -82,6 +82,13 @@ elif [ -f "/etc/default/clustercheck" ]; then
 fi
 
 #######################################################################
+# Defaults:
+
+OCF_RESKEY_check_passwd_use_empty_default=0
+
+: ${OCF_RESKEY_check_passwd_use_empty=${OCF_RESKEY_check_passwd_use_empty_default}}
+
+#######################################################################
 
 usage() {
   cat <<UEND
@@ -243,10 +250,21 @@ Cluster check user.
 
 <parameter name="check_passwd" unique="0" required="0">
 <longdesc lang="en">
-Cluster check user password
+Cluster check user password. Empty passwords are ignored unless
+the parameter "check_passwd_use_empty" is set to 1.
 </longdesc>
 <shortdesc lang="en">check password</shortdesc>
 <content type="string" default="" />
+</parameter>
+
+<parameter name="check_passwd_use_empty" unique="0" required="0">
+<longdesc lang="en">
+Use an empty "check_passwd" password. If this parameter is set to 1,
+"check_passwd" will be ignored and an empty password is used
+when calling the "mysql" client binary.
+</longdesc>
+<shortdesc lang="en">check password use empty</shortdesc>
+<content type="boolean" default="${OCF_RESKEY_check_passwd_use_empty_default}"/>
 </parameter>
 
 </parameters>
@@ -955,7 +973,10 @@ fi
 : ${OCF_RESKEY_check_user="root"}
 
 MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
-if [ -n "${OCF_RESKEY_check_passwd}" ]; then
+
+if ocf_is_true "${OCF_RESKEY_check_passwd_use_empty}"; then
+    MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password=''"
+elif [ -n "${OCF_RESKEY_check_passwd}" ]; then
     MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password=${OCF_RESKEY_check_passwd}"
 fi
 


oalbrigt:
ok to test

toabctl:
@krig could you have a look please?

heartbeat/galera
```diff
@@ -954,7 +954,7 @@ if [ -z "${OCF_RESKEY_check_user}" ]; then
 fi
 : ${OCF_RESKEY_check_user="root"}
 
-MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
+MYSQL_OPTIONS_CHECK="--no-defaults -nNE --user=${OCF_RESKEY_check_user}"
```
dciabrin: I think that would break cases where the user instantiated the resource without specifying OCF_RESKEY_check_user and OCF_RESKEY_check_passwd (or relying on /etc/sysconfig/clustercheck) and relies on the contents of /root/.my.cnf to work.
Maybe we should add an option e.g. OCF_RESKEY_no_defaults that would use that flag, and keep the default as is for backward compatibility?

krig: Hmm, yeah, adding `--no-defaults` is definitely a breaking change there..

oalbrigt:
Maybe `ignore_defaults` with description that it's used to ignore settings in `~/.my.cnf`

toabctl:
Adding an extra parameter would mean that the ``--no-defaults``` needs to be added to **all** ```mysql``` commands (including the mysql resource agent?). The problem I want to fix is only for the check command.
I played a bit more with this and with the current galera resource agent, it is not possible to have a empty `check_passwd` with a given `check_user`.
There are multiple problems:
1. if you leave `check_passwd` empty (```check_passwd=""```), the ```check_passwd``` will be overwritten with `${MYSQL_PASSWORD}` (if that is sourced from `/etc/{sysconfig,default}/clustercheck`). see https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/galera#L950
1. If you leave `check_passwd` empty **and** do not set `${MYSQL_PASSWORD}`, the password is still not because it is skipped if the length is 0 . See https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/galera#L959

So I would argue that it is currently a bug that you can not use an empty password.

krig:
I guess you could call it a bug, or a missing feature :)

toabctl:
> I guess you could call it a bug, or a missing feature :)

@krig any suggestion how to solve that? Are the resource agent parameters (`OCF_RESKEY_...`) always set as environment variables? Is there a difference between "not set" and "empty (length is zero) ?

krig:
Well, it's a feature gap in the agent. Since the parameter defaults to an empty string and this is interpreted as the password not being set, there is no way to supply an empty password -- and since this is part of the agent interface, changing this would be a breaking change.

I'm not sure what the best solution would be. The whole `--no-defaults` solution honestly sounds like a bit of a hack to me, and yeah, it'd be confusing to have an `ignore-defaults` parameter that only applies to the check call...

krig:
@toabctl I guess one question would be, why do you need to be able to pass an empty password? Is there any way that you could adapt your usage to the interface of the agent as-is?

toabctl:
> @toabctl I guess one question would be, why do you need to be able to pass an empty password? Is there any way that you could adapt your usage to the interface of the agent as-is?

Sure. We can adapt our code (which is https://github.com/crowbar/crowbar-openstack/blob/master/chef/cookbooks/mysql/recipes/ha_galera.rb#L114 ). But that might be also not that easy given that we already have mariadb users without a password.

toabctl:
@krig @dciabrin another possible solution (impelemented now in this PR) would be to add another `check_passwd_use_emtpy` parameter which can be set to fix the current bahavior. Somehow ugly to have this extra paramter but I don't see another way to fix the resource agent currently. Comments?

krig:
> @krig @dciabrin another possible solution (impelemented now in this PR) would be to add another `check_passwd_use_emtpy` parameter which can be set to fix the current bahavior. Somehow ugly to have this extra paramter but I don't see another way to fix the resource agent currently. Comments?

I think that's the most reasonable approach so far. It shouldn't interfere with any existing installations, and the scope of the new parameter is small. :+1:

heartbeat/galera
```diff
@@ -955,7 +965,7 @@ fi
 : ${OCF_RESKEY_check_user="root"}
 
 MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
-if [ -n "${OCF_RESKEY_check_passwd}" ]; then
+if [ -n "${OCF_RESKEY_check_passwd}" ] || [ [ -z "${OCF_RESKEY_check_passwd}" ] && [ "${OCF_RESKEY_check_passwd_use_empty}" = 1 ] ]; then
```
krig: The shellcheck run fails on this with these errors:

```
heartbeat/galera:969:45: error: If grouping expressions inside [..], use \( ..\). [SC1026]
heartbeat/galera:969:47: warning: Unknown binary operator. [SC2057]
```
Looking at it, it looks like you are nesting `[ [ ... ]  && [ ... ] ]` - that looks off to me.

krig:
It would also be good to add a note in the documentation for the `check_passwd` parameter that says that an empty password is ignored unless `check_passwd_use_empty` is set.

heartbeat/galera
toabctl: urgs. fixed. thanks

toabctl:
> It would also be good to add a note in the documentation for the `check_passwd` parameter that says that an empty password is ignored unless `check_passwd_use_empty` is set.

@krig done. Hopefully you meant the parameter doc (and not README.galera) ?

heartbeat/galera
```diff
@@ -955,7 +966,7 @@ fi
 : ${OCF_RESKEY_check_user="root"}
 
 MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
-if [ -n "${OCF_RESKEY_check_passwd}" ]; then
+if [ -n "${OCF_RESKEY_check_passwd}" -o \( -z "${OCF_RESKEY_check_passwd}" -a "${OCF_RESKEY_check_passwd_use_empty}" = 1 \) ]; then
```
oalbrigt: I think it'd be better to do:
`if ! ocf_is_true "${OCF_RESKEY_check_passwd_use_empty}" && [ -n "${OCF_RESKEY_check_passwd}" ]; then`
here, and undo the other if-change above that wont be needed.

ocf_is_true returns true/false for 1/0, true/false, yes/no values.

heartbeat/galera
```diff
@@ -243,12 +243,23 @@ Cluster check user.
...
 </parameter>
 
+<parameter name="check_passwd_use_empty" unique="0" required="0">
+<longdesc lang="en">
+Use an empty "check_passwd" password. If this parameter is set to 1
+and "check_passwd" is an empty string, an empty password is used
+when calling the "mysql" client binary.
+</longdesc>
+<shortdesc lang="en">check password use empty</shortdesc>
+<content type="boolean" default="0"/>
```
oalbrigt: This should be assigned a `OCF_RESKEY_check_passwd_use_empty_default` value to be used for `default=` here, and assigned to `OCF_RESKEY_check_passwd_use_empty`:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/nagios#L19-L36

heartbeat/galera
toabctl: So if `check_passwd` is set **and** `check_passwd_use_empty` is `1 (or yes or true)`, `check_passwd_use_empty` should have precedence? Current implementation is the other way arround...

oalbrigt: Presedence doesnt matter when it's only an AND.

heartbeat/galera
toabctl: Done

heartbeat/galera
toabctl: But your check seems to be wrong. I updated the code and I think that should be ok now.

heartbeat/galera
```diff
@@ -955,8 +974,13 @@ fi
 : ${OCF_RESKEY_check_user="root"}
 
 MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
-if [ -n "${OCF_RESKEY_check_passwd}" ]; then
-    MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password=${OCF_RESKEY_check_passwd}"
+
+if ocf_is_true "${OCF_RESKEY_check_passwd_use_empty}"; then
+    MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password=''"
+else
```
oalbrigt: You can use `elif` here.

toabctl: Why should I use `elif` here?

oalbrigt: instead of else followed by if/fi.

heartbeat/galera
```diff
@@ -243,12 +250,23 @@ Cluster check user.
...
 </parameter>
 
+<parameter name="check_passwd_use_empty" unique="0" required="0">
+<longdesc lang="en">
+Use an empty "check_passwd" password. If this parameter is set to 1,
+"check_passwd" will be ignored and an empty password is used
+when calling the "mysql" client binary.
+</longdesc>
+<shortdesc lang="en">check password use empty</shortdesc>
+<content type="boolean" default="${check_passwd_use_empty_default}"/>
```
oalbrigt: You need `OCF_RESKEY_` here as well.

toabctl: fixed

heartbeat/galera
toabctl: urgs. true. done

oalbrigt:
LGTM.

krig:
Looks good to me, thanks!

Issue #1325, opened 2019-04-25
Filesystem: umount does not execute due to readlink failure and remains mounted


yuusuke:
Hi

We have mounted shared disk using Filesystem of resource-agents-4.2.0.
The file system uses xfs.

When an FC connection with a shared disk fails, Filesystem stop succeeds, but umount is not executed and the mount remains.

When I analyzed it, I found that readlink failed and returned an empty string when FC connection failed.
```
# readlink -f /dbfp/pgdata
# echo $?
1
```

With the -v option, you can see that an Input / output error has occurred.
```
# readlink -v -f /dbfp/pgdata
readlink: /dbfp/pgdata: Input/output error
# echo $?
1
```
How do I fix this problem?

Attach crm_report for 2 nodes.
~~[pcmk-Thu-25-Apr-2019.zip](https://github.com/ClusterLabs/resource-agents/files/3115922/pcmk-Thu-25-Apr-2019.zip)~~
~~[pcmk-Thu-25-Apr-2019.zip](https://github.com/ClusterLabs/resource-agents/files/3115939/pcmk-Thu-25-Apr-2019.zip)~~
I'm sorry.
I made a mistake in attaching crm_report.
The correct report is here.
[pcmk-Fri-26-Apr-2019.zip](https://github.com/ClusterLabs/resource-agents/files/3119435/pcmk-Fri-26-Apr-2019.zip)
[pcmk-Fri-26-Apr-2019.zip](https://github.com/ClusterLabs/resource-agents/files/3119436/pcmk-Fri-26-Apr-2019.zip)

PR #1332, opened 2019-05-16
Low: Filesystem: Fix a problem umount is not executed in the event of a disk failure

yuusuke:
This corrects the problem reported in #1325.
The problem is that if readlink fails and the output is empty, it will be mistaken as umounted.


oalbrigt:
That's weird. I'll see if we can test all these scenarioes with this patch.

oalbrigt:
ok to test

oalbrigt:
LGTM.

krig:
LGTM

yuusuke:
Thank you for merging!

Issue #1333, opened 2019-05-16
--logging/--logfile version string comparison fails to handle '-VENDOR trailer'


noelpower:
commit 4a31eb72ad007dfea77b22ef14f1cb766ce799e0 should handle a -VENDOR trailer
      + e.g. 4.0.0-VendorVersion

however when using the sed from 4a31eb72ad007dfea77b22ef14f1cb766ce799e0 on a version string with vendor trailer e.g.

 echo "4.0.0-VendorVersion" | awk '{print $NF}' | sed "s/\.\?[[:alpha:]].*//"

yields

 4.0.0-

PR #1334, opened 2019-05-16
CTDB: fix version string with vendor trailer comparison

noelpower:
commit 4a31eb7 should handle a -VENDOR trailer
+ e.g. 4.0.0-VendorVersion

however when using the sed from 4a31eb7 on a version string with vendor trailer e.g.

echo "4.0.0-VendorVersion" | awk '{print $NF}' | sed "s/.?[[:alpha:]].*//"

yields

4.0.0-

after fix

echo "4.0.0rc1" | awk '{print $NF}' | sed "s/[-\.]\?[[:alpha:]].*//"
4.0.0

echo "4.0.0-VendorVersion" | awk '{print $NF}' | sed "s/[-\.]\?[[:alpha:]].*//"
4.0.0

echo "4.0.0.GIT.1a2b3c4d" | awk '{print $NF}' | sed "s/[-\.]\?[[:alpha:]].*//"
4.0.0

Fixes: https://github.com/ClusterLabs/resource-agents/issues/1333

Signed-off-by: Noel Power <noel.power@suse.com>

### DIFF
--- a/heartbeat/CTDB.in
+++ b/heartbeat/CTDB.in
@@ -588,7 +588,7 @@ ctdb_start() {
 	local log_option
 	# --logging supported from v4.3.0 and --logfile / --syslog support 
 	# has been removed from newer versions
-	version=$(ctdb version | awk '{print $NF}' | sed "s/\.\?[[:alpha:]].*//")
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM.

oalbrigt:
ok to test

ddiss:
LGTM too, thanks Noel.

PR #1336, opened 2019-05-16
Fails docker RA gracefully when command not found or daemon is not running

zaenk:
Using the docker RA, when a docker resource is configured then the cluster reports the following error from all nodes without docker:

```
* docker-tutorial1_stop_0 on mgr 'unknown error' (1): call=253, status=complete, exitreason='/usr/lib/ocf/resource.d/heartbeat/docker:241: /usr/lib/ocf/resource.d/heartbeat/docker: docker: not found',
    last-rc-change='Thu May 16 12:56:14 2019', queued=0ms, exec=35ms
```

This is caused by `container_exists` which does not catch this error when running `docker inspect`.

Other RAs (e.g.: mysql, jboss) fail more gracefully on nodes where the commands are not available.

This patch adds 2 more checks to the monitor action:
* Checks if the `docker` command is available on the node. If not, fails with `OCF_ERR_INSTALLED`
* Checks if the docker daemon is runnig by checking the pid file (similar to what the mysql RA does). If not, then fails with `OCF_NOT_RUNNING`

This patch introduces a new optional parameter:
* `daemon_pidfile` which defaults to `/var/run/docker.pid`

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -33,6 +33,11 @@
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
+# Parameter defaults
+
+OCF_RESKEY_daemon_pidfile_default="/var/run/docker.pid"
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

krig:
Looks good to me.

PR #1338, opened 2019-05-17
Add support for LIO-T incoming CHAP auth for TPG

mattinx:
Existing model allows incoming CHAP authentication only when specifying a list of initiators, and sets the auth attribute for each initiator
Add support to set CHAP username/password at target portal group level (tpg auth attributes) if no list of initiators is provided

### DIFF
--- a/heartbeat/iSCSITarget.in
+++ b/heartbeat/iSCSITarget.in
@@ -390,7 +390,13 @@ iSCSITarget_start() {
 				fi
 			done
 		else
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
LGTM.

chrboe:
Is there anything blocking this from being merged?

oalbrigt:
Thanks for the ping. I obviously forgot about it.

PR #1339, opened 2019-05-17
Add option to override LIO backstore name

mattinx:
By default, the LIO backing store is named after the resource name. When this backing store is mapped to a LUN, the SCSI INQUIRY reponse is derived from this truncated to the first 16 characters. As it is common to name resources within a group with a similar prefix, the first 16 characters are often less useful than those later on
Add an backstore_name option to allow a custom name to be specified, making it a little easier to identify LUNs on initiators


oalbrigt:
ok to test

oalbrigt:
LGTM.

PR #1340, opened 2019-05-21
Add resource agent for dovecot.

sebastic:
Based on the postfix resource agent.

We're using it at $DAYJOB for about a year now.

### DIFF
--- a/heartbeat/dovecot
+++ b/heartbeat/dovecot
@@ -1,0 +1,333 @@
+#!/bin/sh
+#
+# Resource script for Dovecot
+#
+# Description:  Manages Dovecot as an OCF resource in
+#               an high-availability setup.
+#
+# Author:       Raoul Bhatia <r.bhatia@ipax.at> : Original Author
+# License:      GNU General Public License (GPL)
+#
+#
+#       usage: $0 {start|stop|reload|monitor|validate-all|meta-data}
+#
+#       The "start" arg starts a Dovecot instance
+#
+#       The "stop" arg stops it.
+#
+# OCF parameters:
+#  OCF_RESKEY_binary
+#  OCF_RESKEY_config_file
+#  OCF_RESKEY_parameters
+#
+##########################################################################
+
+# Initialization:
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+: ${OCF_RESKEY_binary="/usr/sbin/dovecot"}
+: ${OCF_RESKEY_config_file=""}
+: ${OCF_RESKEY_pid_file=""}
+: ${OCF_RESKEY_parameters=""}
+USAGE="Usage: $0 {start|stop|reload|monitor|validate-all|meta-data}";
+
+##########################################################################
+
+usage() {
+    echo $USAGE >&2
+}
+
+meta_data() {
+        cat <<END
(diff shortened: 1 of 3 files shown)


heartbeat/dovecot
```diff
@@ -0,0 +1,333 @@
...
+
+if [ $ret -eq 0 ]; then
+    status_support=true
+fi
+
+
+dovecot_validate_all
+ret=$?
+
+LSB_STATUS_STOPPED=3
```
krig: This doesn't seem to be used?

heartbeat/dovecot
```diff
@@ -0,0 +1,333 @@
...
+        fi
+        return $ret
+    fi
+
+    # manually check Dovecot's pid
+
+    PIDFILE=$OCF_RESKEY_pid_file
+    if [ -f $PIDFILE ]; then
+         PID=`head -n 1 $PIDFILE`
+         kill -s 0 $PID >/dev/null 2>&1 && [ `ps -p $PID | grep dovecot | wc -l` -eq 1 ]
```
krig: You could probably use `pgrep` instead of `ps .. | grep ..` here

heartbeat/dovecot
sebastic: Good catch, the postfix RA has the same issue.

heartbeat/dovecot
sebastic: Not sure how portable that is.

Like the issue above, it came from the postfix RA.

heartbeat/dovecot
```diff
@@ -0,0 +1,333 @@
...
+
+<parameters>
+
+<parameter name="binary" unique="0" required="0">
+<longdesc lang="en">
+Full path to the Dovecot binary.
+For example, "/usr/sbin/dovecot".
+</longdesc>
+<shortdesc lang="en">Full path to Dovecot binary</shortdesc>
+<content type="string" default="/usr/sbin/dovecot" />
```
oalbrigt: You should use `_default` parameters in initialization, and reuse them here to avoid having to update it several places in the future.

Example:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/VirtualDomain#L20

oalbrigt:
ok to test

heartbeat/dovecot
sebastic: Like the other issues, it came from the postfix RA which doesn't use `_default` variables either.

oalbrigt: It's still a good idea to implement it with `_default` variables, and not add bad code practice in new agents.

heartbeat/dovecot
```diff
@@ -0,0 +1,335 @@
...
+
+: ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
+. ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
+
+# Defaults
+OCF_RESKEY_binary_default="/usr/sbin/dovecot"
+
+: ${OCF_RESKEY_binary=${OCF_RESKEY_binary_default}}
+: ${OCF_RESKEY_config_file=""}
+: ${OCF_RESKEY_pid_file=""}
```
oalbrigt: It seems like you should also use a `_default` variable for thisone, as it's empty by default, while the metadata shows a path it wont use by default.

oalbrigt:
Looks good to me now... @krig you're welcome to merge if you're OK with it now.

krig:
Alright, merging :)

PR #1341, opened 2019-05-21
Add systemctl resource agent to manage systemd services.

sebastic:
An alternative to LSB resource agents.

We're using it at $DAYJOB for about a year now.


oalbrigt:
Pacemaker supports running systemd services as "agents", so what's the point in using this agent instead?

sebastic:
It's an alternative.

I developed it back then because I didn't find documentation how to use systemd services with pacemaker.

That documentation is still hard to find.

oalbrigt:
@kgaillot Sounds like we should consider making the systemd documentation easier to find (and maybe mention it where appropriate).

kpfleming:
Agreed! I'm using the systemd RA and it works fine, but there is literally zero documentation available for it.

kgaillot:
There's a brief mention in "Pacemaker Explained", but it could definitely be improved.

PR #1342, opened 2019-05-21
galera: Log message when changing content of grastate.dat file

toabctl:
There is one case in the RA which changes the grastate.dat file for
galera. Write a log message when the file is changed.
This is useful when trying to find out why "safe_to_bootstrap" was
set.


toabctl:
@krig could you have a look please?

oalbrigt:
ok to test

krig:
Makes sense to me,thanks!

PR #1352, opened 2019-06-12
dhcpd: keep SELinux context

oalbrigt:


phagara:
I think this simpler patch should work too:

```diff
diff --git a/heartbeat/dhcpd b/heartbeat/dhcpd
index 8b2d8b69..75b99923 100755
--- a/heartbeat/dhcpd
+++ b/heartbeat/dhcpd
@@ -342,7 +342,11 @@ dhcpd_initialize_chroot() {
        fi
     done
 
-   return $OCF_SUCCESS
+    if selinuxenabled; then
+        chcon --reference "/${libdir}" -R "${OCF_RESKEY_chrooted_path}/${libdir}"
+    fi
+
+    return $OCF_SUCCESS
 }
 
 # Initialize a non-chroot environment
```

oalbrigt:
It also needs to work on systems without the SELinux binaries.

oalbrigt:
Updated from -pL to -aL, as -a copies SELinux context when SELinux is used.

oalbrigt:
Thanks @phagara for the suggestion.

oalbrigt:
Moving CI updates to separate PR: https://github.com/ClusterLabs/resource-agents/pull/1353

PR #1353, opened 2019-06-12
CI: shellsheck fixes

oalbrigt:


krig:
hmm, shouldn't BASH_SHELL files still be checked by shellcheck?

krig:
although as bash scripts, perhaps.

oalbrigt:
Correct. I realized they are already created before testing.

krig:
ah, I see

PR #1354, opened 2019-06-13
Podman fixes 

mbaldessari:
This is a three patch series to improve the podman RA. I tested them as per commit-messages description


oalbrigt:
LGTM.

PR #1355, opened 2019-06-13
azure-events: change message log level for the RA non action messages

diegoakechi:
Reduces the verbosity on the log when the RA has no events to process.
The messages can still be seen using the verbose parameter.


krig:
LGTM

PR #1359, opened 2019-06-21
aws-vpc-route53.in: Add missing quotes in ec2ip_stop()

deacomm:
If AWS API returns an error for some reason (e.g. API unavailable or insufficient IAM rights), the stop function in aws-vpc-route53 resource agent will have a syntax error on line 254, because the ARECORD variable will be empty.

Adding quotes around the variables will prevent that. This is how the if condition is written in other places within the script e.g. in ec2ip_monitor() function.


oalbrigt:
LGTM. Thanks.

PR #1360, opened 2019-06-21
aws-vpc-route53 - Improvements on monitor function and general changes to the code

gguifelixamz:
This pull request is to add improvements to this agent's monitor function and general code cleanup.

Currently, the monitor function uses AWS CLI to call "list-resource-record-sets" AWS API call, and if the AWS CLI fails for any reason (ex.: throttling, network errors, etc) it may cause the cluster to failover because there is not exception handling for the AWS CLI commands.

Route53 API calls have a limit of five API calls / second per account [1] it is desirable to avoid doing API calls, so I added a "dig" DNS check and only in case of a DNS lookup failure from the "dig" DNS check the monitor function will fallback and try to execute the "list-resource-record-sets" API call for monitoring.

I also changed the way the EC2 instance's IP address it checked, instead of using a 3rd party script (ec2metadata), which may not be default for all distros, I'm using "curl" to get the IP address from the local EC2 metadata server.

Let me know if you have questions about the changes.

Reference:
[1] https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/DNSLimitations.html#limits-api-requests

### DIFF
--- a/heartbeat/aws-vpc-route53.in
+++ b/heartbeat/aws-vpc-route53.in
@@ -190,17 +190,68 @@ r53_validate() {
 }
 
 r53_monitor() {
+	#
+	# For every start action the agent  will call Route53 API to check for DNS record
+	# otherwise it will try to get results directly bu querying the DNS using "dig".
+	# Due to complexity in some DNS architectures "dig" can fail, and if this happens
+	# the monitor will fallback to the Route53 API call.
+	#
+	# There will be no failure, failover or restart of the agent if the monitor operation fails
+	# hence we only return $OCF_SUCESS in this function
+	#
+	# In case of the monitor operation detects a wrong or non-existent Route53 DNS entry
+	# it will try to fix the existing one, or create it again
+	#
+	#
+	ARECORD=""
+	IPREGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$"
 	r53_validate
 	ocf_log debug "Checking Route53 record sets"
+	#
 	IPADDRESS="$(ec2metadata aws ip | grep local-ipv4 | /usr/bin/awk '{ print $2 }')"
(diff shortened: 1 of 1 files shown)


oalbrigt:
Based on https://github.com/ClusterLabs/resource-agents/pull/1359 you might want to use quotes in the ARECORD / IPADRESS if's.

gguifelixamz:
> Based on #1359 you might want to use quotes in the ARECORD / IPADRESS if's.

The unquoted ifs are for REGEX comparisons where quotes are not required.

Quick example:

```
#!/bin/bash

IPREGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$"

if  [[ $1 =~ $IPREGEX ]]; then
   echo "$1 is an IP address"
else
   echo "Could not match REGEX - probably not an IP address"
fi

```

Results:

```
# ./test.sh 1.2.3.4
1.2.3.4 is an IP address
```

Adding double quotes:

```
#!/bin/bash

IPREGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$"

if  [[ "$1" =~ "$IPREGEX" ]]; then
   echo "$1 is an IP address"
else
   echo "Could not match REGEX - probably not an IP address"
fi
```

Result:

```
# ./test.sh 1.2.3.4
Could not match REGEX - probably not an IP address
```

Leaving $REGEX unquoted also works:

```
#!/bin/bash

IPREGEX="^[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}$"

if  [[ "$1" =~ $IPREGEX ]]; then
   echo "$1 is an IP address"
else
   echo "Could not match REGEX - probably not an IP address"
fi
```

Result:

```
# ./test.sh 1.2.3.4
1.2.3.4 is an IP address
```

oalbrigt:
Perfect. Then everything looks good to me.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1362, opened 2019-06-24
Medium: Filesystem: Prevents to all root user processes are killed wh…

ytakeshita:
…en bind mounting a directory on rootfs.

if a directory is bound mounting on rootfs and "force_umount" is not set "safe", change "force_umount" to "safe".

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -314,6 +314,24 @@ bind_kernel_check() {
 	[ $? -ne 0 ] &&
 		ocf_log warn "kernel `uname -r` cannot handle read only bind mounts"
 }
+
+bind_rootfs_check() {
+        local SOURCE
+        local TARGET
+        local ROOTFS
+
+        SOURCE=$1
+        TARGET=$(df --output=target $SOURCE | tail -n 1)
+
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

jnpkrn:
Not disputing the fix as such, only the very limited usability, i.e.,
incomplete coverage of the problem surface.

In my experience, there is (perhaps usually?) no rootfs on the system,
more so when LVM is used as a means to partitioning.

Moreover, the presented solution is not POSIX compatible, which is bad
enough alone.

What's wrong with something like (untested, but even simpler!):

```
if test "$(df -P "$1"  | tr -s ' ' | cut -d\  -f6)" = /; then 
    ...
fi
```

?

Please, pay more attentions to concerns like these (addressed especially
to @oalbrigt).

oalbrigt:
@jnpkrn I've created a PR to solve the POSIX issue based on your suggestion: https://github.com/ClusterLabs/resource-agents/pull/1394

I dont think there is a good way to be able to know 100% if another filesystem is mounted correctly (correct device -> mount point) though.

PR #1365, opened 2019-07-10
[apache-RA]: Improve PidFile pattern

MalloZup:
take resource ID as a pattern for PIDFile


MalloZup:
@gao-yan CC. This PR is nothing special, you might know about it. For any info just feel free to ping me.

krig:
Looks good to me!

gao-yan:
Nice work. BTW, the change is to let multiple apache resources/instances have their own pid files so that they are correctly tracked.

Thanks, @MalloZup and @krig !

PR #1368, opened 2019-07-15
Load apache status modules from RA cmdline for SUSE only OS

MalloZup:
Load apache status modules from RA cmdline for SUSE only OS

### DIFF
--- a/heartbeat/apache
+++ b/heartbeat/apache
@@ -197,6 +197,14 @@ validate_default_suse_config() {
 			ocf_run -q /usr/sbin/start_apache2 -t || return 1
 		fi
 	fi
+
+        # load module only if module exists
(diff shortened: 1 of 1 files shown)


krig:
I think this looks good to me, but maybe @oalbrigt wants to have look as well. Could this cause issues for someone who wants to implement their own status API and not use `mod_status`, though? At least in theory, I could imagine that there are existing users who don't use `mod_status`..

MalloZup:
@krig  right. I have added an additional check for last comment. Thx for comments :couch_and_lamp: :+1:

oalbrigt:
ok to test

oalbrigt:
LGTM

MalloZup:
5hf @krig feel fre to merge

PR #1372, opened 2019-07-30
changing bashism to generic solution

synq:
The VAR=(< filename) syntax does not work in all situations (i.e. not on debian 10), changing it to VAR=(cat filename) does work.

Found this problem because I am using debian 10, will create a debian bug report that points to this request.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1374, opened 2019-08-01
Generate addition drop-in dependencies for podman containers

dciabrin:
When podman creates a container, it creates two additional systemd
scope files dynamically:

  - libpod-conmon-<CONTAINERID>.scope - runs a conmon process that
    tracks a container's pid1 into a dedicated pidfile.
  - libpod-<CONTAINERID>.scope - created dynamically by runc,
    for cgroups accounting

On shutdown, it can happen that systemd stops those scope early,
which in turn sends a SIGTERM to pacemaker-managed containers
before pacemaker has scheduled any stop operation. That
confuses the cluster and may break shutdown.

Add a new option in the resource-agent to inject additional
dependencies into the dynamically created scope files, so that
systemd is not allowed to stop scopes before the pacemaker
service itself is stopped.

When that option is enabled, the scopes look like:

  # podman ps | grep galera
  c329819a1227  192.168.122.8:8787/rhosp15/openstack-mariadb:latest                     dumb-init -- /bin...  About an hour ago  Up About an hour ago         galera-bundle-podman-0

  # systemctl cat libpod*c329819a1227*
  # /run/systemd/transient/libpod-conmon-c329819a1227ec548d678861994ef755b1fde9a244e1e4d966d17674df88ce7b.scope
  # This is a transient unit file, created programmatically via the systemd API. Do not edit.
  [Scope]
  Slice=machine.slice
  Delegate=yes

  [Unit]
  DefaultDependencies=no

  # /run/systemd/transient/libpod-conmon-c329819a1227ec548d678861994ef755b1fde9a244e1e4d966d17674df88ce7b.scope.d/dep.conf
  [Unit]
  Before=pacemaker.service

  # /run/systemd/transient/libpod-c329819a1227ec548d678861994ef755b1fde9a244e1e4d966d17674df88ce7b.scope
  # This is a transient unit file, created programmatically via the systemd API. Do not edit.
  [Unit]
  Description=libcontainer container c329819a1227ec548d678861994ef755b1fde9a244e1e4d966d17674df88ce7b

  [Scope]
  Slice=machine.slice
  Delegate=yes
  MemoryAccounting=yes
  CPUAccounting=yes
  BlockIOAccounting=yes

  [Unit]
  DefaultDependencies=no

  # /run/systemd/transient/libpod-c329819a1227ec548d678861994ef755b1fde9a244e1e4d966d17674df88ce7b.scope.d/dep.conf
  [Unit]
  Before=pacemaker.service

Effectively, this prevents systemd from managing the shutdown of any
pacemaker-managed podman container.

Related: rhbz#1726442

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -158,6 +158,16 @@ to have the particular one persist when this happens.
 <shortdesc lang="en">reuse container</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_reuse_default}"/>
 </parameter>
+
+<parameter name="drop_in_dependency" required="0" unique="0">
+<longdesc lang="en">
+Use transient drop-in files to add extra dependencies to the systemd
+scopes associated to the container. During reboot, this prevents systemd
+to stop the container before pacemaker.
+</longdesc>
+<shortdesc lang="en">drop-in dependency</shortdesc>
+<content type="boolean"/>
+</parameter>
 </parameters>
 
 <actions>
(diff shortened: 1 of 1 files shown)


oalbrigt:
add to whitelist

oalbrigt:
LGTM. Thanks.

PR #1375, opened 2019-08-08
Filesystem: don't call readlink on path if it does not exist

### LINKED ISSUE #1370: Filesystem actions fail if mountpoint does not exist
PR #1332 seems to have broken status checks in the Filesystem resource if the parent directory of a mountpoint does not exist.

Previously, the status would determine that the filesystem was unmounted if a readlink operation on the mountpoint would fail. On start, the resource would create the directory before mounting it.

Now that the readlink is performed before any actions are processed, the script returns an error when the status is checked. This results in errors like this:

  * hass_fs_monitor_0 on dx1 'unknown error' (1): call=82, status=complete, exitreason='Could not canonicalize /var/lib/machines/hass because readlink failed',

This has a disastrous consequence. If a new node is brought up, as I did earlier today, the status will be checked for each Filesystem resource on the node. Since it fails, and is unable to get the actual status, the resource is taken offline on *all* other nodes. This results in all resources depending on a Filesystem resource going offline.

Since all of my resources depend on a Filesystem resource, this has taken my entire cluster out unexpectedly.

jimfunk:
This fixes #1370

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -858,10 +858,14 @@ if [ -z "$OCF_RESKEY_directory" ]; then
 else
 	MOUNTPOINT=$(echo $OCF_RESKEY_directory | sed 's/\/*$//')
 	: ${MOUNTPOINT:=/}
-	CANONICALIZED_MOUNTPOINT=$(readlink -f "$MOUNTPOINT")
-	if [ $? -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1376, opened 2019-08-08
Don't try to log in as mysql with a login shell

heyakyra:
The mysql user doesn't have a usable home directory.


oalbrigt:
This would make users unable to customize PATH and other things for the mysql user (by adding the "missing" shell files and editing them).

heyakyra:
> This would make users unable to customize PATH and other things for the mysql user (by adding the "missing" shell files and editing them).

Fair enough, so the expected workaround is to just create a usable home directory

Issue #1377, opened 2019-08-09
Node which failed replication can be mistakenly promoted to master because MySQL resource doesn't report the failure


heyakyra:
We've encountered an issue where replication failed, but the resource agent simply reported success to Pacemaker, which meant that we had no idea one of the masters was out of date, and we then failed over to an out of date master which caused divergent database and customer downtime.

```
Sep  5 19:21:22 es-lnsp-flow-admin1 mysql(editshare.stack.3927dac6-a628-4485-a69d-f2934ea0b94c.mysql.primitive)[719]: ERROR: MySQL instance configured for replication, but replication has failed.
Sep  5 19:21:22 es-lnsp-flow-admin1 mysql(editshare.stack.3927dac6-a628-4485-a69d-f2934ea0b94c.mysql.primitive)[719]: ERROR: See  for details
Sep  5 19:21:22 es-lnsp-flow-admin1 lrmd[1528]:   notice: operation_finished: editshare.stack.3927dac6-a628-4485-a69d-f2934ea0b94c.mysql.primitive_monitor_10000:719:stderr [ ocf-exit-reason:MySQL instance configured for replication, but replication has failed. ]
```

This happens because when we the resource agent runs into almost any error on the slave, it just sets the "reader" attribute to 0 (which indicates "read only" IP resources should migrate away), but doesn't do anything to prevent the node from being promoted due to out of date data:

In https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/mysql#L418 
```
        if [ $last_errno -ne 0 -a $last_errno -ne "$MYSQL_TOO_MANY_CONN_ERR" ]; then
            # Whoa. Replication ran into an error. This slave has
            # diverged from its master. Make sure this resource
            # doesn't restart in place.
            ocf_exit_reason "MySQL instance configured for replication, but replication has failed." 
            ocf_log err "See $tmpfile for details" 

            # Just pull the reader VIP away, killing MySQL here would be pretty evil
            # on a loaded server

            set_reader_attr 0
            exit $OCF_SUCCESS

        fi
```

It would be better to properly warn about a degraded replica in this case, and disallow or strongly discourage failing over to such an out of date replica. If this happens, and failover occurs, you now have divergent databases that can't be merged easily, and must choose to lose one side or other.

We're attaching a simple pull request we've been running with for a couple of years for review, and are curious to know if this is an acceptable solution or if there is a better way of addressing this.

PR #1378, opened 2019-08-09
Exit with failure on MySQL replication failure

heyakyra:
When we the resource agent runs into almost any error on the slave, it just
sets the "reader" attribute to 0 (which indicates "read only" IP resources
should migrate away), but doesn't do anything to prevent the node from being
promoted due to out of date data.

We've encountered an issue where replication failed, but the resource agent
simply reported success to Pacemaker, which meant that we had no idea one of
the masters was out of date, and we then failed over to an out of date master
which caused divergent database and customer downtime.

Change the return value of `check_slave()` from $OCF_SUCCESS to
$OCF_ERR_GENERIC, to denote failure with MySQL replication.

Addresses #1377

### DIFF
--- a/heartbeat/mysql
+++ b/heartbeat/mysql
@@ -422,11 +422,11 @@ check_slave() {
             ocf_exit_reason "MySQL instance configured for replication, but replication has failed."
             ocf_log err "See $tmpfile for details"
 
-            # Just pull the reader VIP away, killing MySQL here would be pretty evil
-            # on a loaded server
+            # Just pull the reader VIP away and report the error. Even though killing MySQL evil on 
+            # a loaded server, having a split database is worse!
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
It seems like you should create a VIP resource with the IPaddr2 agent and set colocation restraints based on the attribute mysql manipulates:
https://github.com/Percona-Lab/pacemaker-replication-agents/blob/master/doc/PRM-setup-guide.rst#reader-vip-location-rules

fabbione:
@heyakyra apparently github has closed this issue when renaming master branch to main. I am unable to reopen the PR for you unfortunately, as you need to change the target branch from master to main.

PR #1379, opened 2019-08-12
mysql/mariadb/galera: use runuser/su to avoid using SELinux DAC_OVERRIDE

oalbrigt:


oalbrigt:
Tested with MariaDB and Galera without any issues.

PR #1383, opened 2019-08-16
mysql: add support for SSL replication

igolman:
Committer: Ivan G.

Hello,

just added an SSL-replication for own needs and wanted to share it.

**_Prerequisites:_**

- the certificate files are available on all nodes.

**_Working example:_**

```
pcs resource create mysql ocf:heartbeat:mysql replication_user="repl_user" replication_passwd="randompassword" test_user="test" test_passwd="anotherpassword" replication_master_ssl_ca="/etc/mysql/mysql-ca-cert.pem" replication_master_ssl_cert="/etc/mysql/mysql-client-cert.pem" replication_master_ssl_key="/etc/mysql/mysql-client-key.pem" meta notify="true"
```

Cheers!


heartbeat/mysql
```diff
@@ -544,14 +568,20 @@ set_master() {
     # Informs the MySQL server of the master to replicate
     # from. Accepts one mandatory argument which must contain the host
     # name of the new master host. The master must either be unchanged
-    # from the laste master the slave replicated from, or freshly
+    # from the last master the slave replicated from, or freshly
     # reset with RESET MASTER.
+    if [ -n "$OCF_RESKEY_replication_master_ssl_ca" -a -n "$OCF_RESKEY_replication_master_ssl_cert" -a -n "$OCF_RESKEY_replication_master_ssl_key" ]; then
```
oalbrigt: You should use `&&` instead of `-a`:
https://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o

heartbeat/mysql
```diff
@@ -544,14 +568,20 @@ set_master() {
...
+        master_ssl_params=", MASTER_SSL=1, \
```
oalbrigt: This parameter should be defined as an empty parameter before the `if` to avoid any issues undefined variables might cause.

heartbeat/mysql
igolman: Hi ,

> You should use `&&` instead of `-a`:
Damn right... However, I was following the `Bash` style of this file, where the `&&` was used just once in `IF` clause...  Everywhere else is the `-a` in use..

Should I check the whole file and make appropriate changes?  Although, this is a bit out of the scope of this ticket...

Looking for your feedback.

heartbeat/mysql
igolman: Correct, will be updated.

heartbeat/mysql
oalbrigt: No need to do that, unless you find it causing issues.

We just try to avoid introducing more of it, and fix it if it's causing issues.

oalbrigt:
ok to test

oalbrigt:
LGTM.

igolman:
Thanks, looking forward for upcoming release.

PR #1388, opened 2019-08-28
Make the check for the docker daemon being up more robust

mbaldessari:
This amends 5941b98140b09e39b4dc2ee155817b287ef32859 (Fails docker RA
gracefully when command not found). That commit checked for a pidfile
which tends to be less robust in the presence of stale pidfiles and
also adds a configuration option for the pidfile location which is
more churn than needed to simply check for a service availability.

Let's simply call 'docker version'. When that commands returns 1 the docker
daemon is not running and also return OCF_ERR_GENERIC instead of
OCF_NOT_RUNNING. This is a key point because if the docker daemon
is stopped and not running it can very well be that the containers
are still up (e.g. when you use live-restore in docker). In this
situation we want an explicit fence event to be triggered due to
the failure of stopping.

Not doing so would mean that the stop operation returned ok and
for example we'd be starting an A/P resource on a second node all
the while it was still running on the node there the docker daemon
was stopped.

We also explicitely catch OCF_ERR_GENERIC in the docker_stop function
to make our intent clearer.

Tested this in an Openstack deployment and observed the following:
A) All the usual pcmk operations still correctly work
B) A 'systemctl stop docker' will eventually trigger a fence operation
   on the node.

Co-Authored-By: Luca Miccini <lmiccini@redhat.com>

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -36,10 +36,8 @@
 # Parameter defaults
 
 OCF_RESKEY_reuse_default="0"
-OCF_RESKEY_daemon_pidfile_default="/var/run/docker.pid"
 
 : ${OCF_RESKEY_reuse=${OCF_RESKEY_reuse_default}}
-: ${OCF_RESKEY_daemon_pidfile=${OCF_RESKEY_daemon_pidfile_default}}
 
 #######################################################################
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

PR #1390, opened 2019-09-03
exportfs: doc clarification for clientspec format

zzhou1:
Point out the standard of the format is aligned with `man exportfs`, and also point out the correct way to deal with the use case to export the same directory(-ies) to multiple subnets.


oalbrigt:
LGTM. Thanks.

PR #1392, opened 2019-09-06
rabbitmq-cluster: also restore users/perms/policies when starting in single node mode

lemenkov:
See https://bugzilla.redhat.com/1744467#c1

Signed-off-by: Peter Lemenkov <lemenkov@gmail.com>


oalbrigt:
add to whitelist

oalbrigt:
LGTM. Thanks.

PR #1393, opened 2019-09-09
bsc#1138281  apache resource agent is not working when I set statusurl value

varkoly:
/usr/lib64/apache2-prefork/mod_status.so is a symlink. Thats why -f fails everytime. We have to use -e


oalbrigt:
LGTM. Thanks.

PR #1394, opened 2019-09-10
Filesystem: improve "/" check for bind mounts

oalbrigt:


jnpkrn:
Without actual testing, LGTM, thanks.

Issue #1357, opened 2019-06-20
IPv6 notation for ocf:heartbeat:IPaddr2


rajubi1981:
Hi,

We are trying to add a resource to a group. For us only the short IPv6 notation works.

fd5d:d50a:8c17:2040::251

pcs resource create MNGT_VIP ocf:heartbeat:IPaddr2 ip=fd5d:d50a:8c17:2040::251 op monitor interval=10s --group MNGT

But if we try the below IPv6 notation

fd5d:d50a:8c17:2040:0000:0000:0000:0251
fd5d:d50a:8c17:2040:0:0:0:0251

And same with Upper Case notation.

We wanted to know is there any restriction from pcs package for specific IPv6 notation.

Can you please help.

oalbrigt:
It has to match the format the `ip` command shows it as after adding it, so you could add it manually to see how it's represented.

You can also disable the resource and run `pcs resource debug-start --full MNGT_VIP` and similar for monitor and stop to see why it fails/what it's comparing to.

rajubi1981:
pcs resource debug-start --full MNGT_VIP

 >  stderr: ++ 19:48:57: run_send_ua:855: ip -o -f inet6 addr show dev eth0 to fd5d:d50a:8c17:2040:0000:0000:0000:0250/64
 >  stderr: + 19:48:57: run_send_ua:855: ipstatus='2: eth0    inet6 fd5d:d50a:8c17:2040::250/64 scope global tentative \       valid_lft forever preferred_lft forever
 >  stderr: 2: eth0    inet6 fd5d:d50a:8c17:2040::251/64 scope global \       valid_lft forever preferred_lft forever'
 >  stderr: + 19:48:57: run_send_ua:856: case "$ipstatus" in
 >  stderr: + 19:48:57: run_send_ua:866: '[' 1 -eq 10 ']'
 >  stderr: + 19:48:57: run_send_ua:874: sleep 1
 >  stderr: + 19:48:58: run_send_ua:854: for i in '$(seq 1 10)'
 >  stderr: ++ 19:48:58: run_send_ua:855: ip -o -f inet6 addr show dev eth0 to fd5d:d50a:8c17:2040:0000:0000:0000:0250/64
 >  stderr: + 19:48:58: run_send_ua:855: ipstatus='2: eth0    inet6 fd5d:d50a:8c17:2040::250/64 scope global tentative dadfailed \       valid_lft forever preferred_lft forever
[...]

PR #1396, opened 2019-09-12
Low: IPaddr2: fix to work properly with unsanitized IPv6 addresses

kskmori:
Fix for #1109 and #1357.

I have tested it locally with the OCFT test cases below:
* https://gist.github.com/kskmori/65a715627be207c3353cd31bad485a8c

### Limitations:
It can not be sanitized if the address is unreachable and on the recent distributions
such as RHEL8. (probably depending on the iproute package version)
* CentOS 7: can be sanitized 
```
[root@centos73-1 ~]# /sbin/ip route get 2001:db8:201::0001
unreachable 2001:db8:201::1 dev lo  table unspec  proto kernel  src 2001:db8:101:0:XXXX:XXXX:XXXX:XXXX  metric 429496
```
* RHEL 8: can **NOT** be sanitized
```
[root@rhel80-1 ~]# /sbin/ip route get 200:db8:201::0001
RTNETLINK answers: Network is unreachable
```

I believe that it's not the case of the typical network configurations and only affects to very limited use cases.
Even in such use case, it eventually works successfully (test case 8 of the OCFT scenario above), although a bit tricky thing is happening inside - the first start action is processed with an unsanitized address, and once it is assigned, the succeeding monitor/stop actions are processed with the sanitized address.
Finally if still there's a problem, there is a workaround - ask users to specify the address exactly same as how `ip` shows in such particular use cases.

Please review it if it's acceptable, or if there is a better idea, feedbacks are welcome.
Thanks,

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -477,6 +477,12 @@ ip_init() {
 		fi
 	else
 		FAMILY=inet6
+		# address sanitization defined in RFC5952
+		SANITIZED_IP=$($IP2UTIL route get $OCF_RESKEY_ip | awk '$1~/:/ {print $1}  $2~/:/ {print $2}')
+                if [ -n "$SANITIZED_IP" ]; then
+                    OCF_RESKEY_ip="$SANITIZED_IP"
+                fi
+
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

PR #1398, opened 2019-09-12
nfsserver: performance improvements for systemd enabled systems

jnpkrn:
> I found two critical actions in the script:
> - systemctl status nfs-server  (which also calls journalctl)
> - systemctl list-unit-files

source:
https://lists.clusterlabs.org/pipermail/developers/2019-September/002214.html

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -233,24 +233,25 @@ set_exec_mode()
 		return 0
 	fi
 
-	##
-	# Attempt systemd (with nfs-lock.service).
-	##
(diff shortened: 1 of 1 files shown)


oalbrigt:
Nice improvements. LGTM.

I agree regarding adding systemd-functions to avoid these performance culprits in the future, when this patch has seenn some testinng.

oalbrigt:
Seems like that parameter was introduced in 2011 (https://github.com/systemd/systemd/commit/ebed32bfe832071cde95f6adc01b1097ff079c90), so we should be able to use that without any issues.

PR #1407, opened 2019-09-26
LVM-activate: move pvscan --cache to validate

oalbrigt:
It needs to be called before validate attempts to look at the VG.


oalbrigt:
Patch from David Teigland to avoid issues w/System ID / lvmetad in older versions.

PR #1408, opened 2019-09-27
Adapt azure-lb Resource Agent to support socat usage too

varkoly:
Netcat doesn't handle multiple connections, breaking connectivity to the application in pacemaker cluster. Netcat doesn't allow to restrict which clients can initiate connections.

### DIFF
--- a/heartbeat/azure-lb
+++ b/heartbeat/azure-lb
@@ -17,8 +17,8 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
 # Defaults
-OCF_RESKEY_nc_default="/usr/bin/nc"
 OCF_RESKEY_port_default="61000"
+OCF_RESKEY_nc_default="/usr/bin/nc"
 
 : ${OCF_RESKEY_nc=${OCF_RESKEY_nc_default}}
 : ${OCF_RESKEY_port=${OCF_RESKEY_port_default}}
@@ -53,9 +53,11 @@ Resource agent to answer Azure Load Balancer health probe requests
 
 <parameter name="nc">
 <longdesc lang="en">
-The full name of the nc binary.
+The full path of the used binary. This can be nc or socat path.
+The default is /usr/bin/nc.
+If you need /usr/bin/socat this parameter should be set.
 </longdesc>
-<shortdesc lang="en">Full path name of the nc binary</shortdesc>
+<shortdesc lang="en">Full path of the used binary (nc or socat are allowed)</shortdesc>
 <content type="string" default="${OCF_RESKEY_nc_default}"/>
 </parameter>
 
@@ -100,6 +102,10 @@ lb_monitor() {
 
 lb_start() {
 	cmd="$OCF_RESKEY_nc -l -k $OCF_RESKEY_port"
+	if [ $( basename $OCF_RESKEY_nc ) = 'socat' ]; then
+		#socat has different parameters
+		cmd="$OCF_RESKEY_nc -U TCP-LISTEN:$OCF_RESKEY_port,backlog=10,fork,reuseaddr /dev/null"
+	fi
 	if ! lb_monitor; then
 		ocf_log debug "Starting $process: $cmd"
 		# Execute the command as created above


heartbeat/azure-lb
```diff
@@ -15,11 +15,16 @@
...
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
-
+. ${OCF_FUNCTIONS_DIR}/ocf-distro
 # Defaults
 OCF_RESKEY_nc_default="/usr/bin/nc"
 OCF_RESKEY_port_default="61000"
 
+# On SLES we use socat
+if [ $( get_release_id ) = "SUSE" ]; then
```
diegoakechi: @varkoly Some comments here. The usage of `socat` is not SUSE specific, it can be used in any distro, so this check does not make sense here.

heartbeat/azure-lb
```diff
@@ -15,11 +15,17 @@
...
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
-
+. ${OCF_FUNCTIONS_DIR}/ocf-distro
 # Defaults
-OCF_RESKEY_nc_default="/usr/bin/nc"
 OCF_RESKEY_port_default="61000"
 
+# On SLES we use socat
+if [ -e /usr/bin/socat ]; then
```
diegoakechi: @varkoly as usually both are installed by default, it will always use `socat`, right? Also, there is a OCF function that already tests the binary existence. There is no sense to duplicate that. https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/azure-lb#L164

More than that, if the binary is installed in different location, the RA will not work.

I would remove that check, and if the user does not inform socat, nc will be used by default.

heartbeat/azure-lb
```diff
@@ -53,9 +59,11 @@ Resource agent to answer Azure Load Balancer health probe requests
 
 <parameter name="nc">
 <longdesc lang="en">
-The full name of the nc binary.
+The full name of the used binary. This can be path to nc or socat.
```
diegoakechi:
```suggestion
The full path of the used binary. This can be nc or socat path.
```

heartbeat/azure-lb
```diff
@@ -53,9 +59,11 @@ Resource agent to answer Azure Load Balancer health probe requests
...
+The default is the installed tool. 
+If both are installed nc is the default.
 </longdesc>
-<shortdesc lang="en">Full path name of the nc binary</shortdesc>
+<shortdesc lang="en">Full path name of the used binary</shortdesc>
```
diegoakechi:
```suggestion
<shortdesc lang="en">Full path of the used binary (nc or socat are allowed)</shortdesc>
```

heartbeat/azure-lb
```diff
@@ -15,11 +15,17 @@
 # Initialization:
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
-
+. ${OCF_FUNCTIONS_DIR}/ocf-distro
```
oalbrigt: You can remove this, and the comment about SLES below.

heartbeat/azure-lb
oalbrigt: +1 regarding letting it use nc by default.

heartbeat/azure-lb
```diff
@@ -15,10 +15,9 @@
 # Initialization:
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
-
```
oalbrigt: Can you re-add this empty line for readability, and squash the commits into a single commit?

varkoly:
close for squash the commits

PR #1411, opened 2019-09-30
Adapt azure-lb Resource Agent to support socat usage too

varkoly:


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks. (for future ref: https://github.com/ClusterLabs/resource-agents/pull/1408)

PR #1414, opened 2019-10-07
apache: testing to-execute-file for mere presence is not enough

jnpkrn:
Permission to execute that file must be granted so as not to preclude
the happy case (also, hopefully, the respective "test" program
implementation is not naive about "uid=0 equals superprivileges" for
when the script is not owned by such user while there can be further
security domain constraints, such as a lack of "CAP_DAC_OVERRIDE"
on Linux [imposed, e.g., with SELinux]).


oalbrigt:
LGTM. Thanks.

PR #1418, opened 2019-10-12
Based on the new version of Aliyun CLI 

bd233:


bd233:
> Can one of the admins verify this patch?

yes, they can.

oalbrigt:
You'll have to add some logic that detects whether the old or new version is present, and use correct command/parameters based on that.

bd233:
> 您将必须添加一些逻辑以检测是否存在旧版本或新版本，并基于此使用正确的命令/参数。

The older version of the cli tool aliyun is no longer maintained and should be installed manually by the user before use this agent. Therefore, it does not happen at the same time as both the old version and the new version.

We will update its operation instructions on aliyun official website and SUSE official website.

oalbrigt:
RHEL uses the older version, and you cant expect all versions of all distros to be upgraded to the newest version of the CLI tool.

oalbrigt:
I will cherry-pick your commit, and add the logic to detect if the old or new version is available.

Should `region_id` and `endpoint` be available as parameters to set by users (if so we'll need to add it to the agents metadata)?

bd233:
> I will cherry-pick your commit, and add the logic to detect if the old or new version is available.
> 
> Should `region_id` and `endpoint` be available as parameters to set by users (if so we'll need to add it to the agents metadata)?
Yes, I have added the "aliyuncli path" and "endpoint" parameters. I have modified them to support the new version and the old version. I will submit them later.

heartbeat/aliyun-vpc-move-ip
```diff
@@ -37,27 +54,37 @@ USAGE="usage: $0 {start|stop|status|meta-data}";
 #
 ###############################################################################
 
+request_create_route_entry() {
+	cmd="${OCF_RESKEY_aliyuncli} vpc CreateRouteEntry --RouteTableId $OCF_RESKEY_routing_table --DestinationCidrBlock ${OCF_RESKEY_address}/32 --NextHopId $ECS_INSTANCE_ID --NextHopType Instance ${ENDPOINT}"
+	ocf_log debug "executing command: $cmd"
+	$cmd
+}
+
+request_delete_rote_entry() {
```
oalbrigt: You got a typo here in "route" (and where it's called below)

bd233: you are right，tks

oalbrigt:
ok to test

oalbrigt:
Looks good to me. Thanks.

I'll merge and add some auto-detection so it doesnt fail for users when they upgrade.

oalbrigt:
I've added auto-detection in https://github.com/ClusterLabs/resource-agents/pull/1420.

Feel free to test it, and tell me if there are any issues.

PR #1421, opened 2019-10-23
Fix apache rs on suse

varkoly:
Apache resource agent is not working when statusurl was set.
The reason was bad detecting the OS and bad start code.

### DIFF
--- a/heartbeat/apache
+++ b/heartbeat/apache
@@ -167,13 +167,16 @@ silent_status() {
 
 # May be useful to add other distros in future
 validate_default_config() {
-	if [ -e /etc/SuSE-release ]; then
-		validate_default_suse_config
-	elif [ -e /etc/debian_version ]; then
-		validate_default_debian_config
-	else
-		return 0
-	fi
+	case "$( get_release_id )" in
+                *SUSE)
+                        validate_default_suse_config
+                        ;;
+                Debian)
+                        validate_default_debian_config
+                        ;;
+                *)
+                        return 0
+        esac
 }
 
 # When using the default /etc/apache2/httpd.conf on SUSE, the file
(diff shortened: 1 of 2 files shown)


heartbeat/apache
```diff
@@ -242,9 +243,17 @@ apache_start() {
 	validate_default_config || return $OCF_ERR_CONFIGURED
 
 	if [ -z $PIDFILE_DIRECTIVE ]; then
-		ocf_run $HTTPD $HTTPDOPTS $OPTIONS -f $CONFIGFILE
+		if [ -z "$LOAD_STATUS_MODULE" ]; then
```
oalbrigt: This seems to do what `HTTPDOPTS="$HTTPDOPTS -C $LOAD_STATUS_MODULE"` already did, but in 2 places which makes it more prone to future errors.

varkoly: It does not do it. If I do it so you have proposed bash makes a lot of sensless quoting and httpd does not start. Do not forget $LOAD_STATUS_MODULE already contains quotes. I've already spent a lot of time to find the right solution but this seemed to me to be the most lear and stable.

oalbrigt: Have you tried changing it to `HTTPDOPTS="$HTTPDOPTS -C \"$LOAD_STATUS_MODULE\""`?

varkoly: Yes I've. Result is:
... -C '"LoadModule' status_module '/usr/lib64/apache2-prefork/mod_status.so"'  ...
instead of:
... -C 'LoadModule status_module /usr/lib64/apache2-prefork/mod_status.so"  ...
And apache2 can not start. I've tested it with the combination of single quote too. Noting hase helped.

PR #1428, opened 2019-10-30
Low: MailTo: fix variable expansion

vvidic:
Wrong kind of braces used (expand to 0 or fail
to expand with dash as shell).


oalbrigt:
Nice catch. Thanks.

PR #1431, opened 2019-11-13
azure-lb: set socat to default on SUSE distributions.

varkoly:
Socat was tested on SUSE and it works fine. To avoid the need to change the configuration we set socat to default on SUSE distributions.

### DIFF
--- a/heartbeat/azure-lb
+++ b/heartbeat/azure-lb
@@ -18,6 +18,11 @@
 
 # Defaults
 OCF_RESKEY_nc_default="/usr/bin/nc"
+case "$( get_release_id )" in
+	*SUSE)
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
Can you update the commit message with prefix "azure-lb:"? So it's clear from the commit log which agent it's updating.

varkoly:
> Can you update the commit message with prefix "azure-lb:"? So it's clear from the commit log which agent it's updating.
Changed.

diegoakechi:
@varkoly please, rebase your branch and squash this change into a single commit with the right title.

varkoly:
will create new PR

PR #1432, opened 2019-11-14
azure-lb Set socat to default on SUSE distributions.

varkoly:
Socat was tested on SUSE and it works fine. To avoid the need to change the configuration we set socat to default on SUSE distributions.


oalbrigt:
ok to test

PR #1433, opened 2019-11-14
Filesystem: refresh UUID in the start phase

zzhou1:
In the case a fresh filesystem is just created from another node on the
shared storage, is not visible yet. Then try partprobe to refresh
/dev/disk/by-uuid/* up to date.

Signed-off-by: Roger Zhou <zzhou@suse.com>


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1435, opened 2019-11-19
iSCSI logical unit fix

ezaton:
When using XenServer with multipath as iSCSI client for PCS, the lio-t module calculates the scsi_sn automatically, if not specified explicitly. The scsi_sn (saved in /sys/kernel/config/target/core/<backend>/<lun name>/wwn/vpd_unit_serial when targetcli defines the backend) is not maintained during failovers. Clients 'multipath' daemon might fail to re-identify the disks. XenServer is extremely vulnerable to this, because the wwn is kept in a file and explicitly used. Failover of a cluster iSCSIlogicalUnit to another node, and then reboot to XenServer host might lead to storage inaccessible. This was proven true, especially when multiple backend devices (and LUNs) are present in the same resource group.
The change added some note about it for the meta-data section of scsi_sn and added a condition in which, during the backend creation in lio-t (targetcli), the scsi_sn would be set, if specified.

### DIFF
--- a/heartbeat/iSCSILogicalUnit.in
+++ b/heartbeat/iSCSILogicalUnit.in
@@ -150,7 +150,11 @@ is the resource name, truncated to 24 bytes.
 <parameter name="scsi_sn" required="0" unique="1">
 <longdesc lang="en">
 The SCSI serial number to be configured for this Logical Unit.
-The default is a hash of the resource name, truncated to 8 bytes.
+The default is a hash of the resource name, truncated to 8 bytes,
+meaning 26 hex characters.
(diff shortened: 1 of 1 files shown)


oalbrigt:
It seems like you need to rebase your local git repo.

Just do a `cp heartbeat/iSCSILogicalUnit.in ..` to avoid breaking any of your changes (so you can copy the file back if you do any mistakes.

You can rebase by running:
```
git fetch --all
git rebase upstream/master
```

If you cant `git push` after rebasing just add `-f` to force-push it (e.g. if a commit has changed it's commit hash, message or anything).

oalbrigt:
Thanks.

PR #1439, opened 2019-12-04
IPaddr2: ipt_CLUSTERIP "iptables" extension not "nft" backend compatible

jnpkrn:
Reference:
https://lists.clusterlabs.org/pipermail/users/2019-December/026674.html
(thread also sketches a future ambition for a [presumably, to revert
the habit of a functional overloading] separate agent to use
"xt_cluster" extension/cluster match).

Signed-off-by: Jan Pokorný <jpokorny@redhat.com>

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -123,6 +123,8 @@ VLDIR=$HA_RSCTMP
 SENDARPPIDDIR=$HA_RSCTMP
 CIP_lockfile=$HA_RSCTMP/IPaddr2-CIP-${OCF_RESKEY_ip}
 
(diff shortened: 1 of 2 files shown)


oalbrigt:
LGTM.

oalbrigt:
Btw. I think we should also add a note to metadata that support is deprecated in newer kernels, but that it still might work if your distro has iptables-legacy.

jnpkrn:
Good point, thanks.

jnpkrn:
Came to conclusion said gory details cannot be grasped without a bit
more context introduced beforehand :-)

oalbrigt:
Thanks.

PR #1442, opened 2019-12-23
Filesystem: respect udevd need time to create symlinks and also accomodate -U -L device

zzhou1:
There is a race condition that partprobe might return before the UUID
symlink get created. Particularly, when the system has many devices, the
udev daemon could need visible time to process the udev event queue.
Hence, wait udev for a moment.

Filesystem RA's DEVICE parameter accepts "-U <uuid>" and "-L <label>" in
addition to "/dev/xxx". Let's add a new function
trigger_udev_rules_if_need() to accommodate them all.

Merry X'mas and Happy New Year!

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -431,6 +431,47 @@ fstype_supported()
 
 
 #
+# In the case a fresh filesystem is just created from another
+# node on the shared storage, and is not visible yet. Then try
+# partprobe to refresh /dev/disk/by-{label,uuid}/* up to date.
+#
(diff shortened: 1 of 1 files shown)


oalbrigt:
Merry xmas and a happy new year.

Can you rename the function to `trigger_udev_rules_if_needed()`?

zzhou1:
> Can you rename the function to `trigger_udev_rules_if_needed()`?

errr, pool my English ;)

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1443, opened 2020-01-07
podman: simplify the code for checking if an image exists

mbaldessari:
Let's switch to the existing 'podman image exists' to determine if
an image is local to the machine or if it needs pulling.

Let's change this for a number of reasons:
- It simplifies the code greatly
- It makes everything work even when the image is prefixed with 'localhost/'
  names which are not detected as local images
- It makes the code faster as we do not need to list all the images as
  that is more expensive

Tested by running a couple of full containerized OpenStack deployments,
one with images tagged as 'localhost/foo/mysql:pcmklatest' and one
with a normal registry host prefix.

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -441,31 +441,7 @@ podman_stop()
 
 image_exists()
 {
-	# if no tag was specified, use default "latest"
-	local COLON_FOUND=0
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

PR #1444, opened 2020-01-09
LVM-activate: verify vg or lv validity

zhaohem:
When LVM underlying devs disappear, eg iSCSI SAN disks suffer network
disconnect, the pvs/vgs/lvs won't show related PVs/VGs/LVs, but dm
devs won't disappear. This makes the lvm_status() doesn't work as
expect.

This patch uses simple & stupid method "read 1 byte from underlaying
dev" to detect abnormal underlying devices.


oalbrigt:
ok to test

zhaohem:
Sorry, I forgot to signed-off, force-pushed.

zhaohem:
my solution didn't think about iscsi delay scenario. when iscsi connection timeout, read lun will return EIO.

oalbrigt:
You should move the information down to the bottom of the notes section, as the rest of the description is more general.

oalbrigt:
LGTM. Thanks.

PR #1448, opened 2020-01-16
rabbitmq-cluster: ensure we delete nodename if stop action fails

jeckersb:
If the stop action fails, we want to remove the nodename from the crm
attribute.  Currently it is possible for the stop action to fail but
the rabbitmq server does actually stop.  This leaves the attribute
still present.  This means if the entire rabbitmq cluster is stopped,
it is not possible to start the cluster again because the first node
to start will think there is at least one other node running.  Then
the node tries to join an existing cluster instead of rebootstrapping
the cluster from a single node.


dciabrin:
LGTM thanks eck!

oalbrigt:
LGTM. Thanks.

PR #1451, opened 2020-01-24
New tickle_tcp implementation

raltnoeder:
I originally stumbled upon a buffer handling problem when I attempted to use the previous tickle_tcp utility manually and began to investigate in order to fix that problem, but found more problems on the way and then decided to start from scratch instead.


oalbrigt:
We do not want to rewrite this from scratch, as it's likely to introduce new issues to users where it worked fine before.

raltnoeder:
I don't think the risk of introducing new issues to users where it worked even despite the problems in the previous implementation is anywhere near the risk that the previous implementation might break in the future, or for any new users or new use cases, and is also generally very low.

The only current user of the utility is the portblock RA, the inputs accepted by the utility are backwards compatible, and any changed error messages to stderr or changed exit codes are never checked or otherwise used by anything.
For any potential users who integrated the utility in their own custom software (unlikely in the first place), a simple changelog entry should suffice.

The reason why I started from scratch was that I was convinced that fixing the bugs in the previous implementation has a higher chance of introducing bugs and/or would be harder to develop and verify than a clean new implementation.

If really necessary, I could also add compatibility tests, to enable testing of the input accepted and output and exit codes produced by both utilities under the same conditions.

PR #1273, opened 2018-12-04
Add support for multiple VPC routing tables in routing_tables parameter

gguifelixamz:
Hello,

This pull request is to add support for multiple AWS VPC routing tables within a single aws-vpc-move-ip cluster resource as requested by some customers.

Design options were discussed with different people and the options considered were:

a) A single aws-vpc-move-ip resource with a multiple routing tables in `routing_tables` parameter
b) Multiple aws-vpc-move-ip resources but including a "dummy" flag to the ones that were only supposed to manipulate routing tables and not local IPs - basically an aws-vpc-move-ip resource for managing routing tables only

We decided to go with option "A", as it keeps the cluster configuration and topology the way it is today and only modifies a single configuration parameter. For example:

```
primitive res_AWS_IP ocf:suse:aws-vpc-move-ip \
        params address=192.168.1.9 routing_table="rtb-XXXXXXXXXXXXXXXXXX,rtb-YYYYYYYYYYYYYYY" interface=eth0 profile=AWS_CLI_profile \
        op start interval=0 timeout=180 \
        op stop interval=0 timeout=180 \
        op monitor interval=60 timeout=60 \
        meta target-role=Started
```

Lines 236-238 from this request were changed to match coding standards already in place for ec2ip_monitor() and ec2ip_drop()

Example testing from command line is shown below.

To start:

```
OCF_RESKEY_address=192.168.1.9 OCF_RESKEY_routing_table=rtb-XXXXXXXXXXXXXXXXXX,rtb-YYYYYYYYYYYYYYY OCF_RESKEY_profile=AWS_CLI_profile OCF_RESKEY_interface=eth0 OCF_RESKEY_monapi=true OCF_ROOT=/usr/lib/ocf /usr/lib/ocf/resource.d/suse/aws-vpc-move-ip start
```

To monitor:

```
OCF_RESKEY_address=192.168.1.9 OCF_RESKEY_routing_table=rtb-XXXXXXXXXXXXXXXXXX,rtb-YYYYYYYYYYYYYYYOCF_RESKEY_profile=AWS_CLI_profile OCF_RESKEY_interface=eth0 OCF_RESKEY_monapi=true OCF_ROOT=/usr/lib/ocf /usr/lib/ocf/resource.d/suse/aws-vpc-move-ip monitor
```

To stop:

```
[...]

PR #1454, opened 2020-02-10
Allow user to specify role with which to query/update route table

jason-mccloskey:
When leveraged in Organizations, often times the profile executing commands will have to assume a role in order to access the route table (this is especially the case in a shared VPC model). Being able to optionally assume a role for route table api calls is advantageous to large organizations.

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -145,6 +145,24 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 END
 }
 
+
+execute_cmd_as_role(){
+	cmd=$1
+	role=$2
+	output="$(aws sts assume-role --role-arn $role --role-session-name AWSCLI-RouteTableUpdate --profile $OCF_RESKEY_profile --output=text)"
+	export AWS_ACCESS_KEY_ID="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $5}')"
+	export AWS_SECRET_ACCESS_KEY="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $7}')"
+	export AWS_SESSION_TOKEN="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $8}')"
+
+	#Execute command
+	ocf_log debug "Assumed Role ${role}"
+	ocf_log debug "$(aws sts get-caller-identity)"
+	ocf_log debug "executing command: $cmd"
+	response="$($cmd)"
+	unset output AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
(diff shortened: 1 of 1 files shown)


oalbrigt:
Wouldnt it make more sense to only set it once? With this code it'll be set and unset twice for the start-action due to `monitor` followed by `...get_and_configure`.

oalbrigt:
The indentation needs to be updated to tabs (to follow the rest of the agent).

gguifelixamz:
Helo @jason-mccloskey 

Thank you for your submission.

Please, review the comments made by @oalbrigt, and I also would like to see more comments on the code, and a quick explanation of how to use this feature so our partners can also incorporate this feature into their documentation - something similar as previously done in:
- https://github.com/ClusterLabs/resource-agents/pull/1273
- https://github.com/ClusterLabs/resource-agents/pull/1409

Feel free to reach out directly to me in case you have questions.

jason-mccloskey:
Hi @gguifelixamz 

The spaces have all been updated to tabs.

@oalbrigt if you would prefer, I'd be happy to refactor the code to only assume role once. I didn't want to cause any potential future issues by having left the default credentials updated between function calls and wanted to make sure each function call worked independently of the other.  

What does this feature do?
In a multi-account architecture, the profile being used to update an IP address will not always have permissions to update a route table. This is expected to be the case in a shared VPC architecture, where different accounts own the VPC and EC2 Instances. By setting `OCF_RESKEY_routing_table_role="arn:aws:iam::<account-id>:role/<role-name>"`, the specified role will be assumed when querying and updating the route table. Please note that this requires the profile you are using to have permissions (sts:AssumeRole) to assume the specified role. If `OCF_RESKEY_routing_table_role` is not set, then the profile will be used to query and update the route table as before.

oalbrigt:
Usually you would use the same role for all operations (from my experience), so I think it makes sense to only set it once, unless you know of scenarioes where this wouldnt be the case.

gguifelixamz:
@oalbrigt I had a chat with @jason-mccloskey and I'm fine with these changes. Is it OK for you? If yes, feel free to merge this PR.

@diegoakechi FYI only

oalbrigt:
Sure. We can improve it later if we find that it makes sense to use same role for all of it.

oalbrigt:
Btw. We also need to add it to the metadata, so pcs knows how to handle the parameter, and users can find info about it without reading the code.

oalbrigt:
Patch to add parameter to metadata: https://github.com/ClusterLabs/resource-agents/pull/1462

gguifelixamz:
@oalbrigt is it possible to rollback the changes introduced by this PR and by #1462 ?

oalbrigt:
It's better to fix any issues in case any distros are using it already.

PR #1461, opened 2020-02-29
iSCSILogicalUnit: fix default value for OCF_RESKEY_liot_bstype

vvidic:
Failed with invalid if command because value was not set to default.
Rest of the changes are space and syntax cleanups.


oalbrigt:
LGTM. Thanks.

Issue #1049, opened 2017-11-06
Tempfiles should always be created using mktemp, not using PID or hard-coded filenames


krig:
This was flagged by the SUSE security team:

There are various problems with tmp-races. Some of them may be used
for LPE:

sapdb-nosha.sh:

```
143 oracle_stop() {
144 echo '#!/bin/sh
145 LOG=$HOME/stopdb.log
146 date > $LOG
147 
148 if [ -x "${ORACLE_HOME}/bin/sqlplus" ]
149 then
150     SRVMGRDBA_EXE="${ORACLE_HOME}/bin/sqlplus"
151 else
152    echo "Can not find executable sqlplus" >> $LOG
153    exit 1
154 fi
155 
156 $SRVMGRDBA_EXE /NOLOG >> $LOG << !
157 connect / as sysdba
158 shutdown immediate
159 exit
160 !
161 rc=$?
162 cat $LOG
163 exit $rc' > $TEMPFILE                                                                                                                                   
164 
165 chmod 700 $TEMPFILE
166 chown $sidadm $TEMPFILE
167 
168 su - $sidadm -c $TEMPFILE
...
743 TEMPFILE="/tmp/SAPDatabase.$$.tmp"
```

and more occurrences of this.

mysql:

```
 987 DEBUG_LOG="/tmp/mysql.ocf.ra.debug/log"
 988 if [ "${DEBUG_LOG}" -a -w "${DEBUG_LOG}" -a ! -L "${DEBUG_LOG}" ]; then
 989     DEBUG_LOG_DIR="${DEBUG_LOG%/*}"
 990     if [ -d "${DEBUG_LOG_DIR}" ]; then                                                                                                                 
 991         exec 9>>"$DEBUG_LOG"
 992         exec 2>&9
 993         date >&9
 994         echo "$*" >&9
 995         env | grep OCF_ | sort >&9
 996         set -x
 997     else
 998         exec 9>/dev/null
 999     fi
1000 fi
```

ClusterMon:

```
235 : ${OCF_RESKEY_update:="15000"}
236 : ${OCF_RESKEY_pidfile:="/tmp/ClusterMon_${OCF_RESOURCE_INSTANCE}.pid"}
237 : ${OCF_RESKEY_htmlfile:="/tmp/ClusterMon_${OCF_RESOURCE_INSTANCE}.html"}
```

All these should better use mktemp.

dmuhamedagic:
[...]

PR #1467, opened 2020-03-25
Make secure tmp files

varkoly:
There are various problems with tmp-races. Some of them may be used for LPE.
This changes makes resource-agents safer.

### DIFF
--- a/heartbeat/ClusterMon
+++ b/heartbeat/ClusterMon
@@ -45,8 +45,8 @@
 OCF_RESKEY_user_default="root"
 OCF_RESKEY_update_default="15000"
 OCF_RESKEY_extra_options_default=""
-OCF_RESKEY_pidfile_default="/tmp/ClusterMon_${OCF_RESOURCE_INSTANCE}.pid"
-OCF_RESKEY_htmlfile_default="/tmp/ClusterMon_${OCF_RESOURCE_INSTANCE}.html"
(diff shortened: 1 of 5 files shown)


varkoly:
Fixes the issue: https://github.com/ClusterLabs/resource-agents/issues/1049

oalbrigt:
ok to test

oalbrigt:
LGTM. @fdanapfel can you see if you're fine with these changes for the SAP agents?

diegoakechi:
@fmherschel You might want to take a look here too...

fmherschel:
> @fmherschel You might want to take a look here too...

The only thing I am wondering about is why parts of the patch are moving the temporary files from /tmp to (literal) "/var/run" while others are using the OCF reference "${HA_RSCTMP}".
Is that difference implemented by intention?

fdanapfel:
@oalbrigt The changes for the SAP agent look OK to me.

@fmherschel: looks to me like the second commit fixes this so that "${HA_RSCTMP}" is used instead of /var/un in all places

oalbrigt:
> > @fmherschel You might want to take a look here too...
> 
> The only thing I am wondering about is why parts of the patch are moving the temporary files from /tmp to (literal) "/var/run" while others are using the OCF reference "${HA_RSCTMP}".
> Is that difference implemented by intention?

That seems to be a mistake in the first commit, that he fixes in the 2nd commit. If you check Files Changed in the PR you'll see no reference to "/var/...".

PR #1468, opened 2020-03-26
Fix audit issue with secure tmp file

nick-wang:
* Change the drbd.linbit tmp file location to a more secure place. (https://github.com/ClusterLabs/resource-agents/issues/1049)
* Correct the stdout parameter from start to stop.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

nick-wang:
Thanks for the review.

Issue #1030, opened 2017-08-28
oracle: Insecure design for the monitoring user/password?


krig:
The oracle agent has the `monuser` and `monpassword` parameters which default to `OCFMON` / `OCFMON` respectively. If the monuser doesn't already exist, the agent will automatically create it and set the password in the start action.

So someone unaware of this who doesn't configure or precreate the `monuser` will get a OCFMON user with password "OCFMON" created in their database, it seems.

That seems horribly insecure!

The question is what the right fix is. Should there just not be any default values for monuser and monpassword? Then if a monitor action is configured, the validate action would check that they have been configured.

oalbrigt:
I think the best way would be to leave the defaults as is (to avoid issues for users upgrading resource-agents) and replace the automatic creation with error if the user/pass is incorrect/doesnt exist.

krig:
Yeah, that sounds better.

dmuhamedagic:
On Mon, Aug 28, 2017 at 08:02:30AM +0000, Kristoffer Grönlund wrote:
> The oracle agent has the `monuser` and `monpassword` parameters which default to `OCFMON` / `OCFMON` respectively. If the monuser doesn't already exist, the agent will automatically create it and set the password in the start action.
> 
> So someone unaware of this who doesn't configure or precreate the `monuser` will get a OCFMON user with password "OCFMON" created in their database, it seems.
> 
> That seems horribly insecure!
> 
> The question is what the right fix is. Should there just not be any default values for monuser and monpassword? Then if a monitor action is configured, the validate action would check that they have been configured.

A bit of history: it was initially the sqldba user which is
passwordless if run as root. However, every connection would be
logged which resulted in rather big logs, hence monuser got
introduced. It was necessary to create it automatically as
otherwise, on upgrade, the monitor op would fail. This user is
granted select only on one system table, so it shouldn't pose a
security risk.

krig:
Aha, I see.

Hmm. I guess at this point, making it an error instead of automatic should be OK then? Since anyone upgrading would already have the user, and anyone installing fresh would then have to create it.

On the other hand I can see some potential for creating security holes when introducing a human into the mix, since the admin would need to create the user with the right restrictions and so on. Not sure what the best thing is, here, actually.

dmuhamedagic:
On Wed, Aug 30, 2017 at 08:36:17PM +0000, Kristoffer Grönlund wrote:
[...]

PR #1472, opened 2020-04-02
Increase the security of monitor user in oracle

nick-wang:
With static default credentials for a user is not safe
even for limited privilege. A local user may login to
the DB to do some attack by bugs or leaks. (https://github.com/ClusterLabs/resource-agents/issues/1049)

It is better to replace the automatic creation with
new user and leave the default password for legacy update.

Also an open issue discussion (#1030) for this.

### DIFF
--- a/heartbeat/oracle
+++ b/heartbeat/oracle
@@ -132,6 +132,8 @@ that the password for this user does not expire.
 <longdesc lang="en">
 Password for the monitoring user. Make sure
 that the password for this user does not expire.
+Need to explicitly set a password to a new monitor
+user for the security reason.
 </longdesc>
 <shortdesc lang="en">monpassword</shortdesc>
 <content type="string" default="$OCF_RESKEY_monpassword_default" />
@@ -440,6 +442,12 @@ check_mon_user() {
 			return 1
 		fi
 	fi
+
+	if [ -z "$OCF_RESKEY_monpassword" ]; then
+		ocf_exit_reason "Please explicitly set a password for $MONUSR oracle user"
+		exit $OCF_ERR_CONFIGURED
+	fi
+
 	output=`dbasql mk_mon_user show_mon_user`
 	if echo "$output" | grep -iw "^$MONUSR" >/dev/null; then
 		return 0


oalbrigt:
MONPWD is automatically set to the _default value if you provide an empty password: https://github.com/ClusterLabs/resource-agents/blob/f68497ff92b42e986ddd56e8f0190f9ce7633a06/heartbeat/oracle#L779

And you also need to set it to empty to use default value, so this change would break already working configurations.

nick-wang:
@oalbrigt Thanks for the quick review. I don't get the idea so clear, sorry.

> MONPWD is automatically set to the _default value if you provide an empty password:
> 
> https://github.com/ClusterLabs/resource-agents/blob/f68497ff92b42e986ddd56e8f0190f9ce7633a06/heartbeat/oracle#L779
>
I checked the `${OCF_RESKEY_monpassword}` instead of `${MONPWD}`. 
I think it is still empty or NULL if not set explicitly.
 
> And you also need to set it to empty to use default value, so this change would break already working configurations.
For a working configurations which already have `MONUSR` created in the database, it will continue to use the default `${OCF_RESKEY_monpassword_default}` if not set.

heartbeat/oracle
```diff
@@ -439,7 +441,13 @@ check_mon_user() {
 			ocf_log err "sqlplus output: $output( $output2 )"
 			return 1
 		fi
+	else
```
gao-yan, line 444: IIUC, an existing setup with existing user profile would have already returned from above and continued in oracle_start() no matter what.

When it's coming to here, it means the user profile doesn't exist and it's on its way to create the user profile in below. So it seems like a right place to check what `OCF_RESKEY_monpassword` is like.

Not necessary to change. It'd be the same without `else`, but it'd probably be clearer ...

nick-wang:
@oalbrigt Could you please help to review again?
I move the exit conditions of creating new user out of the check block for better readability.

For a exist user with legacy update,before reaching the `if` it should have already exit the function.

Thanks in advanced.

heartbeat/oracle
oalbrigt: It's set to `OCFMON` from $OCF_RESKEY_monpassword_default on the following line when monpassword isnt specified by the user:
https://github.com/ClusterLabs/resource-agents/blob/f68497ff92b42e986ddd56e8f0190f9ce7633a06/heartbeat/oracle#L779

nick-wang: Thanks for the review!

I moved the check out of the whole block for a better readability.

oalbrigt:
The correct check would be to do `if [ -z "$MONPWD" ]; then`, which is the variable that's used in the rest of the agent, and it's set to the _default which is `OCFMON` if the user hasnt specified a password to use.

nick-wang:
> The correct check would be to do `if [ -z "$MONPWD" ]; then`, which is the variable that's used in the rest of the agent, and it's set to the _default which is `OCFMON` if the user hasnt specified a password to use.

Yes, it will use $MONPWD. But it is already set there by the code
```
MONPWD=${OCF_RESKEY_monpassword:-$OCF_RESKEY_monpassword_default}
```
so it is impossible empty. And can use for legacy database if the `MONUSR` exist.

While `${OCF_RESKEY_monpassword}` suppose to be empty if admin not explicitly set in resource agent for pacemaker. Only in this case, can know whether admin set the `monpassword` or not, then force quit to ask admin to reconfigure for a password in case not set.

gao-yan:
> The correct check would be to do `if [ -z "$MONPWD" ]; then`, which is the variable that's used in the rest of the agent, and it's set to the _default which is `OCFMON` if the user hasnt specified a password to use.

But it's `OCF_RESKEY_monpassword` rather than `MONPWD`  that's interesting for this specific check. The check is not about what will be used for the rest of the agent. It's about whether a "new" setup without user profile yet is allowed to proceed without `OCF_RESKEY_monpassword` being explicitly set.

oalbrigt:
If you do that you'll break already working setups for users who upgrade and has it setup to use default.

gao-yan:
> If you do that you'll break already working setups for users who upgrade and has it setup to use default.

Why? Existing setups will return from the above block `if echo "$output" | grep -iw "^$MONUSR" >/dev/null; then` without reaching here, right?

oalbrigt:
Good point. I get it now :)

oalbrigt:
ok to test

heartbeat/oracle
```diff
@@ -440,6 +442,12 @@ check_mon_user() {
 			return 1
 		fi
 	fi
+
+	if [ -z "$OCF_RESKEY_monpassword" ]; then
+		ocf_exit_reason "Please explicitly set a password for $MONUSR oracle user"
+		return $OCF_ERR_CONFIGURED
```
gao-yan, line 448: Only thing is, this return code won't be kept/interpreted when check_mon_user() returns.

oalbrigt: Nice catch. Seems like using exit instead of return would be the easiest way to solve that.

nick-wang:
Oyvind, Yan

Thanks for the review.
Rebased with the change from `return $OCF_ERR_CONFIGURED` to `exit $OCF_ERR_CONFIGURED`

oalbrigt:
LGTM. Thanks.

PR #1473, opened 2020-04-02
galera: Fix value used for connecting with the empty password.

jsuchome:
When check_passwd_use_empty option was introduced, it added a possibility
to connect to mysql explicitly using empty password, thus bypassing defaults
set in ~/.my.cnf.
However the the empty password set for MYSQL_OPTIONS_CHECK variable was wrong,
adding too many levels of escaping and ultimately failing when passed to mysql client.

This commit changes the password value in MYSQL_OPTIONS_CHECK so that it really
ends in passing the empty string.

Original PR here: https://github.com/ClusterLabs/resource-agents/pull/1330

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -989,7 +989,7 @@ fi
 MYSQL_OPTIONS_CHECK="-nNE --user=${OCF_RESKEY_check_user}"
 
 if ocf_is_true "${OCF_RESKEY_check_passwd_use_empty}"; then
-    MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password=''"
+    MYSQL_OPTIONS_CHECK="$MYSQL_OPTIONS_CHECK --password="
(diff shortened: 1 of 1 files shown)


jsuchome:
CC @toabctl

jsuchome:
Further explanation: when I run `crm_resource  -r galera -N d52-54-77-77-01-01 --force-check` with `-VVV` I can see (without my patch of course)

`MYSQL_OPTIONS_CHECK='-nNE --user=monitoring --password='\'''\'''`

being set. This is the quoting I meant and it is not passing empty password to the client.

dciabrin:
LGTM, tested locally on my 3-node cluster, with empty root password.
  . without the patch I get an error that prevents the first promotion to finish succesfully, blocking the bootstrap altogether.
  . with the patch I get the cluster bootstrapped

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1476, opened 2020-04-06
Remove lio_iblock usage in lio-t implementation of iSCSILogicalUnit

### LINKED ISSUE #1256: iSCSILogicalUnit don't set the correct sn
Hello, 
when I've multiple LUNs, the iSCSILogicalUnit-agent set only for the first LUN the correct sn, all others get a different one, I guess something is wrong in Line 419 or so. It looks like that only the first sn will be set correct, I add some echo's output to the iSCSILogicalUnit script:
```
diff /usr/lib/ocf/resource.d/heartbeat/iSCSILogicalUnit iSCSILogicalUnit.org
49,51d48
< echo "" >> /tmp/iscsilun_debug.log
< echo date >> /tmp/iscsilun_debug.log
< 
62d58
< echo ${OCF_RESOURCE_INSTANCE} >> /tmp/iscsilun_debug.log
64,65d59
< echo -n "die sn:  " >> /tmp/iscsilun_debug.log
< echo $sn >> /tmp/iscsilun_debug.log
68,69d61
< echo -n "die ocf_sn:  " >> /tmp/iscsilun_debug.log
< echo $OCF_RESKEY_scsi_sn >> /tmp/iscsilun_debug.log
417,418d408
<               echo -n "vor if reskey_sn:  " >> /tmp/iscsilun_debug.log
<               echo ${OCF_RESKEY_scsi_sn} >> /tmp/iscsilun_debug.log
421,422d410
<                       echo -n "in if schleife:  " >> /tmp/iscsilun_debug.log
<                       cat /sys/kernel/config/target/core/iblock_${OCF_RESKEY_lio_iblock}/${OCF_RESOURCE_INSTANCE}/wwn/vpd_unit_serial  >> /tmp/iscsilun_debug.log
447d434
<               echo "fertig ........................" >> /tmp/iscsilun_debug.log
```
the output is:
```
date
iscsi-lun0
die sn:  d0e5396557d77dc7ba590376f1bf5b0f
die ocf_sn:  d0e53965
vor if reskey_sn:  d0e53965
in if schleife:  T10 VPD Unit Serial Number: d0e53965
fertig ........................

date
iscsi-lun1
die sn:  db01ac2635d5fd02aa628af69b9542fc
die ocf_sn:  db01ac26
vor if reskey_sn:  db01ac26
in if schleife:  fertig ........................

date
iscsi-lun2
die sn:  0acb0780261a5de6f191831f74a13a80
die ocf_sn:  0acb0780
vor if reskey_sn:  0acb0780
in if schleife:  fertig ........................
```
(some other errors here but for quick and dirty its ok) 
so I will look deeper into it

fpiecka:
lio_iblock parameter is used in iSCSILogicalUnit_start(), although it is unsupported in lio-t implementation (see iSCSILogicalUnit_validate()).

This PR removes the use of lio_iblock and instead tries to autodetect the backstore index (for lio-t only).

Indexes are assigned sequentially and their value depends on the order in which the backstores are created by targetcli (i.e. on the order in which resources are started).

In current implementation, only 1 LUN resource based on lio-t works (tested in CentOS 8). The second one fails to start as it has a higher index than 0.

I believe this PR should resolve #1256.

### DIFF
--- a/heartbeat/iSCSILogicalUnit.in
+++ b/heartbeat/iSCSILogicalUnit.in
@@ -273,6 +273,8 @@ Using distinct values here avoids a warning in LIO "LEGACY: SHARED HBA";
 and it is necessary when using multiple LUNs started at the same time
 (eg. on node failover) to prevent a race condition in tcm_core on mkdir()
 in /sys/kernel/config/target/core/.
+
+This parameter is only used by LIO. LIO-T ignores it.
 </longdesc>
 <shortdesc lang="en">LIO iblock device number</shortdesc>
 <content type="integer" default="${OCF_RESKEY_lio_iblock_default}"/>
@@ -426,7 +428,8 @@ iSCSILogicalUnit_start() {
 	lio-t)
 		ocf_take_lock $TARGETLOCKFILE
 		ocf_release_lock_on_exit $TARGETLOCKFILE
-		iblock_attrib_path="/sys/kernel/config/target/core/iblock_${OCF_RESKEY_lio_iblock}/${OCF_RESOURCE_INSTANCE}/attrib"
+		iblock_path="/sys/kernel/config/target/core/*_*/${OCF_RESOURCE_INSTANCE}"
+		iblock_attrib_path="${iblock_path}/attrib"
 		# For lio, we first have to create a target device, then
 		# add it to the Target Portal Group as an LU.
 		# Handle differently 'block' and 'fileio'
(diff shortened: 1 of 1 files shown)


fpiecka:
This PR seems a bit unambitios when compared with  #1072 which would solve the problem too. I'll ask the author whether ho could resolve the conflicts.

PR #1478, opened 2020-04-06
ZFS: Importforce should not be a unique parameter.

guilbaults:
In a case of multiple pools on the same server, the unique argument for
importforce will prevent a 2nd resource from being added with the following
message:
```
Error: Value 'true' of option 'importforce' is not unique across 'ocf:heartbeat:ZFS' resources.
```


oalbrigt:
Thanks.

PR #1479, opened 2020-04-13
gcp-vpc-move: Disable google api cache discovery

arbulu89:
The underlying google api code cache discovery process
uses deprecated code. In order to avoid the usage of that
code and the related warning messages the cache usage is
disabled


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1489, opened 2020-04-30
Eliminated logging.basicConfig()

yandroskaos:
That method installed a root logger to be used with logging methods.
The root logger is a stream logger which sends output to stderr,
so every log data ended up in stderr as well


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1490, opened 2020-04-30
Dummy RA in Python

yandroskaos:
Mimicking the behaviour of the shell Dummy RA, a dummy RA wirtten in Python is provided as example.

### DIFF
--- a/heartbeat/dummypy.in
+++ b/heartbeat/dummypy.in
@@ -42,6 +42,9 @@ sys.path.append(OCF_FUNCTIONS_DIR)
 
 import ocf
 
+OCF_RESKEY_state_default=f'{ocf.get_parameter("HA_RSCTMP")}/dummypy-{ocf.get_parameter("OCF_RESOURCE_INSTANCE")}.state'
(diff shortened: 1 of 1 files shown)


oalbrigt:
Please only change indentation where it's inconsistant, so patches are easy to apply/review for distros instead of being a full replacement of the code.

oalbrigt:
Nevermind. I thought it ws the other PR.

oalbrigt:
ok to test

oalbrigt:
LGTM.

oalbrigt:
Thanks.

oalbrigt:
I've made it Python 2.x compatible, so it passes our CI:
https://github.com/ClusterLabs/resource-agents/pull/1521

Can you check if I've interpreted line 97 correctly?

yandroskaos:
Nice!
Regarding line 97, looks good to me. Original $$ name was obtained by dummy bash version, and looks like validation was only to check being able to write in the containing directory.

oalbrigt:
Ah. So either way is fine then. I'll merge it.

Thanks for your prompt reponse.

PR #1493, opened 2020-05-06
Increase the rabbitmqctl wait timeout during start()

mbaldessari:
After we start the rabbitmq process we wait for the pid to show up
and then declare the server to be started successfully.
This wait is done via 'rabbitmqctl wait'. Now from
From https://www.rabbitmq.com/rabbitmqctl.8.html we have:

  If the specified pidfile is not created or erlang node is not started
  within --timeout the command will fail. Default timeout is 10 seconds.

This default of 10 seconds might not be enough in overloaded
environments. So what we want to do here is wait for as much time as
the start() operation allows us. So we wait for OCF_RESKEY_CRM_meta_timeout
minus 5 seconds. In the rare and non-sensical case that it is less than
10s we do not pass a timeout string at all to rabbitmqctl.

Co-Authored-By: John Eckersberg <jeckersb@redhat.com>

### DIFF
--- a/heartbeat/rabbitmq-cluster
+++ b/heartbeat/rabbitmq-cluster
@@ -294,6 +294,8 @@ rmq_monitor() {
 rmq_init_and_wait()
 {
 	local rc
+	local wait_timeout
+	local timeout_string
 
 	prepare_dir $RMQ_PID_DIR
 	prepare_dir $RMQ_LOG_DIR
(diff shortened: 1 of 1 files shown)


dciabrin:
LGTM

oalbrigt:
LGTM. Thanks.

PR #1494, opened 2020-05-11
Low: galera: Fix handling of Galera nodes when they include a port number

petrpavlu:
When Galera is configured to use a non-default base port, this port number needs to be specified also in `wsrep_cluster_address`. Example:

    wsrep_cluster_address="gcomm://bsc1170646-node1:6666,bsc1170646-node2:6666"

This causes a problem for the Galera resource agent because it splits nodes in `wsrep_cluster_address` on character ',' and then expects to map obtained Galera node names to Pacemaker node names. The mapping is done either directly or indirectly via the `cluster_host_map` but fails in both cases as `cluster_host_map` is also not expected to include any port numbers. A result of the wrong mapping is that the agent invokes `crm_attribute -N <node>` with a non-existent node name and is therefore not able to obtain correct attribute values.

Example of a wrong command invoked by the agent:

    # /usr/sbin/crm_attribute -N bsc1170646-node1:6666 -l reboot --name "rsc_galera-last-committed" --quiet
    Could not map name=bsc1170646-node1:6666 to a UUID

Correct command:

    # /usr/sbin/crm_attribute -N bsc1170646-node1 -l reboot --name "rsc_galera-last-committed" --quiet
    7

A proposed patch addresses the problem by stripping any port number from a Galera node in function `galera_to_pcmk_name()`.

Note: As an alternative approach, handling of `cluster_host_map` could be improved/fixed to accept also port numbers:

    cluster_host_map=bsc1170646-node1:bsc1170646-node1:6666,bsc1170646-node2:bsc1170646-node2:6666

This provides more complete and explicit control. I am not sure if this would be potentially a more preferred approach.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -491,7 +491,7 @@ greater_than_equal_long()
 
 galera_to_pcmk_name()
 {
-    local galera=$1
+    local galera=${1%:*}
     if [ -z "$OCF_RESKEY_cluster_host_map" ]; then
         echo $galera
     else


heartbeat/galera
```diff
@@ -491,7 +491,7 @@ greater_than_equal_long()
 
 galera_to_pcmk_name()
 {
-    local galera=$1
+    local galera=${1%:*}
```
petrpavlu, line 494: Stripping a port number in this way would actually process incorrectly Galera nodes that are defined as '[IPv6-address]'. Such a format is also mishandled by the following `cluster_host_map` mapping code. Fixing that code instead to split a single mapping on the first ':' character should also allow specifying a port number in Galera node names.

dciabrin: The situation of galera with IPv6 is a bit messy. Using IPv6 directly in gcomm or wsrep_listen_addr used to cause problem due to some version of galera not parsing IPv6 properly [1]. The situation is clean nowadays but for sure the galera_to_pcmk_name is another complication.

[1] https://github.com/codership/galera/issues/519

dciabrin:
The problem with cluster_host_map is that I originally replicated a host_map facility from another agent, without considering port to be part of the names. So it's going to be hard to modify the separators without breaking everybody using this option.
I've never tried to us IPv6 in square brackets directly in the wsrep_cluster_address. But if that works, maybe we could explicitely say that IPv6 are only allowed between brackets and strip ports from gcomm address before passing it to galera_to_pcmk_name?

petrpavlu:
I am sorry, I misread the sed command in `galera_to_pcmk_name()`. The split of each mapping is already made only on the first occurence of ':', so both `cluster_host_map="pcmk1:gal1:6666;pcmk2:gal2:6666;pcmk3:gal3:6666"` and `"cluster_host_map=pcmk1:[gal1-ipv6];pcmk2:[gal2-ipv6];pcmk3:[gal3-ipv6]"` should work. I have checked the former in my lab.

The syntax when a port is specified is a bit ugly but if `cluster_host_map` is ok to work in this way then I think no change is needed here.

petrpavlu:
Closing since no change in the resource agent seems to be necessary, as discussed in the previous comment. Apologies for the noise.

PR #1499, opened 2020-05-27
High: docker-compose: new resource agent

kennychennetman:
This new resource agent adds functionality to manage docker services using docker-compose as an OCF resource in an High Availability setup.
It relies on a well-tested docker compose YAML file which distributed on an identical location on all cluster nodes.

### DIFF
--- a/heartbeat/docker-compose
+++ b/heartbeat/docker-compose
@@ -1,0 +1,301 @@
+#!/bin/sh
+# Version: 1.0
+# Date: 2020-05-27
+#
+# Resource script for running docker-compose
+#
+# Description:  Manages docker services using docker-compose as an OCF
+#               resource in an High Availability setup.
+#               It relies on a well-tested docker compose YAML file which
+#               distributed on an identical location on all cluster nodes.
+#
+# Caveat: 1. A YAML file (docker-compose.yml) and an optional Dockerfile
+#            must be provided in a working directory.
+#         2. It is suggested to test run the docker-compose and verify
+#            on all cluster nodes before enabling this agent.
+#
+# docker-compose OCF script's Author: Kenny Chen <netman@study-area.org>
+# License: GNU General Public License (GPL)
+#
+#	usage: $0 {start|stop|status|monitor|validate-all|meta-data}
+#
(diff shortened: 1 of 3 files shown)


oalbrigt:
You should use _default variables to use in metadata and use as defaults for the parameters.

Example: https://github.com/ClusterLabs/resource-agents/commit/191528539317694bd69f64091365d91c3aef82ac

Also CI fails due to `required="true"`, which should be `required="1"`.

heartbeat/docker-compose
```diff
@@ -113,26 +120,15 @@ END
 exit $OCF_SUCCESS
 }
 
-if [ -n "$OCF_RESKEY_binpath" ]; then
+if [ -r "$OCF_RESKEY_binpath" -x "$OCF_RESKEY_binpath" ]; then
```
oalbrigt, line 123: I think you meant to do `if [ -r "$OCF_RESKEY_binpath" ] && [ -x "$OCF_RESKEY_binpath" ]; then` here.

heartbeat/docker-compose
```diff
@@ -0,0 +1,297 @@
...
+	done
+	if [ "$RTV" = false ]; then
+		ocf_log err "failed to kill docker"
+		return $OCF_ERR_GENERIC
+	else
+		RUN=false
+	fi
+}
+
+docker-compose_status()
```
oalbrigt, line 147: It seems like CI fails due to the -'s in the function names (only _'s are supported in function names from what I recall).

oalbrigt:
ok to test

heartbeat/docker-compose
```diff
@@ -0,0 +1,297 @@
+#!/bin/bash
```
oalbrigt, line 1: You need to rename it to `docker-compose.in` and update this to `#!@BASH_SHELL@` and add it to configure.ac for it to detect the bash path properly:
https://github.com/ClusterLabs/resource-agents/blob/master/configure.ac#L1005

Or change it to /bin/sh if all the code is

oalbrigt:
LGTM. Thanks.

PR #1504, opened 2020-06-04
Add error log output

bd233:
1. Add error log output
2. Modify 'grep "$OCF_RESKEY_address"' to 'grep "\s${OCF_RESKEY_address}/"', to solve the problem that the corresponding IROUTE_TO_INSTANCE cannot be obtained when both 10.0.1.2 and 10.0.1.21 exist at the same time


oalbrigt:
ok to test

oalbrigt:
LGTM.

For future PRs please prefix the commits with the agent name they're changing.

PR #1518, opened 2020-06-16
ocf-shellfuncs/pgsql fixes and improvements

oalbrigt:
- ocf_is_ms(): also check OCF_RESKEY_CRM_meta_promotable to make it work w/Pacemaker 2.x
- pgsql: use XML output for better backward and forward compatibility, and "Promoted" keyword to be ready for Pacemaker 2.1

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1040,7 +1040,7 @@ pgsql_replication_monitor() {
 
     # I can't get master node name from $OCF_RESKEY_CRM_meta_notify_master_uname on monitor,
     # so I will get master node name using crm_mon -n
-    print_crm_mon | tr -d "\t" | tr -d " " | grep -q -e "^${RESOURCE_NAME}[(:].*[):].*Master" -e "^\*${RESOURCE_NAME}[(:].*[):].*Master"
+    print_crm_mon | grep -q -i -E "<resource id=\"${RESOURCE_NAME}\" .* role=\"(Promoted|Master)\""
     if [ $? -ne 0 ] ; then
         # If I am Slave and Master is not exist
         ocf_log info "Master does not exist."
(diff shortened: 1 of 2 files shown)


oalbrigt:
Based on discussion with @kgaillot about how to get the most consistant output.

heartbeat/pgsql
```diff
@@ -1040,7 +1040,7 @@ pgsql_replication_monitor() {
 
     # I can't get master node name from $OCF_RESKEY_CRM_meta_notify_master_uname on monitor,
     # so I will get master node name using crm_mon -n
-    print_crm_mon | tr -d "\t" | tr -d " " | grep -q -e "^${RESOURCE_NAME}[(:].*[):].*Master" -e "^\*${RESOURCE_NAME}[(:].*[):].*Master"
+    print_crm_mon | grep -q -E "<resource id=\"${RESOURCE_NAME}\" .* role=\"Promoted|Master\""
```
kgaillot, line 1043: Use -i, I'll probably use lowercase when it changes to "promoted"

heartbeat/pgsql
```diff
@@ -2118,7 +2118,14 @@ check_socket_dir() {
...
     if [ -z "$CRM_MON_OUTPUT" ]; then
-        CRM_MON_OUTPUT=`exec_with_retry 0 crm_mon -n1`
+        ocf_version_cmp "$OCF_RESKEY_crm_feature_set" "3.1.0"
+        res=$?
+        if [ -z "$OCF_RESKEY_crm_feature_set" ] || [ $res -eq 2 ]; then
+            XMLOPT="--output-as=xml"
+        else
+            XMLOPT="--as-xml"
+        fi
+        CRM_MON_OUTPUT=`exec_with_retry 0 crm_mon $XMLOPT`
```
kgaillot, line 2128: probably a good idea to use -1, I think it's the default with xml but who knows if that will change one day

heartbeat/pgsql
oalbrigt: Updated to use -i.

heartbeat/pgsql
oalbrigt: You probably confuse this with crm_mon parameter?

This is for the exec_with_retry() function, which tells it to retry if it fails.
https://github.com/ClusterLabs/resource-agents/blob/8fea11f81fcd50653a002c87ff26e72622031640/heartbeat/pgsql#L1790

oalbrigt: Updated to use crm_mon -1 in case the default changes in the future.

PR #1523, opened 2020-06-24
docker-compose: bugfix

kennychennetman:
1. removed svcname and used prefix name instead, to avoid terminlogogy confusion.
2. improved accuracy of service status determination using 'docker ps' instead of 'docker-compose ps'

### DIFF
--- a/heartbeat/docker-compose
+++ b/heartbeat/docker-compose
@@ -1,6 +1,6 @@
 #!/bin/sh
-# Version: 1.0.3
-# Date: 2020-05-29
+# Version: 1.1.0
+# Date: 2020-06-24
 #
 # Resource script for running docker-compose
 #
@@ -26,7 +26,6 @@
 # OCF_RESKEY_binpath
 # OCF_RESKEY_dirpath
 # OCF_RESKEY_ymlfile
-# OCF_RESKEY_svcname
 #
 ##########################################################################
 # Initialization:
@@ -36,7 +35,10 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
 # Defaults
-OCF_RESKEY_binpath_default=/usr/bin/docker-compose
+OCF_RESKEY_binpath_default=/usr/local/bin/docker-compose
+[ -e $OCF_RESKEY_binpath_default ] || OCF_RESKEY_binpath_default=/usr/bin/docker-compose
+[ -e $OCF_RESKEY_binpath_default ] || OCF_RESKEY_binpath_default=$(which docker-compose 2>/dev/null)
+
 OCF_RESKEY_ymlfile_default=docker-compose.yml
 : ${OCF_RESKEY_binpath=${OCF_RESKEY_binpath_default}}
 : ${OCF_RESKEY_ymlfile=${OCF_RESKEY_ymlfile_default}}
@@ -97,15 +99,6 @@ For example, "docker-compose.yml"
 <content type="string" default="$OCF_RESKEY_ymlfile_default"/>
 </parameter>
 
-<parameter name="svcname" required="1">
-<longdesc lang="en">
-The docker service name.
-For example, "myservice"
-</longdesc>
-<shortdesc lang="en">The docker service name</shortdesc>
-<content type="string"/>
-</parameter>
-
 </parameters>
 
 <actions>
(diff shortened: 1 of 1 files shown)


heartbeat/docker-compose
```diff
@@ -148,12 +141,19 @@ docker_compose_status()
...
 	# use docker-compose ps if YML found, otherwise try docker ps and kill containers
 	if [ -r "$DIR/$YML" ]; then
-		STAT_MSG=$(cd $DIR ; $COMMAND ps)
-		LNWTH=$(echo "$STAT_MSG" | expand | head -n1 | wc -c)
-		STATEPOS=$(echo "$STAT_MSG" | expand | head -n1 | egrep -o 'State.*$' | wc -c)
-		OFFSET=$(($LNWTH-$STATEPOS+1))
-		PSNU=$(echo "$STAT_MSG" | grep "^${SVC}_" | wc -l)
-		UPNU=$(echo "$STAT_MSG" | grep "^${SVC}_" | expand | cut -c ${OFFSET}- | awk '{print $1}' | grep -w 'Up' | wc -l)
+
+                DKPS=$(cd $DIR; $COMMAND ps -q)
```
oalbrigt, line 145: You should use tabs for indentation to stay consistent with the rest of the agent.

heartbeat/docker-compose
```diff
@@ -148,12 +141,19 @@ docker_compose_status()
...
+
+		DKPS=$(cd $DIR; $COMMAND ps -q)
+		STAT_MSG=$(docker ps --no-trunc | grep -E $(echo "$DKPS" | tr '\n' '|') | expand)
+		LNWTH=$(echo "$STAT_MSG" | head -n1 | wc -c)
+		STATEPOS=$(echo "$STAT_MSG" | head -n1 | egrep -o 'STATUS.*$' | wc -c)
 		OFFSET=$(($LNWTH-$STATEPOS+1))
-		PSNU=$(echo "$STAT_MSG" | grep "^${SVC}_" | wc -l)
-		UPNU=$(echo "$STAT_MSG" | grep "^${SVC}_" | expand | cut -c ${OFFSET}- | awk '{print $1}' | grep -w 'Up' | wc -l)
+
+		if [ "$DKPS" ]; then
```
oalbrigt, line 151: This should be `if [ -n "$DKPS" ]; then` (not zero length).

heartbeat/docker-compose
```diff
@@ -36,7 +35,10 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
 # Defaults
-OCF_RESKEY_binpath_default=/usr/bin/docker-compose
+OCF_RESKEY_binpath_default=/usr/local/bin/docker-compose
```
oalbrigt, line 38: You could set this to "" (you shouldnt manipulate the _default values, and the 2nd line below wont ever run.

If you set default to "" and update the metadata to say it auto-detects when not specified you can add the following line after line 43: `[ -z "$OCF_RESKEY_binpath" ] && OCF_RESKEY_binpath="$(which docker-compose 2>/dev/null)"`

heartbeat/docker-compose
kennychennetman: I think it should be okay without the -n operator, according to the man(1) sh:
            string             string has non-zero length.
            -n string          string is not empty.
Anyway, I will put the -n to the code at the next commit. Thanks for your comment.

heartbeat/docker-compose
```diff
@@ -123,16 +113,16 @@ exit $OCF_SUCCESS
 if [ -r "$OCF_RESKEY_binpath" -a -x "$OCF_RESKEY_binpath" ]; then
 	COMMAND="$OCF_RESKEY_binpath"
 else
-	COMMAND=$(which docker-compose 2>/dev/null)
+	COMMAND="$OCF_RESKEY_binpath_default"
 fi
 
 DIR="$OCF_RESKEY_dirpath"
+PRE="$(echo ${DIR##*/} | tr A-Z a-z | sed 's/[^a-z0-9]//g')"
```
oalbrigt, line 120: We will have to do an `if [ -z "$OCF_RESKEY_svcname" ]; then` here, and
else with `PRE="$OCF_RESKEY_svcname"` to avoid it breaking for backwards compatibility.
```

kennychennetman: Well, the svcname is actually a buggy setting and should be replaced by the PREfix version. If just for consideration of backward compatibility, I'd rather to reset it to be inherited from $PRE. Is it okay?

oalbrigt: In my example code it set PRE to the svcname value if it's specified.

kennychennetman: No, the svcname is not equal to the PRE value. I wrongly use the svcname at the earlier version since I misunderstood the naming convention of docker-compose service. I just simply thought the service prefix was the same as we specified in the environment. However it is not the case, I've found the service prefix is from the directory name by removing punctuation with only alphabets and numbers remained. Using the svcname may cause inaccurate status determination to the docker processes and fail the status function.

oalbrigt: OK. I guess we're fine the way it is then.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1524, opened 2020-06-30
[gcp-vpc-move-vip] Support for multiple alias IPs, and metadata improvements

nrwahl2:
If a cluster contains two `gcp-vpc-move-vip` resources, only one can run
on a particular node at a given time. If a second `gcp-vpc-move-vip`
resource starts up on a node where one is already running, the existing
alias IP range is removed before the new one is added.

This places unnecessary limits on functionality. Per the GCP
documentation: "A VM instance virtual interface can have up to 10 alias
IP ranges assigned to it."
  - [Configuring alias IP ranges](https://cloud.google.com/vpc/docs/configure-alias-ip-ranges)

The existing behavior prevents one node from being able to effectively
host two floating IP addresses simultaneously (unless they are in a
contiguous range and can be managed as a single unit, which is
uncommon).

This patch modifies the RA so that it updates the existing list of
alias IP ranges attached to a VM, instead of re-initializing the list
to hold only `OCF_RESKEY_alias_ip`. With these improvements, multiple
`gcp-vpc-move-vip` resources can run on a single node, thus supporting as
many simultaneous alias assignments as GCP allows.

This patch also improves the RA metadata and log messages, making it 
clear that the RA modifies alias IP ranges rather than single addresses.

Resolves: RHBZ#1846724
Resolves: RHBZ#1846726
Resolves: RHBZ#1846732
Resolves: RHBZ#1846733

### DIFF
--- a/heartbeat/gcp-vpc-move-vip.in
+++ b/heartbeat/gcp-vpc-move-vip.in
@@ -22,7 +22,8 @@ import os
 import sys
 import time
 
-OCF_FUNCTIONS_DIR = os.environ.get("OCF_FUNCTIONS_DIR", "%s/lib/heartbeat" % os.environ.get("OCF_ROOT"))
+OCF_FUNCTIONS_DIR = os.environ.get("OCF_FUNCTIONS_DIR", "%s/lib/heartbeat"
+                                   % os.environ.get("OCF_ROOT"))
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

PR #1525, opened 2020-07-01
Fix typos in galera RA

lmiccini:


oalbrigt:
add to whitelist

oalbrigt:
Thanks.

PR #1526, opened 2020-07-02
send_ua/IPv6addr: use sigaction() instead of deprecated siginterrupt()

oalbrigt:
based on https://pubs.opengroup.org/onlinepubs/009695399/functions/siginterrupt.html

Thanks @wenningerk for reviewing and suggesting improvements.


oalbrigt:
retest this please

PR #1532, opened 2020-07-10
Filesystem: Support whitespace in device or directory name

nrwahl2:
Whitespace in a device name (e.g., a CIFS share) or a directory name
breaks resource operations.

One issue is that many of the variable occurrences aren't quoted, so a
string containing whitespace is split into multiple tokens. This is a
problem when the string meant to be passed as a single argument to a
function (e.g., `list_submounts()`).

Another issue involves the parsing of `list_mounts()` output.
`list_mounts()` can pull data from `/proc/mounts`, `/etc/mtab`, or the
`mount` command. `/proc/mounts` and `/etc/mtab` represent spaces within
a field as octal `\040` strings, while `mount` represents them as
literal space characters.

`list_mounts()` had to be modified to output the mount list as three
distinct fields ((`device`, `mountpoint`, `fstype`), separated by tab
characters) regardless of the data source. Parsers of `list_mounts()`
were modified to use tabs as field delimiters.

The for loop in `Filesystem_stop()` also had to become a while loop to
read line-by-line irrespective of spaces. A for loop splits on spaces.

This patch also includes some minor syntax changes for portability.

Resolves: RHBZ#1624591

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -90,7 +90,7 @@ fi
 : ${OCF_RESKEY_force_unmount=${OCF_RESKEY_force_unmount_default}}
 
 # Variables used by multiple methods
-HOSTOS=`uname`
+HOSTOS=$(uname)
 TAB='	'
 
 # The status file is going to an extra directory, by default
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

Issue #1527, opened 2020-07-06
pgsql: does not work properly in RHEL 8.1's pacemaker


inouekazu:
[#1518](https://github.com/ClusterLabs/resource-agents/pull/1518/commits/542f6f92d6901c3d45645c2781216bca815bb79b#diff-76d60d1e50b90cc0eb20f28205d2d671R2121-R2124) fix does not work properly when combined with RHEL8.1's pacemaker.
This is because CRM_FEATURE_SET is **3.2.0**, but ```crm_mon --output-as=xml``` is not implemented.

* The RHEL 8.1's pacemaker is based on Pacemaker-2.0.2, which does not have ```crm_mon --outpu-as=xml``` implemented.
  ```
  # cat /etc/redhat-release
  Red Hat Enterprise Linux release 8.1 (Ootpa)

  # pacemakerd -F
  Pacemaker 2.0.2-3.el8 (Build: 744a30d655)
   Supporting v3.2.0:  generated-manpages agent-manpages ncurses libqb-logging libqb-ipc systemd nagios  corosync-native atomic-attrd acls

  # cibadmin --empty | head -n1
  <cib crm_feature_set="3.2.0" validate-with="pacemaker-3.2" epoch="1" num_updates="0" admin_epoch="0">

  # crm_mon -1 --output-as=xml
  crm_mon: unrecognized option '--output-as=xml'
  crm_mon - Provides a summary of cluster's current state.

  Outputs varying levels of detail in a number of different formats.

  Usage: crm_mon mode [options]
   :
  Modes (mutually exclusive):
   -h, --as-html=value		Write cluster status to the named html file
   -X, --as-xml			Write cluster status as xml to stdout. This will enable one-shot mode.
   -w, --web-cgi			Web mode with output suitable for CGI (preselected when run as *.cgi)
   -s, --simple-status		Display the cluster status once as a simple one line output (suitable for nagios)
   :
  ```
* Note that CRM_FEATURE_SET has been changed from 3.1.0 to 3.2.0 by a backport patch.
  ```
  # rpm -ivh pacemaker-2.0.2-3.el8.src.rpm
  # ls -l /root/rpmbuild/SOURCES/
  total 5748
  -rw-r--r--. 1 root root    8873 Aug 29  2019 001-xmldiffs.patch
  -rw-r--r--. 1 root root   10465 Aug 29  2019 002-failed-monitors.patch
[...]

PR #1533, opened 2020-07-15
pgsql: support RHEL 8.1 pacemaker

inouekazu:
CRM_FEATURE_SET of pacemaker bundled with RHEL 8.1 is 3.2.0,
but crm_mon --output-as=xml is not implemented.

Fixes: https://github.com/ClusterLabs/resource-agents/issues/1527


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1539, opened 2020-08-12
filesystem: mmap search not specific enough

kermat:
The grep command that populates mmap_procs is too unspecific, and will
return false positives from /usr or /etc if you attempt to manage
mount points named /u or /e, respectively.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1540, opened 2020-08-19
gcp-vpc-move-vip: Fix sort for list of dicts in python3

nrwahl2:
python2 sorts a list of dicts of `{'ipCidrRange': <alias>}` correctly by
default. python3 fails with a TypeError:

`TypeError: '<' not supported between instances of 'dict' and 'dict'`

Fix this by using the key parameter of sorted(). python2 also supports
this.


oalbrigt:
LGTM. Thanks.

PR #1547, opened 2020-08-27
galera: Fix automatic recovery when a cluster was not gracefully stopped

petrpavlu:
When selecting a bootstrap node, the Galera resource agent primarily depends on the `safe_to_bootstrap` flag in `grastate.dat`. If none of the nodes have this flag set to 1 then functions `detect_last_commit()` + `detect_first_master()` provide a recovery logic to select the bootstrap node based on each node's last commit, as obtained from `grastate.dat` or `mysqld_safe --wsrep-recover`.

Fix 65f35e9172407e64ded90f29ea8fc0dfca9643e3 introduced a problem for this recovery logic. If a whole cluster is not gracefully stopped, `grastate.dat` on every node contains `safe_to_bootstrap: 0` and `seqno: -1`. Function `detect_safe_to_bootstrap()` then considers each node with this seqno as not suitable for bootstraping and clears the `safe_to_bootstrap` attribute. Nonetheless, functions `detect_last_commit()` + `detect_first_master()` successfully find a bootstrap node, relying on the recovery logic. However, when the promote operation is invoked, function `galera_promote()` runs check `"$(get_safe_to_bootstrap)" = "0"` which fails and prevents the code from writing `safe_to_bootstrap: 1` into `grastate.dat` of the selected node to mark it as a bootstrap node. The end result is that Galera refuses to be started on this node and therefore the whole cluster remains down.

The patch fixes the problem by adjusting `detect_safe_to_bootstrap()` to continue rejecting the nil UUID but accepting again `seqno: -1`.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -604,11 +604,16 @@ detect_safe_to_bootstrap()
         seqno=$(sed -n 's/^seqno:\s*\(.*\)$/\1/p' < ${OCF_RESKEY_datadir}/grastate.dat)
     fi
 
-    if [ -z "$uuid" ] || [ -z "$seqno" ] || \
-       [ "$uuid" = "00000000-0000-0000-0000-000000000000" ] || \
-       [ "$seqno" = "-1" ]; then
+    if [ -z "$uuid" ] || \
+       [ "$uuid" = "00000000-0000-0000-0000-000000000000" ]; then
         clear_safe_to_bootstrap
         return
+    fi
+    if [ "$safe_to_bootstrap" = "1" ]; then
+        if [ -z "$seqno" ] || [ "$seqno" = "-1" ]; then
+            clear_safe_to_bootstrap
+            return
+        fi
     fi
 
     if [ "$safe_to_bootstrap" = "1" ] || [ "$safe_to_bootstrap" = "0" ]; then


petrpavlu:
@dciabrin and @vuntz, could you please have a look at this patch?

dciabrin:
> @dciabrin and @vuntz, could you please have a look at this patch?

@petrpavlu Instead of loosing up constraints introduced in 65f35e9, why not catching more cases in galera_promote? I didn't verify that, but it looks like when promote is called, it might happen that both safe_to_bootstrap: 0 in grastate.dat (due to past failure) and  attribute 'safe_to_bootstrap' is not set (due to 65f35e9 clearing up that attribute in case weird info is detected).

Since we want to forcibly tweak grastate.dat in that case, why not doing so when either  "$(get_safe_to_bootstrap)" = "0" or "$(get_safe_to_bootstrap)" = "" ?

This way we'll keep 65f35e9's logic untouched?

Note: the only time I could reliably end up with safe_to_bootstrap:1 _and_ last_seqno = -1 is when you just bootstrapped a new cluster, and you kill -9 the single node in the cluster. In this case, grastate.dat is not updated with latest seqno, but technically it's still safe to bootstrap the cluster from that node because it has the most recent dataset, so galera is technically correct. Actually in my testing, safe_to_bootstrap seems to be updated only as soon as a 2nd node joins the cluster.
But maybe there are other cases that I didn't stumble upon, so I'd say it's safer to keep 65f35e9 untouched.

petrpavlu:
Thank you @dciabrin for looking at this patch and your comments.

> @petrpavlu Instead of loosing up constraints introduced in [65f35e9](https://github.com/ClusterLabs/resource-agents/commit/65f35e9172407e64ded90f29ea8fc0dfca9643e3), why not catching more cases in galera_promote? I didn't verify that, but it looks like when promote is called, it might happen that both safe_to_bootstrap: 0 in grastate.dat (due to past failure) and attribute 'safe_to_bootstrap' is not set (due to [65f35e9](https://github.com/ClusterLabs/resource-agents/commit/65f35e9172407e64ded90f29ea8fc0dfca9643e3) clearing up that attribute in case weird info is detected).
> 
> Since we want to forcibly tweak grastate.dat in that case, why not doing so when either "$(get_safe_to_bootstrap)" = "0" or "$(get_safe_to_bootstrap)" = "" ?
> 
> This way we'll keep [65f35e9](https://github.com/ClusterLabs/resource-agents/commit/65f35e9172407e64ded90f29ea8fc0dfca9643e3)'s logic untouched?

My understanding is that the `${INSTANCE_ATTR_NAME}-safe-to-bootstrap` attribute (in CIB) can have the following state for each node:
* The attribute is not set. This can mean the following:
  * The `safe_to_bootstrap` flag was not present in `grastate.dat` because the cluster runs Galera < 3.19.
  * The `safe_to_bootstrap` flag was not present in `grastate.dat` because the file is broken.
  * The `safe_to_bootstrap` flag was present in `grastate.dat` but `detect_safe_to_bootstrap()` recognized it potentially as not valid.
* The attribute is set to "0". This means the `safe_to_bootstrap: 0` flag was present in `grastate.dat`, is valid and needs setting to "1" if the node is determined as a bootstrap one.
* The attribute is set to "1". This means the `safe_to_bootstrap: 1` flag was present in `grastate.dat`, is valid and the node should become a bootstrap one.

If `galera_promote()` is updated to add `"$(get_safe_to_bootstrap)" = ""` then the code could try to set `safe_to_bootstrap: 1` in `grastate.dat` also in the three additional cases outlined above, which is probably safer to avoid?

> Note: the only time I could reliably end up with safe_to_bootstrap:1 _and_ last_seqno = -1 is when you just bootstrapped a new cluster, and you kill -9 the single node in the cluster. In this case, grastate.dat is not updated with latest seqno, but technically it's still safe to bootstrap the cluster from that node because it has the most recent dataset, so galera is technically correct. Actually in my testing, safe_to_bootstrap seems to be updated only as soon as a 2nd node joins the cluster.
> But maybe there are other cases that I didn't stumble upon, so I'd say it's safer to keep [65f35e9](https://github.com/ClusterLabs/resource-agents/commit/65f35e9172407e64ded90f29ea8fc0dfca9643e3) untouched.

The patch could be updated as follows to preserve the current behavior for the `safe_to_boostrap: 1` and `seqno: -1` case:

```
    if [ -z "$uuid" ] || \
       [ "$uuid" = "00000000-0000-0000-0000-000000000000" ]; then
        clear_safe_to_bootstrap
        return
    fi
    if [ "$safe_to_bootstrap" = "1" ]; then
        if [ -z "$seqno" ] || [ "$seqno" = "-1" ]; then
            clear_safe_to_bootstrap
            return
        fi
    fi
```

The means that a node with `safe_to_boostrap: 1` and `seqno: -1` is not immediately selected to be a bootstrap node, but it is left to `detect_last_commit()` + `detect_first_master()` to determine the first master based on the last commit on each node.

vuntz:
> @dciabrin and @vuntz, could you please have a look at this patch?

It's been quite some time since I wrote https://github.com/ClusterLabs/resource-agents/commit/65f35e9172407e64ded90f29ea8fc0dfca9643e3, so I honestly can't remember if I was overly cautious by having both checks, or if there was a strong reason for this.

That being said, I like the logic you suggested in your last comment.

dciabrin:
@petrpavlu thanks for the detailed explanation.

> The means that a node with `safe_to_boostrap: 1` and `seqno: -1` is not immediately selected to be a bootstrap node, but it is left to `detect_last_commit()` + `detect_first_master()` to determine the first master based on the last commit on each node.

Ok that makes sense, you convinced me, I'd +1 that.

petrpavlu:
Thanks both, the patch has been updated per the discussion.

dciabrin:
LGTM

PR #1548, opened 2020-08-27
aws-vpc-move-ip: Don't warn for expected scenarios

nrwahl2:
Make log levels more appropriate to the situation. Before this patch, a
normal start looked like this:
```
Operation start for aws-vip (ocf:heartbeat:aws-vpc-move-ip) returned: 'ok' (0)
 >  stderr: Nov 09 02:38:20 INFO: EC2: Moving IP address 10.0.1.65 to this host by adjusting routing table rtb-01b4ea1ae0ec0a4d9
 >  stderr: Nov 09 02:38:20 INFO: monitor: check routing table (API call) - rtb-01b4ea1ae0ec0a4d9
 >  stderr: Nov 09 02:38:22 WARNING: IP 10.0.1.65 not assigned to running interface
 >  stderr: Nov 09 02:38:22 INFO: EC2: Adjusting routing table and locally configuring IP address
 >  stderr: RTNETLINK answers: Cannot assign requested address
 >  stderr: Nov 09 02:38:24 WARNING: command failed, rc 2
 >  stderr: Nov 09 02:38:24 INFO: monitor: check routing table (API call) - rtb-01b4ea1ae0ec0a4d9
```

Now it looks like this:
```
Operation start for aws-vip (ocf:heartbeat:aws-vpc-move-ip) returned: 'ok' (0)
 >  stderr: Nov 09 02:40:43 INFO: EC2: Moving IP address 10.0.1.65 to this host by adjusting routing table rtb-01b4ea1ae0ec0a4d9
 >  stderr: Nov 09 02:40:43 INFO: monitor: check routing table (API call) - rtb-01b4ea1ae0ec0a4d9
 >  stderr: Nov 09 02:40:44 INFO: IP 10.0.1.65 not assigned to running interface
 >  stderr: Nov 09 02:40:44 INFO: EC2: Adjusting routing table and locally configuring IP address
 >  stderr: Nov 09 02:40:46 INFO: monitor: check routing table (API call) - rtb-01b4ea1ae0ec0a4d9
```

Under normal circumstances, the call to `ec2ip_drop()` within
`ec2ip_get_and_configure` should not be required at all. The drop
function deletes the address before the get_and_configure function
immediately re-adds the address. This call could probably be removed
altogether. Instead, I left the call and silenced its output just in
case of unexpected edge cases.

Resolves: RHBZ#1895811

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -270,7 +270,13 @@ ec2ip_monitor() {
 	ocf_log debug "executing command: $cmd"
 	RESULT=$($cmd | grep "$OCF_RESKEY_ip")
 	if [ -z "$RESULT" ]; then
-		ocf_log warn "IP $OCF_RESKEY_ip not assigned to running interface"
+		if [ "$__OCF_ACTION" = "monitor" ] && ! ocf_is_probe; then
+			level="error"
+		else
+			level="info"
+		fi
+
+		ocf_log "$level" "IP $OCF_RESKEY_ip not assigned to running interface"
 		return $OCF_NOT_RUNNING
 	fi
 
@@ -282,11 +288,22 @@ ec2ip_monitor() {
 ec2ip_drop() {
 	cmd="ip addr delete ${OCF_RESKEY_ip}/32 dev $OCF_RESKEY_interface"
 	ocf_log debug "executing command: $cmd"
-	$cmd
+	output=$($cmd 2>&1)
 	rc=$?
+
 	if [ "$rc" -gt 0 ]; then
-		ocf_log warn "command failed, rc $rc"
+		if [ "$__OCF_ACTION" = "start" ]; then
+			# expected to fail during start
+			level="debug"
+		else
+			level="warn"
+		fi
+
+		ocf_log "$level" "command failed, rc $rc"
+		ocf_log "$level" "output/error: $output"
 		return $OCF_ERR_GENERIC
+	else
+		ocf_log debug "output/error: $output"
 	fi
 
 	# delete remaining route-entries if any


gguifelixamz:
Hello,

Could we get rid of the "RTNETLINK" on the delete scenario? I don't it important while debugging or troubleshooting this agent. It is a bit obvious that it will fail while trying to delete something that isn't configure.

Thanks!

nrwahl2:
> Could we get rid of the "RTNETLINK" on the delete scenario? I don't it important while debugging or troubleshooting this agent. It is a bit obvious that it will fail while trying to delete something that isn't configure.

I lean toward keeping (as debug output only) the command's stdout and stderr, in case of any weird behavior that we can't predict (or haven't predicted). And perhaps the `ip addr del` output could contain some message besides that one.

I don't have strong feelings on this question. If we're trying to diagnose something as a bug, we can still add extra debug statements to the RA as needed. Will let @oalbrigt break the tie :)

heartbeat/aws-vpc-move-ip
```diff
@@ -261,10 +265,19 @@ ec2ip_monitor() {
 ec2ip_drop() {
 	cmd="ip addr delete ${OCF_RESKEY_ip}/32 dev $OCF_RESKEY_interface"
 	ocf_log debug "executing command: $cmd"
-	$cmd
+	if [ "$__OCF_ACTION" = "start" ]; then
+		ocf_log debug "$($cmd 2>&1)"
+	else
+		$cmd
+	fi
 	rc=$?
```
nrwahl2, line 273: Need to fix this. Looks like in a start action, `rc` would be set to the return code of `ocf_log`, not the return code from `$cmd 2>&1`.

nrwahl2: Do we even **need** to call `ec2ip_drop()` from within the start function? I guess it helps ensure we get the IP addresses and routing table into a fresh state. I don't think that's necessary under normal circumstances, and it might not be necessary under any circumstances.

Considering going a step further than @gguifelixamz 's suggestion, and getting rid of all output from this function during a start.

nrwahl2: I ended up keeping the `ec2ip_drop()` call and the output (at debug level) just in case of unanticipated edge cases.

`ec2ip_get_and_configure()` **should** never be called if the IP address is configured. The one possible exception I can think of: if the IP address is configured but the link is down. Then we do end up in `get_and_configure() -> drop()`. (Ultimately that start operation will fail though, because the second monitor call within the start function will fail since the link is down.)

heartbeat/aws-vpc-move-ip
```diff
@@ -261,10 +265,17 @@ ec2ip_monitor() {
...
+
 	if [ "$rc" -gt 0 ]; then
-		ocf_log warn "command failed, rc $rc"
+		if [ "$__OCF_ACTION" = "start" ]; then
+			level=debug
+		else
+			level=warn
+		fi
+		ocf_log "$level" "command failed, rc $rc"
+		ocf_log "$level" "output/error: $output"
```
nrwahl2, line 278: There should be no stdout from `ip addr del`; only `stderr` if there's a problem. So I figure, no need to print this unless `rc > 0`.

oalbrigt: Unless you do it at debug-level.

nrwahl2: Sure, and if you'd prefer that, I'll update it.

I don't have a strong opinion. I settled on this because it put all the logging stuff in one place, and because an rc=0 never seems to generate any stdout or stderr (so it generates a logging line that only says "output/error:" every time).

oalbrigt: I'm fine either way.

nrwahl2:
I forgot about this for a while, and came back to it now by customer request. I kept the call to `ec2ip_drop()` and reduced its log level to debug for the start operation.

oalbrigt:
LGTM. Thanks.

PR #1549, opened 2020-08-27
aws-vpc-move-ip: Enable eni lookup for AWS shared networks via RAM

sbotman:
In a shared network pattern where the cluster resides in shared subnets the instance ids of the nodes are not retrievable but the eni ids are and this optional feature gives transparent support in that situation.

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -243,7 +243,7 @@ ec2ip_monitor() {
 				ocf_log debug "executing command: $cmd"
 				ROUTE_TO_INSTANCE="$($cmd)"
 			else
-				cmd="$OCF_RESKEY_awscli --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
+				cmd="$OCF_RESKEY_awscli $region_opt --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
 				ROUTE_TO_INSTANCE="$(execute_cmd_as_role "$cmd" $OCF_RESKEY_routing_table_role)"
 			fi
 			ocf_log debug "Overlay IP is currently routed to ${ROUTE_TO_INSTANCE}"
@@ -332,7 +332,7 @@ ec2ip_get_and_configure() {
 			ocf_log debug "executing command: $cmd"
 			$cmd
 		else
-			cmd="$OCF_RESKEY_awscli --output text ec2 replace-route --route-table-id $rtb --destination-cidr-block ${OCF_RESKEY_ip}/32 --network-interface-id $EC2_NETWORK_INTERFACE_ID"
+			cmd="$OCF_RESKEY_awscli $region_opt --output text ec2 replace-route --route-table-id $rtb --destination-cidr-block ${OCF_RESKEY_ip}/32 --network-interface-id $EC2_NETWORK_INTERFACE_ID"
 			update_response="$(execute_cmd_as_role "$cmd" $OCF_RESKEY_routing_table_role)"
 		fi
 		rc=$?
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -44,6 +44,7 @@ OCF_RESKEY_routing_table_default=""
 OCF_RESKEY_routing_table_role_default=""
 OCF_RESKEY_interface_default="eth0"
 OCF_RESKEY_monapi_default="false"
+OCF_RESKEY_lookup_type_default="InstanceId"
 
 : ${OCF_RESKEY_awscli=${OCF_RESKEY_awscli_default}}
 : ${OCF_RESKEY_profile=${OCF_RESKEY_profile_default}}
@@ -54,6 +55,7 @@ OCF_RESKEY_monapi_default="false"
 : ${OCF_RESKEY_routing_table_role=${OCF_RESKEY_routing_table_role_default}}
 : ${OCF_RESKEY_interface=${OCF_RESKEY_interface_default}}
 : ${OCF_RESKEY_monapi=${OCF_RESKEY_monapi_default}}
+: ${OCF_RESKEY_lookup_type=${OCF_RESKEY_lookup_type_default}}
 
 [ -n "$OCF_RESKEY_region" ] && region_opt="--region $OCF_RESKEY_region"
 #######################################################################
@@ -154,6 +156,17 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 <shortdesc lang="en">Enhanced Monitoring</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_monapi_default}" />
 </parameter>
+
+<parameter name="lookup_type" required="0">
+<longdesc lang="en">
+Name of resource type to lookup in route table.
+"InstanceId"         : EC2 instance ID. (default)
+"NetworkInterfaceId" : ENI ID. (useful in shared VPC setups).
+</longdesc>
+<shortdesc lang="en">lookup type for route table resource</shortdesc>
+<content type="string" default="${OCF_RESKEY_lookup_type_default}" />
+</parameter>
+
 </parameters>
 
 <actions>
@@ -187,7 +200,7 @@ execute_cmd_as_role(){
 
 ec2ip_set_address_param_compat(){
 	# Include backward compatibility for the deprecated address parameter
-	if [ -z  "$OCF_RESKEY_ip" ] && [ -n "$OCF_RESKEY_address" ]; then
+	if [ -z "$OCF_RESKEY_ip" ] && [ -n "$OCF_RESKEY_address" ]; then
 		OCF_RESKEY_ip="$OCF_RESKEY_address"
 	fi
 }
@@ -213,16 +226,24 @@ ec2ip_validate() {
 }
 
 ec2ip_monitor() {
-        MON_RES=""
+	MON_RES=""
+	if [ "${OCF_RESKEY_lookup_type}" = "NetworkInterfaceId" ]; then
+		EC2_ID="$(ec2ip_get_instance_eni)"
+		RESOURCE_TYPE="interface"
+	else
+		EC2_ID="$EC2_INSTANCE_ID"
+		RESOURCE_TYPE="instance"
+	fi
+
 	if ocf_is_true ${OCF_RESKEY_monapi} || [ "$__OCF_ACTION" = "start" ] || ocf_is_probe; then
 		for rtb in $(echo $OCF_RESKEY_routing_table | sed -e 's/,/ /g'); do
 			ocf_log info "monitor: check routing table (API call) - $rtb"
 			if [ -z "${OCF_RESKEY_routing_table_role}" ]; then
-				cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile $region_opt --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].InstanceId"
+				cmd="$OCF_RESKEY_awscli --profile $OCF_RESKEY_profile $region_opt --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
 				ocf_log debug "executing command: $cmd"
 				ROUTE_TO_INSTANCE="$($cmd)"
 			else
-				cmd="$OCF_RESKEY_awscli --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].InstanceId"
+				cmd="$OCF_RESKEY_awscli --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
 				ROUTE_TO_INSTANCE="$(execute_cmd_as_role "$cmd" $OCF_RESKEY_routing_table_role)"
 			fi
 			ocf_log debug "Overlay IP is currently routed to ${ROUTE_TO_INSTANCE}"
@@ -230,8 +251,8 @@ ec2ip_monitor() {
 				ROUTE_TO_INSTANCE="<unknown>"
 			fi
 
-			if [ "$EC2_INSTANCE_ID" != "$ROUTE_TO_INSTANCE" ]; then 
-				ocf_log warn "not routed to this instance ($EC2_INSTANCE_ID) but to instance $ROUTE_TO_INSTANCE on $rtb"
+			if [ "$EC2_ID" != "$ROUTE_TO_INSTANCE" ]; then
+				ocf_log warn "not routed to this $RESOURCE_TYPE ($EC2_ID) but to $RESOURCE_TYPE $ROUTE_TO_INSTANCE on $rtb"
 				MON_RES="$MON_RES $rtb"
 			fi
 			sleep 1
@@ -275,7 +296,7 @@ ec2ip_drop() {
 	return $OCF_SUCCESS
 }
 
-ec2ip_get_and_configure() {
+ec2ip_get_instance_eni() {
 	MAC_FILE="/sys/class/net/${OCF_RESKEY_interface}/address"
 	if [ -f $MAC_FILE ]; then
 		cmd="cat ${MAC_FILE}"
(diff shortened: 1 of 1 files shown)


heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,34 @@ ec2ip_validate() {
...
 ec2ip_monitor() {
-        MON_RES=""
+  MON_RES=""
+  if ocf_is_true ${OCF_RESKEY_eni_id_lookup}; then
+    EC2_ATTRIBUTE_LOOKUP="NetworkInterfaceId"
+    EC2_ID="$(ec2ip_get_instance_eni)"
+    identification="interface"
+  else
+    EC2_ATTRIBUTE_LOOKUP="InstanceId"
+    EC2_ID="$EC2_INSTANCE_ID"
```
oalbrigt, line 234: I think it'd be better to do this if/else section where `EC2_INSTANCE_ID` is defined. That way you wont have to do the EC2_INSTANCE_ID to EC2_ID part of the changes.

Remember to update the indentation of this section to use tabs like in the rest of the agent.

heartbeat/aws-vpc-move-ip
```diff
@@ -300,7 +320,11 @@ ec2ip_get_and_configure() {
 		return $OCF_ERR_GENERIC
 	fi
 	ocf_log debug "network interface id associated MAC address ${MAC_ADDR}: ${EC2_NETWORK_INTERFACE_ID}"
+	echo $EC2_NETWORK_INTERFACE_ID
+}
 
+ec2ip_get_and_configure() {
+  EC2_NETWORK_INTERFACE_ID="$(ec2ip_get_instance_eni)"
```
oalbrigt, line 327: This also needs to be updated to follow the same indentation as the rest of the agent.

oalbrigt:
I've added a couple of suggestions to improve patch.

Can you also prefix the commits with "aws-vpc-move-ip:" so it's easy to see which agent they change from git log?

heartbeat/aws-vpc-move-ip
sbotman: I get what you are saying, but more variables are set based on this check.
So moving this out of here would loose the other vars like identification.
Maybe even better to drop the whole instance-id part because it's not needed, the eni will always work.

oalbrigt: If eni also works for older awscli that sounds like a good idea (we dont want to break it for users who upgrade the agent while having older awscli).

sbotman: What is the lowest version of the awscli that you want to support? I can do some testing because I really think we can drop the instance logic. Have to admit that it looks nicer in the logs, but other than that it's overhead..

oalbrigt: 1.14.28 is the oldest we use in RHEL, but maybe some other distro might have a LTS version that provides even older awsli.

sbotman: so what is your suggestion here? shall we leave it in and be safe on the backwards compatibility or test this with the version 1.14.28 of the aws-cli and remove the code if possible. (looking and the aws repo the NetworkInterfaceId is in there, so should work fine I think)

gguifelixamz: Yep, we need to leave it backward compatible until we decide to put a minimal requirement on the AWS CLI, which as of today isn't in place, but it is a good call.

We also need to ensure that whatever feature we're adding, keeps and most importantly, doesn't break anything.

gguifelixamz:
Hello,

While I find the changes interesting, I would like to know if all test cases being considered and we're still fully backward compatible with all features this agent provides. Some examples:

- Single ENI
- Multiple ENIs
- Single Routing Table
- Multiple Routing Tables

Can you also provide some usage guide on the new parameters in this PR for documentation purposes?

sbotman:
Hello, 

I understand your concerns but this logic didn't change, so single or multiple tables are working fine.
While describing the route tables, it returns the instance-id and the network-interface-id, so you have both by default.

Only when you have 1 AWS account (call it the network account) containing all VPC's and sharing these networks via RAM (resource access manager) to another accounts containing the workloads (EC2 instances) then you don't get the instance-id information. This is because you need to check the route-table by assuming the role in the network account and check the table there, but because the workload is not part of this account, you lose this information.

You do get the network-interface-id information, so by setting the eni_id_lookup=true on the resource creation, you can do the matching on network-interface-id.

Hope this convers the documentation part.
Sander

gguifelixamz:
Thank you Sander, I think that cover all the questions I had! :-)

heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,35 @@ ec2ip_validate() {
 }
 
 ec2ip_monitor() {
-        MON_RES=""
+	MON_RES=""
+	if ocf_is_true ${OCF_RESKEY_eni_id_lookup}; then
+		EC2_ATTRIBUTE_LOOKUP="NetworkInterfaceId"
+    	EC2_ID="$(ec2ip_get_instance_eni)"
```
nrwahl2, line 230: Let's use tabs instead of spaces. I strongly prefer four spaces myself, but it's best to stay consistent with the rest of the script that uses tabs.

heartbeat/aws-vpc-move-ip
```diff
@@ -154,6 +156,15 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 <shortdesc lang="en">Enhanced Monitoring</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_monapi_default}" />
 </parameter>
+
+<parameter name="eni_id_lookup" required="0">
```
nrwahl2, line 160: I'd like to see a lookup_type parameter (doesn't have to use the name "lookup_type", just an example) that expects a string instead of a boolean. Its default value can be "interface_id" or something like that, and it can accept "eni_id" as a configured value. It's semantically more clear than asking "is this an eni_id_lookup or not?".

It's also extensible. That is, if for some reason we ever need a third lookup type, then using a boolean-based approach would get clunky quickly. It would require a new boolean parameter for each new lookup type, and you would have to add logic to decide what should be done if two of them are enabled at the same time (a config mistake). A string-based approach, on the other hand, could simply start accepting the new value.

heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,35 @@ ec2ip_validate() {
...
-        MON_RES=""
+	MON_RES=""
+	if ocf_is_true ${OCF_RESKEY_eni_id_lookup}; then
+		EC2_ATTRIBUTE_LOOKUP="NetworkInterfaceId"
+    	EC2_ID="$(ec2ip_get_instance_eni)"
+    	identification="interface"
+  	else
+    	EC2_ATTRIBUTE_LOOKUP="InstanceId"
+    	EC2_ID="$EC2_INSTANCE_ID"
+    	identification="instance"
```
nrwahl2, line 235: Any reason for mixing cases in the variable names (all caps for the `EC2_*` ones and all lowercase for `identification`)?

heartbeat/aws-vpc-move-ip
```diff
@@ -154,6 +156,15 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 <shortdesc lang="en">Enhanced Monitoring</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_monapi_default}" />
 </parameter>
+
+<parameter name="eni_id_lookup" required="0">
+<longdesc lang="en">
+Use eni id lookup instead of the default interface id method. This is useful in shared subnet setups.
```
nrwahl2, line 162: Is this supposed to say "instance id" instead of "interface id"? My AWS knowledge is limited; I'm not familiar with "ENI" without looking it up and don't fully understand the situation behind this PR. But based on context -- especially the changes to the `ec2ip_monitor()` function -- it seems that the default would be "instance id".

heartbeat/aws-vpc-move-ip
nrwahl2: And this is only a pretty minor point on my part: To me, "identification" sounds like something unique to the resource (instance/interface) that we're looking up. This variable seems more like a "resource_type", or something similar.

heartbeat/aws-vpc-move-ip
oalbrigt: He already fixed that.

nrwahl2: No.
```
$ git clone git@github.com:schubergphilis/resource-agents.git
$ cd resource-agents
$ grep -n '^ ' heartbeat/aws-vpc-move-ip 
230:    	EC2_ID="$(ec2ip_get_instance_eni)"
231:    	identification="interface"
232:  	else
233:    	EC2_ATTRIBUTE_LOOKUP="InstanceId"
234:    	EC2_ID="$EC2_INSTANCE_ID"
235:    	identification="instance"
236:  	fi
328:  	EC2_NETWORK_INTERFACE_ID="$(ec2ip_get_instance_eni)"
```

heartbeat/aws-vpc-move-ip
sbotman: correct, updated

heartbeat/aws-vpc-move-ip
sbotman: There is currently not more useful information coming back from the API call. But I have changed the code, still future additions would not integrate that easy because you will also need metadata information for your check. But I hope this makes it more flexible.

heartbeat/aws-vpc-move-ip
sbotman: updated

heartbeat/aws-vpc-move-ip
```diff
@@ -154,6 +156,15 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 <shortdesc lang="en">Enhanced Monitoring</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_monapi_default}" />
 </parameter>
+
+<parameter name="lookup_type" required="0">
+<longdesc lang="en">
+Name of resource type to lookup in route table. In shared subnet setups use NetworkInterfaceId.
```
nrwahl2, line 162: Nice, this seems cleaner :)

Just for clarity I'm in favor of telling the user what the allowed values for this are. See the `Filesystem` RA's `run_fsck` parameter for an example: https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/Filesystem#L192-L202

See also `vg_access_mode` and `activation_mode` params of the `LVM-activate` RA: https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/LVM-activate#L149-L168

Of course, in this case the user could deduce the two available options from the "use NetworkInterfaceId" statement together with the default value. But IMO best to be clear that those are the full set of supported values. Also, the default value isn't printed by `pcs resource describe <RA>`. (I should file an RFE for that today.)

heartbeat/aws-vpc-move-ip
```diff
@@ -154,6 +156,15 @@ Enable enhanced monitoring using AWS API calls to check route table entry
...
+</longdesc>
+<shortdesc lang="en">lookup type for route table resource</shortdesc>
+<content type="boolean" default="${OCF_RESKEY_lookup_type_default}" />
```
nrwahl2, line 165: Let's change content type to "string" accordingly.

heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,33 @@ ec2ip_validate() {
 }
 
 ec2ip_monitor() {
-        MON_RES=""
+	MON_RES=""
+	if [ "${OCF_RESKEY_lookup_type}" == "NetworkInterfaceId"]; then
+		EC2_ID="$(ec2ip_get_instance_eni)"
+		RESOURCE_TYPE="interface"
+  	else
```
nrwahl2, line 231: There are still a few stragglers that start with spaces :)
```
$ grep -n '^ ' heartbeat/aws-vpc-move-ip 
231:  	else
234:  	fi
326:  	EC2_NETWORK_INTERFACE_ID="$(ec2ip_get_instance_eni)"
```

heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,33 @@ ec2ip_validate() {
...
 			fi
 			ocf_log debug "Overlay IP is currently routed to ${ROUTE_TO_INSTANCE}"
 			if [ -z "$ROUTE_TO_INSTANCE" ]; then
 				ROUTE_TO_INSTANCE="<unknown>"
 			fi
 
-			if [ "$EC2_INSTANCE_ID" != "$ROUTE_TO_INSTANCE" ]; then 
-				ocf_log warn "not routed to this instance ($EC2_INSTANCE_ID) but to instance $ROUTE_TO_INSTANCE on $rtb"
+			if [ "$EC2_ID" != "$ROUTE_TO_INSTANCE" ]; then
+				ocf_log warn "not routed to this $identification ($EC2_ID) but to $identification $ROUTE_TO_INSTANCE on $rtb"
```
nrwahl2, line 253: s/identification/RESOURCE_TYPE/g

sbotman: Oops

heartbeat/aws-vpc-move-ip
```diff
@@ -213,25 +224,33 @@ ec2ip_validate() {
 }
 
 ec2ip_monitor() {
-        MON_RES=""
+	MON_RES=""
+	if [ "${OCF_RESKEY_lookup_type}" == "NetworkInterfaceId"]; then
```
nrwahl2, line 228: s/==/=/

`==` is a bashism.

heartbeat/aws-vpc-move-ip
nrwahl2: Thanks :)

heartbeat/aws-vpc-move-ip
nrwahl2: Need space before closing `]`. CI picked up a syntax error.

heartbeat/aws-vpc-move-ip
sbotman: Think it's my editor playing with me :(
Will fix this.. tnx

heartbeat/aws-vpc-move-ip
```diff
@@ -159,10 +159,12 @@ Enable enhanced monitoring using AWS API calls to check route table entry
 
 <parameter name="lookup_type" required="0">
 <longdesc lang="en">
-Name of resource type to lookup in route table. In shared subnet setups use NetworkInterfaceId.
+Name of resource type to lookup in route table.
+"InstanceId"         : (default)
+"NetworkInterfaceId" : usefull in shared vpc setups.
```
nrwahl2, line 164: Looks good. Minor things:
s/usefull/useful/
s/vpc/VPC/ (no big deal but if we're changing 'usefull', might as well change this too)

I would say:
```
"InstanceId"         : Default value, EC2 instance ID.
"NetworkInterfaceId" : ENI ID (useful in shared VPC setups).
```

The rest of the PR now looks good to me, in terms of code conventions and whatnot :) I don't have the cycles to research and test the overall approach, so I'll trust that it all makes sense, and see if Oyvind has anything to add in that regard.

heartbeat/aws-vpc-move-ip
```diff
@@ -243,7 +243,7 @@ ec2ip_monitor() {
 				ocf_log debug "executing command: $cmd"
 				ROUTE_TO_INSTANCE="$($cmd)"
 			else
-				cmd="$OCF_RESKEY_awscli --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
+				cmd="$OCF_RESKEY_awscli $region_opt --output text ec2 describe-route-tables --route-table-ids $rtb --query RouteTables[*].Routes[?DestinationCidrBlock=='$OCF_RESKEY_ip/32'].$OCF_RESKEY_lookup_type"
```
nrwahl2, line 246: This commit isn't needed. The `execute_cmd_as_role()` function adds the region option. See L187.

nrwahl2: Never mind. Deleting and re-evaluating.

nrwahl2, line 246: Thanks! I was thinking in 324f3ba that `$region_opt` wasn't necessary here because it's applied in `execute_cmd_as_role()`. However, I see now that it's inserted for a separate command in `execute_cmd_as_role()`; it still needs to be applied to `cmd`.

nrwahl2:
The final product all looks good to me.

Since this PR isn't merged yet, we don't need separate commits to fix mistakes that were made while developing this PR. I think a total of two commits makes the most sense:

- `aws-vpc-move-ip: Implemented optional eni lookup instead of the defau…`
- `aws-vpc-move-ip: Fix region option`

The other five current commits should be part of an edited "Implemented optional eni lookup" commit.

Can you squash all the commits (except for the region fix commit, which should be separate) into the original commit? I think you can do that in an interactive git-rebase by moving the region commit to the end of the history and then changing "pick" to "fixup" for all the fix commits.

sbotman:
Tnx @nrwahl2 for all the reviews! 
Really appreciate all the feedback and support getting this forward!

nrwahl2:
@oalbrigt If you have any concerns about the overall approach (or you run it by anyone else who does), let us know. As I mentioned in one of the review comments, I don't have the cycles to research exactly what this feature is doing and test it out right now.

Also need to get f4c8daa into RHBZ#1872999. I misread and thought the `execute_cmd_as_role()` function was already appending the region opt, but it isn't, so we do need this commit from Sander.

sbotman:
I have tested the final script and ran into an interesting use case.

Once you specify that you want to assume a role, the `execute_cmd_as_role()` function is executed. 
Here the `--profile default` (or other profile name if specified) is used. 
Now if you don't have any `~/.aws/config` file with an `[default]` section in there, the command will error out with: `The config profile (default) could not be found` and the assume role will not execute. 

So I think the --profile option should become optional and only added if set.
This will brake backwards compatibility for sure, so didn't change this but I do wanted to mention this here.

Otherwise the script is working fine for me.
Kind Regards,
Sander Botman

oalbrigt:
Right. The parameter is optional (see metadata), so it sounds like that's a mistake in the initial code.

I guess you could update the profile parameter to `--profile $OCF...` when it is set and remove --profile from where it's used in the code.

sbotman:
Since it's not really related to this change, shall we move that into another request?
Because that change could start another long discussion :)

oalbrigt:
Sure :)

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1552, opened 2020-09-11
pgsql: fix support for upper-case hostnames and report exit-reason when failing due to remaining locak file

oalbrigt:


knakahira:
@oalbrigt Thank you for your good work.
But I have one question for upper-case hostname.
The pgsq RA convert upper-case hostnames to lower-case using tr command.
${NODENAME} and application_name are always lower-case even if the real host name is upper-case.
Why do you need that "lower(application_name)"?

oalbrigt:
We have a customer who's somehow able to to make it break with upper-case hostnames without this.

knakahira:
@oalbrigt Could you tell me how the problem happened?
I will check other upper-case hostname problems in the pgsql RA.

oalbrigt:
It's a customer of a customer, so we're having issues getting all the info regarding how they've managed it, but it seems to work fine with this patch.

knakahira:
Hmm, I thought about it, but I didn't know how to cause the problem.
Do you have problems if customer change the application name outside the control of pgsql RA?
Could you tell me what customer shouldn't do with current versions of pgsql RA?

PR #1555, opened 2020-09-18
Filesystem: Default fast_stop to no for RHEL 9+ and for other distros

nrwahl2:
Set OCF_RESKEY_fast_stop_default="no" for RHEL and CentOS major releases
9 and above, and for all other distros.

In the past, this attribute has defaulted to "yes", which has caused a
lot of confusion for users. fast_stop preempts the resource's stop
timeout, causing the agent to give up on unmounting the filesystem after
six seconds and declare a stop failure. (The resource operation does not
time out.)

The existence of a stop operation timeout renders fast_stop unnecessary,
and users typically expect that the agent will keep trying to unmount
the filesystem until the full stop operation timeout expires.

Resolves: RHBZ#1843577


oalbrigt:
LGTM. Thanks.

PR #1564, opened 2020-10-22
SAPInstance - introduce MINIMAL_PROBE resource parameter

fmherschel:
The new parameter MINIMAL_PROBE allows special file system set-ups where you need to mount /usr/sap/<SID> on multiple nodes at the same time. The parameter description states that the customer should only set the parameter to "_true_", if instructed either by support or a vendors document. The default is "_false_", which keeps the probe action unchanged.


oalbrigt:
LGTM. Thanks.

PR #1567, opened 2020-10-26
galera: set bootstrap attribute before promote

dciabrin:
When the master detection takes place, the node chosen for
becoming the master is given two attributes in the CIB:
a master score and a bootstrap flag. The master score makes
pacemaker schedule a promote operation, and the bootstrap flag
drives how the galera server is started.

The order in which we set the attributes is racy; it may happen
that a promote operation is started before the current master
detection function has set the bootstrap flag, in which case
the promotion will fail.

Reverse the order in which we set the attributes on a bootstrap
node to close the race.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -587,8 +587,8 @@ detect_first_master()
     fi
 
     ocf_log info "Promoting $best_node to be our bootstrap node"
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

PR #1568, opened 2020-10-26
galera: fix check user configuration with clustercheck

dciabrin:
The new way of setting parameter defaults for the galera
resource agent has a side effect that broke the automatic
set up of OCF_RESKEY_check_user with clustercheck.

Revert to the original semantics of not not setting
default user to root early, as it prevents clustercheck
credentials to be used. Only set user to root as a fallback
if no other options configured it.


oalbrigt:
LGTM. Thanks.

PR #1569, opened 2020-10-26
galera: recover after network split in a 2-node cluster

dciabrin:
Galera maintains its own quorum, and when a network split
occurs in a two node cluster, both node becomes inquorate.

The resource agent always demotes a node when it loses
galera quorum; however it cannot promote it back because
it waits for the other node to advertise its DB sequence
number in the CIB, and that information is unavailable
during the network split.

Overcome this limitation by telling the resource agent
to use the quorum information from pacemaker. After fencing
took place, the surviving node is the only one active, so
it is safe to restart galera from it. Once the fenced node
comes back online, it won't have quorum until network
is restored, so it cannot restart galera locally so
split-brain is avoided.

This heuristic is made available via a new option
"two_node_mode". By default it is disabled, so the resource
agent works as usual.

### DIFF
--- a/heartbeat/galera
+++ b/heartbeat/galera
@@ -87,11 +87,13 @@ OCF_RESKEY_wsrep_cluster_address_default=""
 OCF_RESKEY_cluster_host_map_default=""
 OCF_RESKEY_check_user_default="root"
 OCF_RESKEY_check_passwd_default=""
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM. Thanks.

PR #1572, opened 2020-11-09
podman: recover from killed conmon side process

dciabrin:
When podman containers are created by the resource-agent, the podman
runtime spawns a side process (conmon) to monitor the container and
record the exit status.

If the conmon process dies unexpectedly (e.g. kill -9), the podman
container can still be stopped, even if the cli returns a generic
error.

Try to distinguish this specific failure condition and make the stop
operation resilient; when it happens, just log a warning and finish
the usual stop actions.

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -419,6 +419,7 @@ podman_start()
 podman_stop()
 {
 	local timeout=60
+	local rc
 	podman_simple_status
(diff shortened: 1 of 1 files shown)


mbaldessari (approved):
lgtm

oalbrigt:
LGTM. Thanks.

PR #1573, opened 2020-11-10
podman: recover from podman's storage being out of sync

dciabrin:
If a system crash while podman is stopping a container (e.g. a fencing action
took place), it might happen that on reboot, podman is not able to recreate
a container as requested by the resource agent.

When such a start operation fails, it might be because the internal storage
layer still references an old container with the same name, even though podman
itself thinks there is no such container. If so, purge the storage layer to try
to clean the corruption and try recreating the container.

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -345,6 +345,32 @@ create_transient_drop_in_dependency()
 }
 
 
+run_new_container()
+{
+	local opts=$1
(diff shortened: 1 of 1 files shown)


mbaldessari (approved):
lgtm, nice work!

oalbrigt:
LGTM. Thanks.

Issue #1574, opened 2020-11-10
Kamailio does not stop properly


amegyeri:
isRunning_cmd implemantation is faulty

PR #1575, opened 2020-11-10
Kamailio stop fix

amegyeri:
Fix for issue #1574


amegyeri:
The original author probably thought that pkill takes the same arguments as kill, but it does not.
In the current format the command does not work as expected. Instead of simply returning the status of the processes, it kills them and returns 0, which is incorrectly interpreted by the resource agent. the RA responds with immediate "stopped" state, even though kamailio is still running. Therefore, the following "start" operation fails.

The fix appears to work fine.

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1576, opened 2020-11-11
configure: improvements to work better with CI

oalbrigt:


Makefile.am
```diff
@@ -185,4 +188,4 @@ if BUILD_RGMANAGER
 endif
 
 clean-generic:
-	rm -rf $(SPEC) $(TARFILES)
+	rm -rf $(pkgconfig_DATA) $(SPEC) $(TARFILES)
```
fabbione, line 191: You shouldn´t clean pkgconfig_DATA here. Since it´s a configuration file (according to AC_CONF_FILES) it will be cleaned only at make maintainer-clean (automatically). The only way to regenerate that file is via ./configure otherwise and that would break make all, make clean, make install.

.gitignore
```diff
@@ -29,8 +29,9 @@ missing
 *.la
 *.lo
 *.loT
-rgmanager/src/resources/fs.sh
-rgmanager/src/resources/oracledb.sh
+rgmanager/src/resources/*
+!rgmanager/src/resources/*.in
+rgmanager/src/resources/Makefile.in
```
fabbione, line 34: You probably want to add .pc file here as well since it´s generated.

fabbione (changes requested):
beside those 2 minor bits, everything else looks ready to go.
Thanks
Fabio

Issue #1579, opened 2020-11-17
ocf.py function id_probe() fails on unset env variable


fmherschel:
In my LAB I have seen that pacemaker does not set the environemnt variable OCF_RESKEY_CRM_meta_interval=0, if a probe is running. While the shell bindings seams to ignore that, the python bindings fails and returns "False" which invokes the RA to tun a normal monitor.

fmherschel:
Try to fit it with PR #1580

oalbrigt:
Thanks. I've merged the PR.

PR #1580, opened 2020-11-17
fix problem with unset OCF_RESKEY_CRM_meta_interval by cluster framework

fmherschel:
In my LAB I have seen  that ocf.py is_probe fails on unset OCF_RESKEY_CRM_meta_interval
See also Issue #1579


oalbrigt:
LGTM. Thanks.

Issue #1582, opened 2020-11-19
Asterisk does not start if some dirs are missing


amegyeri:
In some cases, typically after server restart, some directories like /var/run/asterisk do not exist, and this causes the start operation to fail.
The commited fix works for us since months.

PR #1583, opened 2020-11-19
Asterisk dir fix

amegyeri:
Fix for issue #1582 

If directories are as expected by the RA, no action is done.
If they are not writable, they will be created with proper credentials.

### DIFF
--- a/heartbeat/asterisk
+++ b/heartbeat/asterisk
@@ -344,7 +344,7 @@ asterisk_start() {
     if [ "$astcanary_pid" ]; then
       for i in $astcanary_pid; do ocf_run kill -s KILL $astcanary_pid; done
     fi
-	
+
     for dir in $ASTRUNDIR $ASTLOGDIR $ASTLOGDIR/cdr-csv $ASTLOGDIR/cdr-custom; do
         if [ ! -d "$dir" ]; then
             ocf_run install -d -o $OCF_RESKEY_user -g $OCF_RESKEY_group $dir \
--- a/heartbeat/asterisk
+++ b/heartbeat/asterisk
@@ -346,6 +346,15 @@ asterisk_start() {
     fi
 
     for dir in $ASTRUNDIR $ASTLOGDIR $ASTLOGDIR/cdr-csv $ASTLOGDIR/cdr-custom; do
+	# Check if directories are writable and fix them if they aren't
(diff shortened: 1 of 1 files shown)


heartbeat/asterisk
```diff
@@ -345,6 +345,15 @@ asterisk_start() {
       for i in $astcanary_pid; do ocf_run kill -s KILL $astcanary_pid; done
     fi
 
+    for dir in $ASTRUNDIR $ASTLOGDIR $ASTLOGDIR/cdr-csv $ASTLOGDIR/cdr-custom; do
+	# Check if directories are writable and fix them if they aren't
+        if  ! su -s /bin/sh - $OCF_RESKEY_user -c "test -w $dir" ; then
+            ocf_log warn "Directory $dir is not writable by $OCF_RESKEY_user, attempting to fix."
+			mkdir -p $dir
+			chown $OCF_RESKEY_user: $dir
```
oalbrigt, line 353: You should move your chown line here, and the creation is being done by `install -d` just above it:
https://github.com/ClusterLabs/resource-agents/pull/1583#diff-1fc5dbf5fe478c7910a2675238f986c1011c8250e498f648963c7fb3cb0cd342R365

You should also set group with chown, as there's a parameter for that in the agent too.

amegyeri: Not sure where you meant I should move the chown line.  If I just move it inside the "-d" conditional check it will not be executed as long if the directory exists, but is not writable with the designated user, so it will fail.
As for the group - OK, I will correct it.
Can you please clarify?
Thanks.

oalbrigt: Inside the `if ! su -s /bin/sh - $OCF_RESKEY_user -c "test -w $dir"; then` on line 365.

amegyeri: Thanks, commited changes.

heartbeat/asterisk
```diff
@@ -344,7 +344,7 @@ asterisk_start() {
     if [ "$astcanary_pid" ]; then
       for i in $astcanary_pid; do ocf_run kill -s KILL $astcanary_pid; done
     fi
-
+
```
oalbrigt, line 347: Can you remove this whitespace as well?

amegyeri: Very precise :) done.

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1585, opened 2020-11-25
[build] fix systemd paths when using non standard prefix

fabbione:
this also fixes make distcheck as normal user and
does not install ldirector systemd unit in / by mistake

Signed-off-by: Fabio M. Di Nitto <fdinitto@redhat.com>


oalbrigt (approved):
LGTM. Thanks.

PR #1591, opened 2020-11-28
ocf-distro: Update version format

nrwahl2:
Align ocf-distro's accepted version format with the current
ocf_version_cmp()-accepted format.

See commit b4df0873. I apparently wasn't working with the latest version
of ocf-shellfuncs when creating #1558.

I don't think that the literal backslash should be an allowed character in a
version string, but it's present in ocf_is_ver() currently.


oalbrigt:
LGTM. Thanks.

PR #1592, opened 2020-11-29
LVM-activate: Stop before blk-availability.service

nrwahl2:
If storage services (e.g., iscsi-shutdown.service) stop before an
LVM-activate resource stops, the managed VG may become unavailable. Then
the LVM-activate resource may fail to deactivate the volume group and
thus fail its stop operation.

This commit adds a systemd drop-in "After=blk-availability.service"
directive for resource-agents-deps.target during the LVM-activate start
op. blk-availability includes "After=" directives for other storage
services and thus serves as a convenient wrapper.

blk-availability is not enabled by default, and a "Wants=" drop-in
that's created after Pacemaker starts would not be able to start
blk-availability automatically. So here we also start blk-availability
during LVM_start().

Resolves RHBZ#1902208

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -781,25 +781,27 @@ lvm_status() {
 		return $OCF_NOT_RUNNING
 	fi
 
-	if [ $OCF_CHECK_LEVEL -gt 0 ]; then
-		case "$OCF_CHECK_LEVEL" in
-			10)
-				# if there are many lv in vg dir, pick the first name
-				dm_name="/dev/${VG}/$(ls -1 /dev/${VG} | head -n 1)"
+	case "$OCF_CHECK_LEVEL" in
+		0)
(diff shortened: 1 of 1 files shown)


nrwahl2:
FreeBSD build failed. Not sure what's going on, although I assume it has something to do with the bash shebang.
```
19:11:41 Note: Writing ocf_heartbeat_LVM.7
19:11:41 OCF_ROOT=. OCF_FUNCTIONS_DIR=../../heartbeat ../../heartbeat/LVM-activate meta-data > metadata-LVM-activate.xml
19:11:41 /bin/sh: ../../heartbeat/LVM-activate: not found
19:11:41 gmake[3]: *** [Makefile:695: metadata-LVM-activate.xml] Error 127
```

oalbrigt:
LGTM. Thanks.

PR #1597, opened 2020-12-10
heartbeat/Route: Return OCF_NOT_RUNNING status if iface doesn't exist.

### LINKED ISSUE #1588: "monitor" of ocf:heartbeat:Route fails if interface does not exist
The `route_status` function relies on the interface to exist, and otherwise returns a generic error state:
https://github.com/ClusterLabs/resource-agents/blob/5349c1e0b066966d0b53d8a377660a92d43ee8c1/heartbeat/Route#L229-L247

This breaks the following use case:
1. A: `ocf:heartbeat:iface-vlan` resource, handling `bond_ex.90` on `bond_ex` with `vlan_loose_binding=true`. 
2. B: `ocf:heartbeat:IPaddr2` resource, creating the IP address for `bond_ex.90`. 
3. C: `ocf:heartbeat:Route` resource, setting up the default route to go via `bond_ex.90`. 

Given these resources in one resource set, the following order constraints:
- `start A then start B symmetrical`
- `start B then start C symmetrical`

break, since when the `ocf:heartbeat:Route` monitor command always fails, since `bond_ex.90` will be removed and only `bond_ex` will remain if the `ocf:heartbeat:iface-vlan` resource is stopped. 

Is this intentional? 
If now, a proposed fix would be to check existence of the interface first, and if it does not exist, return `$OCF_NOT_RUNNING`.

olifre:
In combinations with other agents such as heartbeat/iface-vlan,
interfaces on which routes are managed might not exist
if that resource is stopped.
In this case, the status check can safely assume that the route
is not present when the interface does not exist.

fixes #1588

### DIFF
--- a/heartbeat/Route
+++ b/heartbeat/Route
@@ -227,6 +227,15 @@ route_stop() {
 }
 
 route_status() {
+    if [ -n "${OCF_RESKEY_device}" ]; then
+	# Must check if device exists or is gone.
+	# If device is gone, route is also unconfigured.
+	ip link show dev ${OCF_RESKEY_device} >/dev/null 2>&1
+	if [ $? -ne 0 ]; then
+	    # Assume device does not exist, and short-circuit here.
+	    return $OCF_NOT_RUNNING
+	fi
+    fi
     show_output="$(ip $addr_family route show $(create_route_spec) 2>/dev/null)"
     if [ $? -eq 0 ]; then
 	if [ -n "$show_output" ]; then


wenningerk:
test this please

olifre:
It seems the CI did not trigger as expected (or did you mean I should test?). I have given this testing on an actual machine on which it worked as expected :-).

heartbeat/Route
```diff
@@ -227,6 +227,15 @@ route_stop() {
 }
 
 route_status() {
+    if [ -n "${OCF_RESKEY_device}" ]; then
+	# Must check if device exists or is gone.
+	# If device is gone, route is also unconfigured.
+	device_exists="$(ip link show dev ${OCF_RESKEY_device})"
+	if [ ! $? -eq 0 ]; then
```
oalbrigt, line 234: You should use `-ne 0` instead of `! -eq`.

olifre: Thanks, indeed that's more readable. Changed that and force-pushed :smile:.

oalbrigt:
ok to test

heartbeat/Route
```diff
@@ -227,6 +227,15 @@ route_stop() {
 }
 
 route_status() {
+    if [ -n "${OCF_RESKEY_device}" ]; then
+	# Must check if device exists or is gone.
+	# If device is gone, route is also unconfigured.
+	device_exists="$(ip link show dev ${OCF_RESKEY_device})"
```
oalbrigt, line 233: I just realised we dont use device_exist, so you can just run the command here wiithout assignment.

olifre: Indeed, I've dropped the assignment and redirect the output of the command (both `stdout` and `stderr`) to `/dev/null` now since we don't use either. Thanks!

oalbrigt:
Thanks.

oalbrigt:
FYI.

I've changed the code to also return NOT RUNNING when the route doesnt exist: https://github.com/ClusterLabs/resource-agents/pull/1724

olifre:
@oalbrigt Thanks, very helpful! :+1:

PR #1598, opened 2020-12-11
docker-compose: Bugfix:

kennychennetman:
- fix status determination for more than one docker-compose instance

### DIFF
--- a/heartbeat/docker-compose
+++ b/heartbeat/docker-compose
@@ -1,6 +1,7 @@
 #!/bin/sh
-# Version: 1.1.2
-# Date: 2020-06-24
+
+# Version: 1.1.1
+# Date: 2020-12-11
 #
 # Resource script for running docker-compose
 #
@@ -140,7 +141,8 @@ docker_compose_status()
 	if [ -r "$DIR/$YML" ]; then
 
 		DKPS=$(cd $DIR; $COMMAND ps -q)
-		STAT_MSG=$(docker ps --no-trunc | grep -E $(echo "$DKPS" | tr '\n' '|') | expand)
+		STAT_MSG=$(docker ps --no-trunc | expand)
+		PSTAT_MSG=$(docker ps --no-trunc | grep -E $(echo "$DKPS" | tr '\n' '|' | sed 's/|$//') | expand)
 		LNWTH=$(echo "$STAT_MSG" | head -n1 | wc -c)
 		STATEPOS=$(echo "$STAT_MSG" | head -n1 | egrep -o 'STATUS.*$' | wc -c)
 		OFFSET=$(($LNWTH-$STATEPOS+1))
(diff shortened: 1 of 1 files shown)


heartbeat/docker-compose
```diff
@@ -1,6 +1,7 @@
 #!/bin/sh
-# Version: 1.1.2
-# Date: 2020-06-24
+
+# Version: 1.1.1
```
oalbrigt, line 3: You should increase the version nr, not decrease it.

heartbeat/docker-compose
```diff
@@ -36,7 +36,7 @@
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
 # Defaults
-OCF_RESKEY_binpath_default=/usr/bin/docker-compose
+OCF_RESKEY_binpath_default=/usr/local/bin/docker-compose
```
oalbrigt, line 39: You shouldnt change the default path for it. The setting is there so you can set what the path is on your system when it's not in the default location.

heartbeat/docker-compose
```diff
@@ -185,19 +196,25 @@ docker_compose_status()
...
 	docker_compose_status >/dev/null 2>&1
 	retVal=$?
 	# return success if docker service is running
 	[ $retVal -eq $OCF_SUCCESS ] && exit $OCF_SUCCESS
 
+	sleep 10
 	cd $DIR
 	$COMMAND up -d || {
 		ocf_log err "Error. docker-compose returned error $?."
+		cd /tmp
```
oalbrigt, line 212: You shouldnt need to run `cd /tmp` in the agent.

heartbeat/docker-compose
```diff
@@ -185,19 +196,25 @@ docker_compose_status()
 
 docker_compose_start()
 {
+
+	sleep 5
```
oalbrigt, line 200: You shouldnt need to sleep here.

heartbeat/docker-compose
```diff
@@ -185,19 +196,25 @@ docker_compose_status()
...
+
 	docker_compose_validate_all
 	docker_compose_status >/dev/null 2>&1
 	retVal=$?
 	# return success if docker service is running
 	[ $retVal -eq $OCF_SUCCESS ] && exit $OCF_SUCCESS
 
+	sleep 10
```
oalbrigt, line 208: If you need to wait here you need to add a loop checking every 1-2s (and fails when Pacemaker timeouts) to avoid waiting too much in general, and also work for cases where it takes more than 10s.

oalbrigt (changes requested):

kennychennetman:
re-commit via #1694

heartbeat/docker-compose
kennychennetman: fixed

heartbeat/docker-compose
kennychennetman: I want to change the default path aligned to Ubuntu distribution.

heartbeat/docker-compose
kennychennetman: fixed

heartbeat/docker-compose
kennychennetman: yes, agree. I am still working on it... it could be improved in the future version.

heartbeat/docker-compose
kennychennetman: yes, agree. I am still working on it... it could be improved in the future version.

heartbeat/docker-compose
oalbrigt: You should be able to use get_release_id() for that (see ocf-distro file for logic).

PR #1601, opened 2020-12-24
pgsql: make wal receiver check compatible with PostgreSQL >= 11

### LINKED ISSUE #1551: pgsql: check for running wal receiver process is not compatible with PostgreSQL 12
In https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/pgsql#L937 the process table is checked for a running WAL receiver process by means of `grep "[w]al receiver process"`. This doesn't work for PostgreSQL 12 (and possibly 11), because the WAL receiver process now goes by the name of `walreceiver` instead of `wal receiver process`.

A check that is compatible with all PostgreSQL versions looks like this:

```
receiver_parent_pids=`ps -ef | tr -s " " | grep "[w]al\s*receiver" | cut -d " " -f 3`
```

I will submit a PR to address this issue.

antaflos:
In PostgreSQL 11 and later the wal receiver process is named
"walreceiver" instead of "wal receiver process". This change makes the
check for a running wal receiver process compatible with both variants.

Fixes #1551

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -934,7 +934,7 @@ pgsql_wal_receiver_status() {
     local pgsql_real_monitor_status=$1
 
     PID=`head -n 1 $PIDFILE`
-    receiver_parent_pids=`ps -ef | tr -s " " | grep "[w]al receiver process" | cut -d " " -f 3`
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1602, opened 2021-01-05
LVM-activate: to avoid auto activation race condition when the active node restart

zzhou1:
Happy New Year of 2021, everyone! 

In this PR, the first commit is a MUST have to support ACTIVATION_SKIP flag of LV.

The second commit is highly suggested for the better user experience. It 
enforces ACTIVATION_SKIP flag to LV whenever RA LVM-activate
starts or stops.

The reason behind that is, for HA_LVM, auto-activation must be disabled
to avoid the race condition among multiple nodes. In the practice, often
this get forgotten and brings the unexpected trouble shooting effort. To
avoid that, the lv_attr 10th bit(ACTIVATION SKIP flag) is very handy to
simplify the user experience than to manage the out-of-band
auto_activation_volume_list option in lvm.conf. [1]

This involves two lvchange options:
-k,--setactivationskip y|n
-K,--ignoreactivationskip

ACTIVATION_SKIP flag of lv metadata is added since v2_02_99 and is
safe for the metadata compatibility.

[1] https://www.redhat.com/archives/linux-lvm/2020-November/msg00005.html

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -595,8 +595,10 @@ lvm_validate() {
 }
 
 # To activate LV(s) with different "activation mode" parameters
+# Note: the LV "ACTIVATION_SKIP" flag is practical useful in the cluster
+# to avoid the auto-activation race condition.
 do_activate() {
-	do_activate_opt=$1
+	do_activate_opt="--ignoreactivationskip $1"
 
 	if ocf_is_true "$OCF_RESKEY_partial_activation" ; then
 		do_activate_opt="${do_activate_opt} --partial"
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -606,11 +606,25 @@ do_activate() {
 
 	# Only activate the specific LV if it's given
 	if [ -n "$LV" ]; then
+		lv_attr_10=$(lvs --foreign --noheadings -olv_attr $VG/$LV|awk '{print $1}'|cut -c10)
+		if ! [ "$lv_attr_10" = "k" ]; then
+			ocf_log info "set ACTIVATION_SKIP flag to /dev/$VG/$LV"
+			ocf_run lvchange -ky $VG/$LV || return $OCF_ERR_GENERIC
+		fi
+
 		ocf_run lvchange $do_activate_opt ${VG}/${LV}
 		if [ $? -ne $OCF_SUCCESS ]; then
 			return $OCF_ERR_GENERIC
 		fi
 	else
+		lv_attr_10=$(lvs --foreign --noheadings -olv_attr $VG 2>/dev/null|awk '{print $1}'|cut -c10)
+		count_k=$(echo $lv_attr_10|sed 's/[^k]//g' | awk '{ print length }')
+		lv_count=$(vgs --foreign -o lv_count --noheadings ${VG} 2>/dev/null|awk '{print $1}')
+		if ! [ "$count_k" = "$lv_count" ]; then
+			ocf_log info "set ACTIVATION_SKIP flag to all LV(s) under /dev/$VG/"
+			ocf_run lvchange -ky $VG || return $OCF_ERR_GENERIC
+		fi
+
 		ocf_run lvchange $do_activate_opt ${VG}
 		if [ $? -ne $OCF_SUCCESS ]; then
 			return $OCF_ERR_GENERIC


zzhou1:
@teigland may I have your comments on this, if any. Thanks in advance!

oalbrigt:
Feedback from @teigland (via mail):
The user should run lvchange -ky to add the flag to the LV metadata.
The agent doesn't need to do that, the agent never modifies the LVM
metadata.  Also, the user may not want that flag set on the LV (the
agent will work fine with or without the LV flag.)

Setting the flag on the LV would usually be part of configuring lvm to be
managed by pacemaker, although it can also be done at any time.  The agent
can use --ignoreactivationskip for activation regardless of whether the
flag has been set on the LV.

zzhou1:
> Feedback from @teigland (via mail):
> The user should run lvchange -ky to add the flag to the LV metadata.
> The agent doesn't need to do that, the agent never modifies the LVM
> metadata. Also, the user may not want that flag set on the LV (the
> agent will work fine with or without the LV flag.)
> 

Hi @teigland,

The point of the second commit is about user expereince. There are 
a number of incidents over years that people just forget to diable 
auto-activation for HA-LVM. 

Documentation isn't really effective. People could just don't have 
opportunity to read it, or just blindly forget it.

> Setting the flag on the LV would usually be part of configuring lvm to be
> managed by pacemaker, although it can also be done at any time. 

No, prabably not the scope of pacemaker to deal with LVM configuration.
Ideally, it is the scope of the higher orchestration layer who mange pacemaker
cluster to manage ACTIVATION_SKIP flag. Well, still, the same problem to 
educate the higher level developer to undersstand about this ACTIVATION_SKIP
flag. Again DOC doesn't help often.

I do think resource agent is the best place to bridge the gap, to balance this not perfect world :)

> The agent
> can use --ignoreactivationskip for activation regardless of whether the
> flag has been set on the LV.

Yes, this the reason why the first commit is a must-have in my view.

teigland:
On Tue, Jan 05, 2021 at 08:22:32AM -0800, Roger Zhou wrote:
> No, prabably not the scope of pacemaker to deal with LVM configuration.
> Ideally, it is the scope of the higher orchestration layer who mange pacemaker
> cluster to manage ACTIVATION_SKIP flag. Well, still, the same problem to 
> educate the higher level developer to undersstand about this ACTIVATION_SKIP
> flag. Again DOC doesn't help often.
> 
> I do think resource agent is the best place to bridge the gap, to balance this not perfect world :)

It's bad enough in practice to modify the LV that we need to require the
proper configuration.

I forgot to mention earlier that we're likely to add an autoactivation
flag for LVs which will be more appropriate than the activationskip flag.
https://bugzilla.redhat.com/show_bug.cgi?id=1899214

zzhou1:
> It's bad enough in practice to modify the LV that we need to require the proper configuration.

I understand your point. Initially feel it is intrusive a little too, hence split the change to two commits. 

Actually, the big argument is customers are complaining the user experience is not intuitive, the whole cluster stack is too hard to configure, so many glitches here and there. We got to improve that and instead of scare them away.

> I forgot to mention earlier that we're likely to add an autoactivation flag for LVs which will be more appropriate than the activationskip flag. https://bugzilla.redhat.com/show_bug.cgi?id=1899214

Thanks, got your point about `autoactivation` flag. The code still room to add it, though it is crowded already and seems 54 bits are used already.

Anyway, the ultimate question is what to do with the condition "if auto activation is not configured well". There are multiple options:

a) RA continues start by notifying users the LV metadata change, as the "intrusive" approach
b) an ERROR log message to bring the knowledge up front 
c) LVM-activate RA rejects to start 
d) do nothing 

However, all options are facing the same awkward situation, how we deteminate that condition. Given that `autoactivation` flag in place, we have multiple approaches besides "auto_activation_volume_list" which is too hard to determinate the condition in theory. 

In long run, `autoactivation` flag is the right choice to be used to depreciate `auto_activation_volume_list` for the pacemaker cluster alone. That says, if not set `autoactivation` flag, the LVM-activate resource agent will react on it, no matter which option we agree in the end. Well, toward this direction, it makes sense to loop in the community for opinions, and hopefully reach to consensus. I want to bring this up now, instead of defer to the future when LVM2 code is ready for that and likely forget. 
@teigland, @oalbrigt, @nrwahl2, @vvidic, @zhaohem, what do yo think? Any other people migh have idea here?


ps, my perference is the order of the options listed above. I'm fine with David's disagreement with a).
b) is acceptable since the race condition only happens when the active node restart.
c) is not good enough. It might shock users when they uprade to the new version.
d) is really a problem we try to sort it out. It's an invalid option ;)

zhaohem:
If I understand correct, David meaning:
RA shouldn't modify lvm metadata automatically. All metadata modification should be triggered by user intent.

But this rule should be changed according to condition. If user chooses to use LVM-activate, the idea world is he does nothing speical config to make LVM working fine. LVM-activate should do the optimal config by default. 

If RA change the LVM metadata, it should/may output a message to notify user like: 
```
LVM-activate does change metadata by command: xxxx. the reason for this change is xxxx. You can use command xxx to retore it when you don't need it.
```

For the autocativation flag, in my opinion, current solution (roger's patch) is enough to void the issue (https://www.redhat.com/archives/linux-lvm/2020-November/msg00005.html)

PR #1604, opened 2021-01-07
gcp-vpc-move-route, gcp-vpc-move-vip: Parameterize project ID

nrwahl2:
The current route and vip resources fail if the instance is in a shared service project.

Resolves: RHBZ#1913932
Resolves: RHBZ#1913936


mike-19:
can we also parametrize service account ?

oalbrigt:
LGTM. Thanks.

PR #1605, opened 2021-01-10
added real servers threshold settings to ldirectord

spaolo:
I got the need of set thresholds for real server on a heavy loaded system installation.
Patch runned fine in production for several monts, i'm posting it in case migth turn useful to someone else.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1606, opened 2021-01-13

varkoly:

This problem can be easily solved when su will not be called by --login (resp '-')


oalbrigt:
This probably breaks how the agent works for users depending on parameters in the profile.

varkoly:
I'll create new PR.

PR #1607, opened 2021-01-15

varkoly:

Introduce a new OCF instance parameter use_environemnt. Default is yes.
Setting it to no the command will be started by su without loginshell.


heartbeat/anything
```diff
@@ -206,7 +213,12 @@ user="$OCF_RESKEY_user"
 [ -z "$user" ] && user=root
 
 anything_validate() {
-	if ! su - $user -c "test -x $binfile"
+	if [ "$use_environment" = "yes" ]; then
```
oalbrigt, line 216: You should move this in front of the `case` line for all the actions to avoid having duplicate code in the functions.

FYI. You can do `git commit --amend` to update your commit and `git push -f` to force-push to update the commit in your PR.

varkoly:
I've found a more simply solution. Sorry for trouble

oalbrigt:
No worries.

PR #1608, opened 2021-01-18

varkoly:

Write only the pid into the pidfile.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1616, opened 2021-03-02
Filesystem: change force_unmount default to safe for RHEL9+

oalbrigt:


HideoYamauchi:
Hi  oalbrigt,

As a comment on the commit of this fix, RHEL9 and above says that force_unmount defaults to "safe".

However, ocf_version_cmp will return 0 if it is less than "9.0" when compared to "9.0".
Therefore, this fix makes RHEL9 and below "safe".

Does the commit comment for this fix match the content?

Best Regards,
Hideo Yamauchi.

oalbrigt:
Nice catch. Apparently that was a mistake. I'll make a PR to fix it.

HideoYamauchi:
> Nice catch. Apparently that was a mistake. I'll make a PR to fix it.

Hi @oalbrigt,

All right!

Thanks!
Hideo Yamauchi.

PR #1617, opened 2021-03-03
docker-compose: Bugfix handle ymls correctly

jvinolas:
The current docker-compose resource only handles docker-compose.yml file. It doesn't matter what you set when creating resource in pacemaker with **ymlfile** parameter.

Fixed missing **-f $YML** when doing start/stop/monitor actions.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1620, opened 2021-03-10
azure-lb: Redirect stdout and stderr to /dev/null

nrwahl2:
This fixes a regression introduced in commit d22700fc.

When the `nc` listener process created by an `azure-lb` resource attempts to
write to `stdout`, it dies with an `EPIPE` error.

This can happen when random/garbage input is sent to the `nc` listener, as
may happen during a port scan. For example, if the listener is on port
62000, and a client sends some text (e.g.,
`echo test | nc node1 62000`), then the listener attempts to echo "test" to its
`stdout`. This fails with an `EPIPE`.

Prior to commit d22700fc, all output was redirected to the pid file.
This caused its own problems, but it prevented this particular behavior.

The fix is to redirect the listener's `stdout` and `stderr` to `/dev/null`.

Resolves: RHBZ#1937142
Resolves: RHBZ#1937151

### DIFF
--- a/heartbeat/azure-lb
+++ b/heartbeat/azure-lb
@@ -93,12 +93,18 @@ getpid() {
 
 lb_monitor() {
 	if test -f "$pidfile"; then
-		if pid=`getpid $pidfile` && [ "$pid" ] && kill -s 0 $pid; then
-			return $OCF_SUCCESS
-		else
-			# pidfile w/o process means the process died
-			return $OCF_ERR_GENERIC
+		[ "$__OCF_ACTION" = "stop" ] && level="debug" || level="err"
+
+		if pid=$(getpid "$pidfile") && [ -n "$pid" ]; then
+			output=$(kill -s 0 "$pid" 2>&1)
+			mon_rc=$?
+
+			[ -n "$output" ] && ocf_log "$level" "$output"
+			[ "$mon_rc" -eq 0 ] && return $OCF_SUCCESS
 		fi
+
+		# pidfile w/o process means the process died
+		return $OCF_ERR_GENERIC
 	else
 		return $OCF_NOT_RUNNING
 	fi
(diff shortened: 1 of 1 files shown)


heartbeat/azure-lb
```diff
@@ -141,13 +141,13 @@ lb_stop() {
                 stop_timeout=10
         fi
 
-	if lb_monitor; then
+	if lb_monitor >/dev/null 2>&1; then
```
oalbrigt, line 144: Seems like we should redirect the output in lb_monitor() instead.

nrwahl2: I think it's better to keep the stdout/stderr from lb_monitor() when it's a monitor or start operation, and only discard it in the stop operation.

We could check the operation type from within the monitor operation and send the output to /dev/null if it's a stop operation. That gives us another nested "if" though.

oalbrigt: It's cleaner to do it in the function, than redirect in multiple calls.

Just do a `[  ] && ` or or-version, and do it on debug-level, so we dont get "spammed" when we dont need the info.

nrwahl2: There's no spam really. If we get output or error during a monitor operation, then that's probably something we should know about. Under normal circumstances, we shouldn't get any stdout or stderr during the monitor or start.

The only time it regularly occurs when it shouldn't, is during the stop operation. Even during a normal, clean stop, there will be stderr from lb_monitor() (due to `kill -s 0` on a no-longer-existing PID) at some point.

I'll get an update pushed shortly.

wenningerk: Wouldn't opening a fd on /dev/null and setting it as stderr & stdout at the beginning of stop be the simplest?

oalbrigt: It's better to use ocf_log, and set level info (for start-action) and debug (for other actions).

nrwahl2: We don't want to redirect everything to `/dev/null` in the lb_stop() -- only the output/error of the lb_monitor() calls.

oalbrigt:
LGTM. Thanks.

PR #1627, opened 2021-03-18
bundle: run crm_mon without performing validation

dciabrin:
We have a use case in OpenStack where the resource agents run
inside bundles (containerized pacemaker remotes), and we cannot
always guarantee an exact match of pacemaker version on the host
and in containers. This can make crm_mon fail to run because
it may not have the latest version of the schema to validate the
CIB that it's getting.

Add a function crm_mon_no_validation to allow clustered services
like galera, redis and rabbitmq to work when there is a minor
version mismatch between host and containers. This doesn't
impact non-containerized use cases, there is a single version
of pacemaker binaries in this case.

Related-Bug: rhbz#1940363


oalbrigt:
LGTM.

PR #1629, opened 2021-03-22
podman: return OCF_NOT_RUNNING when monitor cmd fails (not running)

oalbrigt:


heartbeat/podman
```diff
@@ -207,8 +207,12 @@ monitor_cmd_exec()
 	if [ $rc -eq 125 ] || [ $rc -eq 126 ]; then
 		rc=$OCF_NOT_RUNNING
 	elif [ $rc -ne 0 ]; then
-		ocf_exit_reason "monitor cmd failed (rc=$rc), output: $out"
-		rc=$OCF_ERR_GENERIC
+		if ocf_is_probe; then
+			rc=$OCF_NOT_RUNNING
```
dciabrin, line 211: So there has been a sort of API change in podman w.r.t return code.
In podman 1.6.x, running exec on a non-running container returned 126. Since podman 2.x, the return code is now 255.

I'd prefer that we explicitly return OCF_NOT_RUNNING only when the error code is 255, because that is an expected error. Any other error is potentially an unexpected container state, and we should erturn OCF_ERR_GENERIC to give the resource a chance to get demoted and stopped properly.

PR #1632, opened 2021-04-11
openstack-info: CLI output parsing and storage fixes.

bdperkin:
- Updates due to API output format changes.
- Attempt to future-proof parsing of the output.


oalbrigt:
LGTM. Thanks.

PR #1633, opened 2021-04-11
openstack-virtual-ip: New resource agent for virtual IP support in OpenStack.

bdperkin:
- New resource agent for virtual IP support in OpenStack.

### DIFF
--- a/heartbeat/openstack-virtual-ip
+++ b/heartbeat/openstack-virtual-ip
@@ -1,0 +1,280 @@
+#!/bin/sh
+#
+#
+# OCF resource agent to move a virtual address in an Openstack tenant.
+#
+# Copyright (c) 2018 Mathieu GRZYBEK
+# Based on code of Markus Guertler
+# All Rights Reserved.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.  Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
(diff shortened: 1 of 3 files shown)


heartbeat/openstack-virtual-ip
```diff
@@ -0,0 +1,275 @@
...
+		exit $OCF_SUCCESS
+		;;
+esac
+
+if ! ocf_is_root; then
+	ocf_log err "You must be root for $__OCF_ACTION operation."
+	exit $OCF_ERR_PERM
+fi
+
+osvip_validate
```
oalbrigt, line 260: You'll have to use exit in the validate function, or the result will be ignored when action isnt validate-all.

heartbeat/openstack-virtual-ip
```diff
@@ -0,0 +1,275 @@
...
+		return $OCF_ERR_GENERIC
+	fi
+
+	osvip_monitor
+	if [ $? != $OCF_NOT_RUNNING ]; then
+		ocf_log error "Couldn't unset IP address $OCF_RESKEY_ip."
+		return $OCF_ERR_GENERIC
+	fi
+
+	ocf_log info "Successfully brought unset $OCF_RESKEY_ip"
```
oalbrigt, line 206: This sounds a bit strange to me, and I think detached is a better word for unset in the situation.

oalbrigt (changes requested):

heartbeat/openstack-virtual-ip
bdperkin: Fixed with commit 61aebd2c482b0052b2865fba01acc11bc0fc5fab:
```
-osvip_validate
-
 case $__OCF_ACTION in
 	start)
-		osvip_start;;
+		osvip_validate
+		osvip_start
+		;;
 	stop)
 		osvip_stop;;
 	monitor)
 		osvip_monitor;;
 	validate-all)
-		exit $?;;
+		osvip_validate;;
```

heartbeat/openstack-virtual-ip
bdperkin: Fixed with commit 61aebd2c482b0052b2865fba01acc11bc0fc5fab:
```
-	ocf_log info "Successfully brought unset $OCF_RESKEY_ip"
+	ocf_log info "Successfully brought down $OCF_RESKEY_ip"
```

heartbeat/openstack-virtual-ip
bdperkin: I actually changed this again (once I realize I broke the agent with the change).  Now, osvip_validate is still removed from the main and has been added to the beginning of the monitor function.  The reason for this is that start|stop|monitor all require the openstack CLI and the resource file that the validation checks.  The sourcing of the resource file also happens in the validation function, so by adding validation to monitoring and all the appropriate functions call monitoring, the proper checks and environment variables are now set.

oalbrigt (approved):

oalbrigt:
LGTM. Thanks.

PR #1636, opened 2021-04-12
openstack-cinder-volume: CLI output parsing fixes, fetch of node ID consistency, monitor action simplification, and proper use of osvol_validate.

bdperkin:
- Reduce getting node/instance ID methods down to one via openstack-info (same as all other openstack agents)
- Fix order of operations in checking for OCF_RESKEY_openrc
- Updates due to API output format changes.
- Attempt to future-proof parsing of the output.

### DIFF
--- a/heartbeat/openstack-cinder-volume
+++ b/heartbeat/openstack-cinder-volume
@@ -27,6 +27,7 @@
 # Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
 #
 
+
 #######################################################################
 # Initialization:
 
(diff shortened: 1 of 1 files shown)


bdperkin:
@mgrzybek and @varkoly : would you mind reviewing my changes to see if there are any problems that you foresee?  I removed a lot of code to clean-up and simplify, but I may have missed reasons the code exists.

mgrzybek:
Nice patch! This piece of code looks smarter than mine.

I added the *local check option* to speed up `osvol_monitor`:
* when the *cinder* API is slow or crashed, the resource crashes too even if the *cinder* volume is already attached to the instance. Scanning `/dev/disk` was a quick-win option
* using `/dev/disk/by-id/virtio-$short_volume_id` is a bad idea because some Openstack providers do not map disks this way. For example, OVHCloud *cinder* disks look like `/dev/disk/by-id/QEMU-scsi-$short_volume_id`
* I preferred to trust the *cinder* volume short id instead of the virtual disk path provided by the API, ie. `/dev/vd[a-z]`

Have you encountered some weird behaviours such as slow symbolic links updates in `/dev/disk/by-id` (udev) ?

oalbrigt:
LGTM. Thanks.

PR #1637, opened 2021-04-14
openstack-floating-ip: Proper use of osflip_validate and small log message fixes.

bdperkin:
- Use osflip_validate in start & validate-all actions and fix OCF_SUCCESS ocf_log info to make sense.


oalbrigt:
LGTM. Thanks.

PR #1642, opened 2021-04-28
gcp-vpc-move-vip.in: Adds retries

kj1724:
If the cluster fails a monitoring event, it will try to restart the resource. If the resource agent makes an API/metadata call that fails at that time, the resource will be considered "failed", but in certain case also "unconfigured", which prevents further operations.

These changes can help the agent recover on certain intermittent failures.


oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

PR #1643, opened 2021-04-29
LVM-activate: Disable VG autoactivation in system_id access_mode

ganghe:
The autoactivation property of volume group will be disabled when
vg_access_mode is set to system_id, to fix HA-LVM systemid race
condition, which introduces an error "active on 2 nodes".
For HA-LVM systemid use case, the node might introduce the
unexpected outage of the cluster service. In details, the node
might suffer from some race condition and reports "active on 2 nodes"
on one node after reset the active node. This will cause the active
service to be restarted which is unexpected.

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -115,6 +115,9 @@ if using the cluster to manage at least one of them.  If you manage some manuall
 the stop action of the lvmlockd agent may fail and the node may get fenced,
 because some DLM lockspaces might be in use and cannot be closed automatically.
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

zzhou1:
Great patch! Given `--setautoactivation` flag landed at LVM2 [1], this patch is much appropriate to simplify the user experience for the configuration which is often fogotten.

[1] https://sourceware.org/git/?p=lvm2.git;a=commit;h=0a28e3c44b

oalbrigt:
LGTM. Thanks.

Issue #1645, opened 2021-05-11
VirtualDomain doesn


tparkercbn:

tparkercbn:
The NodeUtilization resource on KVM servers (Enterprise Linux) sets host_memory not hv_memory but the VirtualDomain resource only allows the setting of cpu and hv_memory.  Is there a way to add host_memory to the list so that we can use NodeUtilization with VirtualDomain resources?

oalbrigt:
This is what the utilization_hv_memory and utilization_hv_memory_reservation parameters are for in the NodeUtilization agent.

tparkercbn:
NodeUtilization for hv_memory only works for xen hypervisors.   On kvm it sets hv_memory to 0.   What is the issue setting host_memory requirements in VirtualDomain?

Tom

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

oalbrigt:
It seems like you can use the `autoset_utilization_hv_memory` parameter in the `VirtualDomain` agent to do this:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/VirtualDomain#L229-L236

tparkercbn:
HI Oyvind,

I am running CentOS with KVM, on my servers:

VirtualDomain sets hv_memory utilization only (not host_memory).

NodeUtilization sets hv_memory for the node to 0 and host_memory to the correct value.

These two resource agents are therefore incompatible on a KVM server although they will work find if you are using Xen.  I believe that the correct fix is to have VirtualDomain also set host_memory to exactly the same value that is already being used for hv_memory.  It looks like about 4 lines of code in the update_utilization function will do it unless there is something I am missing.

Tom

oalbrigt:
Seems like you just need to set `utilization_hv_memory=false` for the NodeUtilization to avoid this issue.

PR #1649, opened 2021-05-12
Added code to set the host_memory value for Utilization in the VirtualDomain RA

tparkercbn:
This is required to make it compatible with the values set by the NodeUtilization resource agent when running on a kvm based Hypervisor (RHEL, CentOS)

On CentOS, 
NodeUtilization can set available host_memory and cpu for the node but sets hv_memory to 0
VirtualDomain only sets cpu and hv_memory usage which means these two RAs cannot be used together.

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -25,6 +25,7 @@ OCF_RESKEY_migration_network_suffix_default=""
 OCF_RESKEY_force_stop_default=0
 OCF_RESKEY_monitor_scripts_default=""
 OCF_RESKEY_autoset_utilization_cpu_default="true"
+OCF_RESKEY_autoset_utilization_host_memory_default="true"
 OCF_RESKEY_autoset_utilization_hv_memory_default="true"
 OCF_RESKEY_migrateport_default=$(( 49152 + $(ocf_maybe_random) % 64 ))
 OCF_RESKEY_CRM_meta_timeout_default=90000
@@ -45,6 +46,7 @@ OCF_RESKEY_start_resources_default="false"
 : ${OCF_RESKEY_force_stop=${OCF_RESKEY_force_stop_default}}
 : ${OCF_RESKEY_monitor_scripts=${OCF_RESKEY_monitor_scripts_default}}
 : ${OCF_RESKEY_autoset_utilization_cpu=${OCF_RESKEY_autoset_utilization_cpu_default}}
+: ${OCF_RESKEY_autoset_utilization_host_memory=${OCF_RESKEY_autoset_utilization_host_memory_default}}
 : ${OCF_RESKEY_autoset_utilization_hv_memory=${OCF_RESKEY_autoset_utilization_hv_memory_default}}
 : ${OCF_RESKEY_migrateport=${OCF_RESKEY_migrateport_default}}
 : ${OCF_RESKEY_CRM_meta_timeout=${OCF_RESKEY_CRM_meta_timeout_default}}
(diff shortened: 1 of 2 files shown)


tparkercbn:
This fixes issue #1645

tparkercbn:
> It seems like you just need to set utilization_hv_memory=false for the NodeUtilization agent to be able to use it without this patch.

This also has to be done.  But what sets the VirtualDomain utilization for host_memory? 

if I set utilization_hv_memory=false and utilization_cpu=false I get nodes that have correct utilization values set for host_memory (which I want).  However, without this patch I don't have the corresponding requirements from the VirtualDomain side that says how much my VMs are going to consume from the available host_memory.  This patch fixes that.

oalbrigt:
It's not set based on Xen-specific info (it's set based on /proc/meminfo):
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/NodeUtilization#L162-L164

tparkercbn:
But hv_memory is set using xentool so I need host_memory.  Which IS set in NodeUtilization but not in VirtualDomain.   So it needs to be added to VirtualDomain.

Host_Total_Memory() { local xentool xentool=$(which xl 2> /dev/null || which xm 2> /dev/null) if [ -x "$xentool" ]; then "$xentool" info | awk '/total_memory/{printf("%d\n",$3);exit(0)}' else ocf_log debug "Can only set hv_memory for Xen hypervisor" echo "0" fi}

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

oalbrigt:
You can use NodeUtilization with utilization_hv_memory=false and hv_memory will be left for VirtualDomain to take care of.

tparkercbn:
They are not doing the same thing.

NodeUtilization is setting how much is available on the node.

VirtualDomain is setting how much an individual vm will consume.

Do you have an EL based system you can test on?  What I am trying to accomplish is to have placement_strategy=utilization based on available node memory and consumed vm memory.

Get Outlook for Android<https://aka.ms/AAb9ysg>

________________________________

tparkercbn:
Good afternoon.  Any update on this?

oalbrigt:
What's the point of setting the same value for host_memory and hv_memory?

tparkercbn:
Hi Oyvind.  

To make it work with NodeUtilization on KVM hosts.  Otherwise they have no variables in common and don't work together at all. 

In NodeUtilization why do they set hv_memory and host_memory.  In theory they are also the same value and I can see no other resource agent that uses that consumes host_memory at all.  Ideally there should only be one variable representing available machine memory and the same variable representing consumed VM memory but that's not the current case and I will assume that removing one of the duplicate variables (host_memory or hv_memory) will break someone elses system so that Is not an option.  For the hv_memory case both the Xen and VirtualDomain RAs use hv_memory.

If you think that the code as is should work on a KVM host please tell me what I need to set to have NodeUtilization report the available memory on a host and to have VirtualDomain set the amount of consumed memory for a VM using the same variables so that placement-strategy=utilization will work.  

I have tried the available combinations with the code as it is in the repository and they don't have a variable in common UNLESS you use the Xen hypervisor.  If you use XEN then everything works as expected with hv_memory.  If you use kvm then it doesn't work at all.

oalbrigt:
ok to test

oalbrigt:
LGTM. Thanks.

zzhou1:
Hoops, a bit awkward to reply this after more a year :(

First of all, it is the valid and great use case to have VirtualDomain and NodeUtilization work together for KVM.

However, VirtualDomain::host_memory of this patch really creates unnecessary confusion and complexity for users, when and how to use host_memory with the existing hv_memory concept. It creates us(upstream) unnecessary DOC effort to tell the difference between them.  For example, as of now, for KVM workload, user need understand NodeUtilization::utilization_hv_memory=false and VirtualDomain::autoset_utilization_hv_memory=false. Similar extra confusion for Xen worload. 

To pursue the simplest user experience, it's better to keep the single one concept, hv_memory only,  for both KVM and Xen workload. Actually, hv_memory is used by VirtualDomain for both KVM and XEN workload for very long time without using NodeUtilization. 

We agree NodeUtilization does help to simplify Xen load since the beginning. However, yes, it didn't work well for KVM workload before this patch. 

With the above context in mind, I aggressively suggest to

a) revert this commit to get ride of host_memory for VirtualDomain to simplify terminology of the clusterlabs' world, which is too complex already. 

b) the better approach is to fix NodeUtilization :: set_utilization() -> Host_Total_Memory(), and to make it populate the proper hv_memory for KVM workload, which equals host_memory actually.

c) If we can be more aggressive, it's even better to get ride of NodeUtilization::host_memory, for extreme user experience simplicity, just because of a)

oalbrigt:
We cant remove it now, as that'll break setups for users using the new functionality.

zzhou1:
> We cant remove it now, as that'll break setups for users using the new functionality.

Thanks for your attention, Oyvind! I do understand your concern.

@tparkercbn what do you see?

Even given that to keep host_memory in VirtualDomain for the time being,

- As the compromise said in the other day, how about suggestion b)?  Basically, by doing so, hv_memory supports both KVM and XEN workload to use NodeUtilization. The key benefit is to simplify user experience seamlessly without reading DOC and/or RA code.
 
- the minimum document about the difference is desirable. It really saves users' effort to figure them out on their own ;(

Issue #1647, opened 2021-05-12
[Question] About confirmation of "resource-agents-deps.target" at the start of LVM-activate resource.


HideoYamauchi:
Hi All,

The LVM-activate resource makes the decision to create 99-LVM-activate.conf at start.
But isn't it correct to perform this check with "resource-agents-deps.target" instead of "resource-agents-deps.target.d"?

```
(snip)
lvm_start() {
        if systemd_is_running ; then
        	# Create drop-in to deactivate VG before stopping
		# storage services during shutdown/reboot.
		after=$(systemctl show resource-agents-deps.target.d \
			--property=After | cut -d'=' -f2)

		case "$after" in
			*" blk-availability.service "*)
				;;
			*)
				systemd_drop_in "99-LVM-activate" "After" \
					"blk-availability.service"
				;;
		esac
(snip)
```

After actually generating 99-LVM-activate.conf, the result is as follows.

If you specify "resource-agents-deps.target.d", blk-availability.service is not checked.

```
# systemctl show resource-agents-deps.target.d --property=After | cut -d'=' -f2


```

If you specify "resource-agents-deps.target", blk-availability.service is checked.
```
# systemctl show resource-agents-deps.target --property=After | cut -d'=' -f2
blk-availability.service
```

Best Regards,
Hideo Yamauchi.

oalbrigt:
Nice catch. I'll do some testing next week and make a patch.

HideoYamauchi:
Hi Oyvind,

Okay!
I will wait for your fix.

Many thanks,
Hideo Yamauchi.

oalbrigt:
You are correct, so I'll make a patch to fix this.

Thank you for reporting it.

HideoYamauchi:
Hi Oyvind,

All right!

Many thanks,
Hideo Yamauchi.

nrwahl2:
Thanks @HideoYamauchi. My original patch with the typos ended up causing a really weird issue with `nfsserver` resources. I'm not sure how. But the fix proposed here resolves it.
  - https://bugzilla.redhat.com/show_bug.cgi?id=1972035

HideoYamauchi:
Hi Reid,

> Thanks @HideoYamauchi. My original patch with the typos ended up causing a really weird issue with `nfsserver` resources. I'm not sure how. But the fix proposed here resolves it.
> 
> * https://bugzilla.redhat.com/show_bug.cgi?id=1972035

Okay!
[...]

PR #1652, opened 2021-05-20
LVM-activate: fix drop-in check to avoid re-creating drop-in file when it already exists

oalbrigt:
Fixes https://github.com/ClusterLabs/resource-agents/issues/1647

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -820,17 +820,14 @@ lvm_start() {
         if systemd_is_running ; then
         	# Create drop-in to deactivate VG before stopping
 		# storage services during shutdown/reboot.
-		after=$(systemctl show resource-agents-deps.target.d \
-			--property=After | cut -d'=' -f2)
+		systemctl show resource-agents-deps.target \
+			--property=After | cut -d'=' -f2 | \
+			grep -qE "(^|\s)blk-availability.service(\s|$)"
 
-		case "$after" in
-			*" blk-availability.service "*)
-				;;
-			*)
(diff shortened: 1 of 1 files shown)


heartbeat/LVM-activate
```diff
@@ -820,17 +820,14 @@ lvm_start() {
...
-			--property=After | cut -d'=' -f2)
-
-		case "$after" in
-			*" blk-availability.service "*)
-				;;
-			*)
-				systemd_drop_in "99-LVM-activate" "After" \
+		systemctl show resource-agents-deps.target \
+			--property=After | cut -d'=' -f2 | \
+			grep -qE "(^|\s)blk-availability.service(\s|$)"
```
lge, line 825: `\s` may or may not be supported as shorthand for `[[:space:]]`, depending on the version of grep.
Why not use an actual space? Or even use the same shell "case" logic as before, but like this:
`case " $after " in *" ... "*)` (note the extra space around `$after`)

oalbrigt: The shell logic before only matched if you got 3 or more hits (and blk... wasnt first or last). We use \s (and the combination with beginning/end of line) with grep in other agents, and noone has complained about it.

nrwahl2: Yeah, I had used the `case` idiom in 79fb4b2d as an attempt at portability -- I don't remember the details, but I think I had found something less than ideal about grep at the time. `-q` is also not as portable as `>/dev/null 2>&1`.

But none of that mattered, since my two typos (`"$after"` instead of `" $after "`, and `target.d` instead of `target`) broke the whole thing anyway :)

PR #1658, opened 2021-06-14
Google Cloud serviceaccount file support

oalbrigt:


oalbrigt:
There seems to be a bug in the older implementation when using routes().list() and possible other routes() sub functions, so the fallback for old versions seems to only work correctly with the -vip agent:
https://github.com/ClusterLabs/resource-agents/blob/20915c0f10c6b2089189584b7971f2594cd7ed56/heartbeat/gcp-vpc-move-route.in#L284

The issue is on RHEL8, so maybe there's a bugfix upstream I havent found to fix it.

PR #1659, opened 2021-06-15
podman: workaround race during container creation

dciabrin:
podman and OCI runtime have a race that sometimes causes
a container to fail to be created and run [1] if the
cgroup to be used is not available yet. When that happens,
try to recreate it until it succeeds or the start
timeout is reached.

[1] https://bugzilla.redhat.com/show_bug.cgi?id=1972209

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -358,8 +358,18 @@ run_new_container()
 	local rc
 	
 	ocf_log info "running container $CONTAINER for the first time"
-	ocf_run podman run $opts $image $cmd
+	out=$(podman run $opts $image $cmd 2>&1)
 	rc=$?
+
+	if [ -n "$out" ]; then
+		out="$(echo "$out" | tr -s ' \t\r\n' ' ')"
+		if [ $rc -eq 0 ]; then
+			ocf_log info "$out"
+		else
+			ocf_log err "$out"
+		fi
+	fi
+
 	if [ $rc -eq 125 ]; then
 		# If an internal podman error occurred, it might be because
 		# the internal storage layer still references an old container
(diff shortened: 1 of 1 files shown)


dciabrin:
PR created, I'm still verifying the last revision of this patch locally on my environment.

heartbeat/podman
```diff
@@ -370,6 +380,24 @@ run_new_container()
 		ocf_run podman rm --storage $CONTAINER
 		ocf_run podman run $opts $image $cmd
 		rc=$?
+	elif [ $rc -eq 127 ]; then
+		# rhbz#1972209: podman 3.0.x seems to be hit by a race
+		# where the cgroup is not yet set up properly when the OCI
+		# runtime configures the container. If that happens, recreate
+		# the container as long as we get the same error code or
+		# until start timeout preempts us.
+		while [ $rc -eq 127 ] && (echo "$out" | grep -q "cgroup.*scope not found") ; do
```
cjeanner, line 389: shouldn't we have some limit here, to prevent endless loop? It shouldn't happen, but some more security wouldn't hurt.

dciabrin: Pacemaker itself preempts this infinite loop as soon as the start operation timeout is reached. This is usually the idiom we use in the resource agents to loop while we're allowed.

dciabrin:
OK I verified locally that this patch recovers from that race:

```
Jun 16 09:25:41 controller-2 podman(galera-bundle-podman-2)[7499]: INFO: running container galera-bundle-podman-2 for the first time
Jun 16 09:25:47 controller-2 podman(galera-bundle-podman-2)[8466]: ERROR: Error: OCI runtime error: container_linux.go:370: starting container process caused: process_linux.go:459: container init caused: process_linux.go:422: setting cgroup config for procHooks process caused: Unit libpod-9a7a86ceba0dfdc2357f799e1b259846a9e51cd7537a2a3d67d548ad58f28c66.scope not found.
Jun 16 09:25:48 controller-2 podman(galera-bundle-podman-2)[8479]: WARNING: Internal podman error while assigning cgroup. Retrying.
Jun 16 09:25:54 controller-2 podman(galera-bundle-podman-2)[9195]: INFO: 555e280bbfa9d3afdd739b45aad700c8ef73539f60cf6937521690f39d0ef3c4
Jun 16 09:25:58 controller-2 podman(galera-bundle-podman-2)[9431]: INFO: Creating drop-in dependency for "galera-bundle-podman-2" (555e280bbfa9d3afdd739b45aad700c8ef73539f60cf6937521690f39d0ef3c4)
Jun 16 09:26:02 controller-2 podman(galera-bundle-podman-2)[9778]: NOTICE: Container galera-bundle-podman-2  started successfully
Jun 16 09:26:02 controller-2 pacemaker-controld[3233]:  notice: Result of start operation for galera-bundle-podman-2 on controller-2: ok
```

dciabrin:
I just pushed a new revision to get rid of a spurious trailing whitespace, but the patch is still the same so it should be good to go.

mbaldessari (approved):
lgtm

oalbrigt:
LGTM. Thanks.

Issue #1660, opened 2021-06-18
All CTDB  nodes (the service) restarts when one node is hard shutdown


dev-ddoe:
Hello,

I did some tests on a cluster composed of 3 nodes.
This cluster use a CTDB resource defined with the command:
``
ctdb ocf:heartbeat:CTDB ctdb_recovery_lock="/mnt/nfs/lock" --clone interleave=true
``
When I drop connections of a node or when I hard shutdown a node, all the ctdbs node restarts.

During my investigation, I found that ctdb monitor have timed out on getting status of the node down.
CTDB monitor is invoked using the block:
``
	local timeout=3
	local timelimit=120
	# ...but we override with the timeout for the current op:
	if [ -n "$OCF_RESKEY_CRM_meta_timeout" ]; then
		timeout=$((OCF_RESKEY_CRM_meta_timeout/1000))
		timelimit=$((OCF_RESKEY_CRM_meta_timeout/1000))
	fi
	$OCF_RESKEY_ctdb_binary --socket=$OCF_RESKEY_ctdb_socket \
		-t $timeout -T $timelimit \
		"$@"
``

According to the manpage; 
``
-t TIMEOUT
           Indicates that ctdb should wait up to TIMEOUT seconds for a response to most commands
           sent to the CTDB daemon. The default is 10 seconds.

       -T TIMELIMIT
           Indicates that TIMELIMIT is the maximum run time (in seconds) for the ctdb command.
           When TIMELIMIT is exceeded the ctdb command will terminate with an error. The default
           is 120 seconds.
``

The **-t** defines the timeout of each operation with a node (checking the state of one node)
The **-T** defined the time allowed to the ctdb command.

So I think that **timeout** have to be different and less than the **timelimit**
The block mentioned above was done by https://github.com/ClusterLabs/resource-agents/pull/14/commits/7bd548c0088224d8c99c1ebb1289cc931a9ec598 at a time when the default timeout was 3s.

When I remove the $timeout parameter in order to let it to 10s (CTDB default), my scenario works

Thank you,
David.

PR #1661, opened 2021-06-21
CTDB: replace timeout override w/ctdb_timeout parameter

dev-ddoe:
Timelimit is enough to manage resource timeoutin CTDB module
#1660


oalbrigt:
We probably want to set it to a specific value to avoid any issues if the default is different in different versions of CTDB.

I'm waiting for feedback from our CTDB devs about which value we should use.

obnoxxx:
I tend to agree that 

* having an explicit default value here
* and still having a way to override it 

might be better.

What I don't quite understand from the original code is that both the timeout and the timelimit are overridden with the same value.

oalbrigt:
That was probably just a mistake, when they tried to set it based on the timeout value of the action:
https://github.com/ClusterLabs/resource-agents/commit/7bd548c0088224d8c99c1ebb1289cc931a9ec598

oalbrigt:
Sounds good. I guess we could add it as a parameter, and have it default to 10.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1670, opened 2021-07-07
gcp-ilb: add to Makefiles

varkoly:


oalbrigt:
You should also add it to doc/man/Makefile.am, so that the manpage gets generated.

Also can you update the commit message to something like "gcp-ilb: add to Makefiles"?

oalbrigt:
Can you run `git update-index --chmod=+x heartbeat/gcp-ilb` and add it to the commit? If not it fails e.g. `make rpm` since it's not got the executable flag set.

oalbrigt:
Can you update the name in the agent to gcp-ilb?
https://github.com/ClusterLabs/resource-agents/blob/0edeffbd95cbaab87871d8f0137e17854c15a28a/heartbeat/gcp-ilb#L73

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1671, opened 2021-07-20
NVMe-oF-Target: New RAs nvmet-subsystem nvmet-namespace nvmet-port

Philipp-Reisner:
Three new resource agents that allow to operate NVMe-oF Targets.
You need to have a NVMeSubsystem RA first, followed by namespace(s)
and port(s). In order to enable failover you need to set the UUID
on the namespace. I guess it is a good idea to set the nguid (on
the namespace) and the serial (on the subsystem) as well.

### DIFF
--- a/heartbeat/nvmet-namespace
+++ b/heartbeat/nvmet-namespace
@@ -1,0 +1,205 @@
+#!/bin/sh
+#
+#
+#     NVMe-oF (rdma, tcp, fc) OCF RA. Exports and manages NVMe targets.
+#
+#   (c) 2021 LINBIT HA-Solutions GmbH, written by Philipp Reisner
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.  Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+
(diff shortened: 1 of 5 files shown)


heartbeat/NVMePort
```diff
@@ -0,0 +1,228 @@
...
+	return $OCF_SUCCESS
+}
+
+NVMePort_monitor() {
+	[ -d ${portdir} ] || return $OCF_NOT_RUNNING
+	return $OCF_SUCCESS
+}
+
+NVMePort_validate() {
+	if [[ ! ${OCF_RESKEY_type} =~ ^(tcp|rdma|fc)$ ]]; then
```
oalbrigt, line 177: If you replace these with case-statements you can update the `#!` to sh, which is preferred, and avoid renaming them to .in files and adding to configure.

oalbrigt (changes requested):
The agents should follow the naming convention (e.g. nvme-...). Make sure to update it in the metadata section as well.

https://github.com/ClusterLabs/resource-agents/blob/master/doc/dev-guides/ra-dev-guide.asc#is-there-a-naming-convention

They also need to be added to `heartbeat/Makefile.am` and `doc/man/Makefile.am` to generate/install the manpages and agents.

oalbrigt:
ok to test

heartbeat/nvmet-port
```diff
@@ -0,0 +1,238 @@
...
+	cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="nvmet-port">
+<version>0.9</version>
+
+<longdesc lang="en">
+Manages NVMe-oF ports. An NVMe-oF port is part of an NVMe-oF target.
+A port must be created after the subsystem. It exposes one or multiple
+subsyste(s) including the enclused namespace(s) to the network.
```
oalbrigt, line 55: A missing m in subsystem(s) here.

oalbrigt: Can you update the commit message to include the new names as well?

oalbrigt (approved):

heartbeat/nvmet-namespace
```diff
@@ -0,0 +1,205 @@
...
+esac
+
+# Everything except usage and meta-data must pass the validate test
+nvmet_namespace_validate
+
+case $__OCF_ACTION in
+start)		nvmet_namespace_start;;
+stop)		nvmet_namespace_stop;;
+monitor|status)	nvmet_namespace_monitor;;
+reload)		ocf_log err "Reloading..."
```
oalbrigt, line 195: These should be info. No reason to log as error for something we expect.

oalbrigt:
LGTM. Thanks.

PR #1677, opened 2021-08-04
[build] fix out-of-tree build for man pages

fabbione:
the main problem is that some agents are built in builddir, while
others will stay in srcdir. This include ocflibs and some helper
scripts.

when building man pages, we can only point to one dir for OCF_ROOT
and related scripts, and in this case it has to be builddir.

Hack around this OCF limitations by using symlinks into the builddir
for agents that are not "built".

Signed-off-by: Fabio M. Di Nitto <fdinitto@redhat.com>


fabbione:
@oalbrigt this is just one way to fix it.. feel free to change as you like.

oalbrigt (approved):

oalbrigt:
Thanks.

PR #1678, opened 2021-08-05
VirtualDomain: drop prefix xenmigr from migrate uri

ganghe:
Since 'xenmigr' scheme was no longer supported by libvirt, we
should remove the prefix 'xenmigr' from migrate uri.
Otherwise, there will be an error when migrate VM.

### DIFF
--- a/heartbeat/VirtualDomain
+++ b/heartbeat/VirtualDomain
@@ -854,7 +854,7 @@ mk_migrateuri() {
 				echo "tcp:${migrate_target}:${OCF_RESKEY_migrateport}"
 				;;
 			xen)
-				echo "xenmigr://${migrate_target}"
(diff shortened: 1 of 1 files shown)


oalbrigt:
Will this also work with older versions of libvirt?

ganghe:
> Will this also work with older versions of libvirt?
Hello Oyvind,
please refer to the link below, it looks compatiable.
https://github.com/virt-manager/virt-manager/commit/561f5cd3e68fa3f1fb6745463a5c1a486171d8c9#diff-0e12fb33fe64939787586f0476754d30a17cbc546f8275d7a51544dd457762d0

oalbrigt:
Great. Thanks.

Issue #1680, opened 2021-08-09
SAPInstance fails to detect systemd integration


fmherschel:
SAPInstance does start sapstartsrv directly by starting the process, even if the service is directly integrated into systemd using unit files.

oalbrigt:
Here's an example of how to detect systemd (and services in it) in the agent:
https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/nfsserver#L236-L256

fmherschel:
We can close this issue as commit 5ad5025 and 9a85152 are solving this problem.

PR #1681, opened 2021-08-09
SAPInstance: Fix for issue #1680 - SAPInstance fails to detect systemd integration

fmherschel:
SAPInstance failed to detect the systemd integration, because path of systemctl was not correct. In the fixed version the RA now uses "which" to detect the correct path.


oalbrigt:
Thanks.

PR #1422, opened 2019-10-23
Low: mysql-common: fix startup check

vvidic:
PID value is not captured correctly so the startup
fails with the wrong exit code.

Starting 'mysql' case 8 'check lib file':
Setting agent environment:    export OCF_RESKEY_CRM_meta_timeout=15000
Setting system environment:   chmod u-w /var/lib/mysql
Running agent:                ./mysql start
ERROR: The agent was hanging, killed it, maybe you damaged the agent or system's environment, see details below:
Oct 23 18:46:06 INFO: MySQL is not running
runuser: warning: cannot change directory to /nonexistent: No such file or directory
runuser: warning: cannot change directory to /nonexistent: No such file or directory
runuser: warning: cannot change directory to /nonexistent: No such file or directory
Oct 23 18:46:06 INFO: MySQL is not running
Oct 23 18:46:08 INFO: MySQL is not running
Oct 23 18:46:10 INFO: MySQL is not running
Oct 23 18:46:12 INFO: MySQL is not running
Oct 23 18:46:14 INFO: MySQL is not running
Oct 23 18:46:16 INFO: MySQL is not running
Oct 23 18:46:18 INFO: MySQL is not running
Oct 23 18:46:20 INFO: MySQL is not running
Oct 23 18:46:22 INFO: MySQL is not running
Oct 23 18:46:24 INFO: MySQL is not running

PR #1682, opened 2021-08-11
MySQL: add Support for local SSL connection

igolman:
Added support for local MySQL connection over SSL.

**Use Case:**

once a `REQUIRE  SSL` for MySQL user is set, the cluster will not be able to manage its MySQL resource any further.

A good practice is to require SSL for all MySQL external users. A replication user is one of those.  A replication by itself may _(and should)_ be configured over SSL, however the enforcement for replication user to use SSL may be occasionally not set.
This patch allows to configure MySQL resource to use SSL _(by default is set to `false` - so no running setups will be affected)_.

**_How To:_**
1. Enforce SSL for MySQL user _(MySQL v5.7+)_:
```
mysql> show variables like 'version';
+---------------+-----------------------------+
| Variable_name | Value                       |
+---------------+-----------------------------+
| version       | 5.7.35-0ubuntu0.18.04.1-log |
+---------------+-----------------------------+
1 row in set (0,00 sec)

mysql> SHOW CREATE USER 'repl'@'%';
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| CREATE USER for repl@%                                                                                                                                           |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| CREATE USER 'repl'@'%' IDENTIFIED WITH 'mysql_native_password' AS 'XXXXXXXXXX' REQUIRE NONE PASSWORD EXPIRE DEFAULT ACCOUNT UNLOCK |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0,00 sec)

mysql> ALTER USER 'repl'@'%' REQUIRE SSL;
Query OK, 0 rows affected (0,02 sec)

mysql> FLUSH PRIVILEGES;
Query OK, 0 rows affected (0,02 sec)

mysql> SHOW CREATE USER 'repl'@'%';
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| CREATE USER for repl@%                                                                                                                                           |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
| CREATE USER 'repl'@'%' IDENTIFIED WITH 'mysql_native_password' AS 'XXXXXXXXXX' REQUIRE SSL PASSWORD EXPIRE DEFAULT ACCOUNT UNLOCK |
+------------------------------------------------------------------------------------------------------------------------------------------------------------------+
1 row in set (0,00 sec)
```
2. Update clusters `mysql` resource to use SSL:
```
pcs resource update mysql replication_require_ssl="true"
```

### DIFF
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -96,8 +96,7 @@ MYSQL_BINDIR=`dirname ${OCF_RESKEY_binary}`
 # Convenience variables
 
 MYSQL=$OCF_RESKEY_client_binary
-if [ "$OCF_RESKEY_replication_require_ssl" = "true" ]
-then
+if ocf_is_true "$OCF_RESKEY_replication_require_ssl"; then
   MYSQL_OPTIONS_LOCAL_SSL_OPTIONS="--ssl"
 fi
 MYSQL_OPTIONS_LOCAL="-S $OCF_RESKEY_socket"
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -98,6 +98,8 @@ MYSQL_BINDIR=`dirname ${OCF_RESKEY_binary}`
 MYSQL=$OCF_RESKEY_client_binary
 if ocf_is_true "$OCF_RESKEY_replication_require_ssl"; then
   MYSQL_OPTIONS_LOCAL_SSL_OPTIONS="--ssl"
+else
+  MYSQL_OPTIONS_LOCAL_SSL_OPTIONS=""
 fi
 MYSQL_OPTIONS_LOCAL="-S $OCF_RESKEY_socket"
 MYSQL_OPTIONS_REPL="$MYSQL_OPTIONS_LOCAL_SSL_OPTIONS $MYSQL_OPTIONS_LOCAL --user=$OCF_RESKEY_replication_user --password=$OCF_RESKEY_replication_passwd"
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -49,6 +49,7 @@ OCF_RESKEY_additional_parameters_default=""
 OCF_RESKEY_replication_user_default="root"
 OCF_RESKEY_replication_passwd_default=""
 OCF_RESKEY_replication_port_default="3306"
+OCF_RESKEY_replication_require_ssl_default="false"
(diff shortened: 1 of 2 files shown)


heartbeat/mysql-common.sh
```diff
@@ -239,8 +245,8 @@ mysql_common_start()
     --datadir=$OCF_RESKEY_datadir \
     --log-error=$OCF_RESKEY_log \
     $OCF_RESKEY_additional_parameters \
-    $mysql_extra_params >/dev/null 2>&1" &
-    pid=$!
+    $mysql_extra_params >/dev/null 2>&1 &
+    pid=$!"
```
oalbrigt, line 249: You shouldnt change this:
https://github.com/ClusterLabs/resource-agents/pull/1422/files

igolman: Done.

heartbeat/mysql-common.sh
```diff
@@ -94,8 +96,12 @@ MYSQL_BINDIR=`dirname ${OCF_RESKEY_binary}`
 # Convenience variables
 
 MYSQL=$OCF_RESKEY_client_binary
+if [ "$OCF_RESKEY_replication_require_ssl" = "true" ]
```
dciabrin, line 99: This should probably use shell function ocf_is_true() to handle uppercase/lowercase

heartbeat/mysql-common.sh
```diff
@@ -94,8 +96,12 @@ MYSQL_BINDIR=`dirname ${OCF_RESKEY_binary}`
...
+then
+  MYSQL_OPTIONS_LOCAL_SSL_OPTIONS="--ssl"
```
dciabrin, line 101: Could you set the else branch to "" as well? this way bash won't complain if we run the agent with bash -u

heartbeat/mysql-common.sh
igolman: 👍 Done.

oalbrigt:
ok to test

heartbeat/mysql-common.sh
igolman: 👍 Done.

oalbrigt:
Thanks.

oalbrigt:
I've patched it to work correctly with newer MySQL versions as well (and checked that it should still work for v5.7): https://github.com/ClusterLabs/resource-agents/pull/1737

Issue #1644, opened 2021-05-07
NFS: __nfs4_reclaim_open_state: Lock reclaim failed!


lge:
Problem: linux nfsd initializes part of its "server scope" based on the nodename, see [nfsd_name](https://github.com/torvalds/linux/blob/a48b0872e69428d3d02994dcfad3519f01def7fa/fs/nfsd/netns.h#L179) and [nfsd_svc](https://github.com/torvalds/linux/blob/a48b0872e69428d3d02994dcfad3519f01def7fa/fs/nfsd/nfssvc.c#L784).

After a failover to an other node with different nodename, that breaks the linux nfs 4 client lock recovery:
`_nfs4_proc_exchange_id: server_scope mismatch detected`
after that it does not even attempt to recover lost locks, unless you set the nfs module parameter recover_lost_locks=Y, which you should not, as that may lead to data corruption in other situations.

To demonstrate the problem, with sufficiently recent linux distributions / kernels, you do not even need a cluster, or failover.
You need: NFS Server, NFS Client, `flock` (a real test program would use fcntl; modern linux nfs handles it just fine, though), and `dd` [and a root shell ;-)]

    client# mount -t nfs4 -o vers=4.1 $server:/ /mnt
    client# ( exec 7<>/mnt/testfile; flock -x 7; while date | dd conv=fsync >&7 ; do sleep 1; done )
    
    server# echo 0 > /proc/fs/nfsd/threads
    server# unshare -u sh -c 'hostname other-name; echo 8 > /proc/fs/nfsd/threads'
 
 That should quickly lead to  `dd: fsync failed for 'standard output': Input/output error`  on the client.
 
If you compare with rgmanager nfsserver, you see we [set "uts nodename"](https://github.com/ClusterLabs/resource-agents/blob/5c54175d39a81344535977a475b015f6ec6b123a/rgmanager/src/resources/nfsserver.sh.in#L224), since 2016, for good reason.

Now, we know the problem, we know the solution, how are we going to implement it?
Nowadays, with systemd, just calling systemctl from an other uts namespace won't do.

I see several options:
 * `rpc.nfsd` learns a new option to set the scope / nodename from command line
 * admin replaces `rpc.nfsd` with some wrapper that does the `unshare --uts` + `hostname` before `exec`ing the real one.
 * admin overrides the ~~init script~~ systemd unit to do the same
 * pacemaker resource agent nfsserver learns to generate that override

The latter could be implemented like this:

[...]

PR #1688, opened 2021-09-06
nfsserver: Fix NFSv4 lock failover: set NFS Server Scope

lge:
Problem: https://github.com/ClusterLabs/resource-agents/issues/1644
RFC8881, 8.4.2.1 State Reclaim:

| If the server scope is different, the client should not attempt to
| reclaim locks. In this situation, no lock reclaim is possible.
| Any attempt to re-obtain the locks with non-reclaim operations is
| problematic since there is no guarantee that the existing
| filehandles will be recognized by the new server, or that if
| recognized, they denote the same objects. It is best to treat the
| locks as having been revoked by the reconfiguration event.

That's why for lock reclaim to even be attempted, we have to define and set
the same server scope for NFSD on all cluster nodes in the NFS failover
cluster. And in linux, that is done by setting the uts nodename for the
command that starts the nfsd kernel threads.

For "init scripts", just set it directly using unshare --uts.
For systemd units, add NFS_SERVER_SCOPE to some environment files
and inject the "unshare --uts" into the ExecStart command lines
using override drop-in files.

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -697,7 +697,7 @@ inject_unshare_uts_name_into_systemd_units ()
 		test -d "$dir" || mkdir -p "$dir"
 		test -e "$dropin" && rm -f "$dropin"
 
-		edited_exec_start=$(systemctl cat $svc | sed -ne "s#^ExecStart=\\(.*\\)#ExecStart=/usr/bin/unshare --uts /bin/sh -ec 'hostname \${NFS_SERVER_SCOPE}; exec \"\$@\"' -- \\1#p")
+		edited_exec_start=$(systemctl cat $svc | sed -ne "s#^ExecStart=\\([-+:!@]*\\)\\(.*\\)#ExecStart=\\1/usr/bin/unshare --uts /bin/sh -c 'hostname \${NFS_SERVER_SCOPE}; exec \"\$@\"' -- \\2#p")
 		cat > "$dropin" <<___
 [Service]
 EnvironmentFile=$SYSTEMD_ENVIRONMENT_FILE_NFS_SERVER_SCOPE
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -687,6 +687,8 @@ inject_unshare_uts_name_into_systemd_units ()
 	services=$(systemctl list-unit-files --no-legend $NFSD_RELATED_SYSTEMD_SERVICE_FOR_UNSHARE_UTS_NAMESPACE | cut -d ' ' -f1)
 
 	local svc dir dropin edited_exec_start do_reload=false
+	local old_umask=$(umask)
+	umask 0022
 	for svc in $services ; do
 		dir=/run/systemd/system/$svc.d
 		dropin=$dir/$SYSTEMD_UNSHARE_UTS_DROPIN
@@ -710,6 +712,7 @@ ___
 
 	mkdir -p "${SYSTEMD_ENVIRONMENT_FILE_NFS_SERVER_SCOPE%/*}"
 	echo "NFS_SERVER_SCOPE=$OCF_RESKEY_nfs_server_scope" > "$SYSTEMD_ENVIRONMENT_FILE_NFS_SERVER_SCOPE"
+	umask $old_umask
 
 	$do_reload && systemctl daemon-reload
 }
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -12,7 +12,7 @@
 # did the unshare for gssd and idmapd as well, even though it seems unclear why.
 # Let's start with just the nfs-server, and add others if/when we have clear
 # indication they need it.
-#NFSD_RELATED_SYSTEMD_SERVICE_FOR_UNSHARE_UTS_NAMESPACE="nfs-idmapd.service nfs-mountd.service nfs-server.service nfsdcld.service rpc-gssd.service rpc-statd.service rpcbind.service"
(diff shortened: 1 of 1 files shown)


lge:
It took me long enough :-/ So this is my "request for review and test", please.
Specifically:

- do some other commands need the `unshare --uts`, besides the `ExecStart=` lines?
- is the list of units in `NFSD_RELATED_SYSTEMD_SERVICE_FOR_UNSHARE_UTS_NAMESPACE` correct? complete? which ones should be dropped?
- is there something distribution specific to be done?

Thanks,
    Lars

lge:
And again I forgot the at-mentions: @oalbrigt @nrwahl2 @krig @dmuhamedagic @kgaillot
not sure if you got it anyways via the issue.

heartbeat/nfsserver
```diff
@@ -269,7 +301,7 @@ nfs_exec()
 	set_exec_mode
 
 	case $EXEC_MODE in 
-		1) ${OCF_RESKEY_nfs_init_script} $cmd;;
+		1) unshare -u /bin/sh -c "hostname ${OCF_RESKEY_nfs_server_scope}; exec ${OCF_RESKEY_nfs_init_script} $cmd" ;;
```
oalbrigt, line 304: You should check if nfs_server_scope is set before doing unshare.

bfields:
So, sorry, I'm a little ignorant of clustering.  Are you actually trying to run multiple NFS servers on one machine?  Or is that you're only running one NFS server, but you want it to get a different hostname without having to change the actual hostname in the init namespace?

oalbrigt:
> So, sorry, I'm a little ignorant of clustering. Are you actually trying to run multiple NFS servers on one machine? Or is that you're only running one NFS server, but you want it to get a different hostname without having to change the actual hostname in the init namespace?

It's to use a different hostname (e.g. Kerberos hostname), when used with a floating IP (link to old description in RGManager agent):
https://github.com/ClusterLabs/resource-agents/blob/master/rgmanager/src/resources/nfsserver.sh.in#L120-L129

heartbeat/nfsserver
```diff
@@ -5,6 +5,13 @@
 # by hxinwei@gmail.com
 # License: GNU General Public License v2 (GPLv2) and later
 
+
+# I don't know for certain whether all services actuall _need_ this,
+# I know that at least nfs-server needs it.
+NFSD_RELATED_SYSTEMD_SERVICE_FOR_UNSHARE_UTS_NAMESPACE="nfs-idmapd.service nfs-mountd.service nfs-server.service nfsdcld.service rpc-gssd.service rpc-statd.service rpcbind.service"
```
scottmayhew, line 11: I can't think of a reason why you would need to change the uts namespace for any of those services besides nfs-server.service.  The rest all start userspace daemons, and none of them call uname() except for rpc.mountd (which does it to get the kernel version).

heartbeat/nfsserver
```diff
@@ -636,6 +736,13 @@ nfsserver_start ()
 	is_redhat_based && set_env_args
 	bind_tree
 	prepare_directory
+	case $EXEC_MODE in [23])
+		if [ -z "$OCF_RESKEY_nfs_server_scope" ] ; then
+			remove_unshare_uts_dropins
+		else
+			inject_unshare_uts_name_into_systemd_units
+		fi
```
dmuhamedagic, line 744: Need ;; at the end of a case?

heartbeat/nfsserver
```diff
@@ -854,6 +961,8 @@ nfsserver_stop ()
 		ocf_log info "NFS server stopped"
 	fi
 
+	case $EXEC_MODE in [23]) remove_unshare_uts_dropins; esac
```
dmuhamedagic, line 964: Ditto.

heartbeat/nfsserver
dmuhamedagic: Right. Perhaps use a function which would print the scope to stdout. The function could then perhaps try to use a different cluster-specific string as the scope; iirc, we discussed something along those lines.

heartbeat/nfsserver
lge: if you look at [the rgmanager agent](https://github.com/ClusterLabs/resource-agents/blob/8dc8b40b69d522400e72c883e60e90ddb6392680/rgmanager/src/resources/nfsserver.sh.in#L224), nfs, rpcgssd and rpcidmapd have the unshare. Sure, it could be unnecessary there as well, but to me that sounds plausible at least.

heartbeat/nfsserver
lge: right, I'll add the "original" in the unset branch.
I decided against "making up" a scope name, it's just too much implicit magic (that needs to be correct and maintained), compared to just setting it explicitly, and pointing it out in documentation.

heartbeat/nfsserver
lge: only if there was an other case ;-) but I'll add it for style.

heartbeat/nfsserver
lge: ditto ;-) not necessary, but I'll add it for style.

heartbeat/nfsserver
```diff
@@ -269,7 +301,11 @@ nfs_exec()
 	set_exec_mode
 
 	case $EXEC_MODE in 
-		1) ${OCF_RESKEY_nfs_init_script} $cmd;;
+	    	1) if [ -z "$OCF_RESKEY_nfs_server_scope" ] ; then
```
oalbrigt, line 304: This doesnt match the former indentation.

oalbrigt: Looks good now.

heartbeat/nfsserver
oalbrigt: Let's try to do it for nfs-server.service, and just have the above list as a comment, so we can add any of those if we see any issues.

It's best to try and minimize the amount of changes we do to the system.

oalbrigt:
LGTM. Thanks.

PR #1691, opened 2021-09-10
fix symlink vs target realpath comparison

bmfp:
Hi,
this change allows using a symlink in the target path by getting canonical path of target as well as wanted link


oalbrigt:
ok to test

PR #1698, opened 2021-10-01
RabbitMQ OCF RA M/S HA cluster agent migration

bogdando:
New home for rabbitmq-server-ha RA OCF m/s HA

Moving it from repo:
https://github.com/rabbitmq/rabbitmq-server
The original path:
scripts/rabbitmq-server-ha.ocf

Also preserve list of authors and changes history since its
very initial commit in the Fuel for OpenStack project,
now archieved: https://github.com/openstack-archive/fuel-library

To get the history use:
$ git log --follow heartbeat/rabbitmq-server-ha.ocf

Reasoning behind: the OCF RA script provides M/S HA
pacemaker resource for RabbitMQ cluster and better fits
this place.

Background
==========
It's been actively maintained for years.
And now it needs a new home due to requests of RabbitMQ team,
since it is no longer possible to run CI tests for changes
proposed against it by the old location.

TripleO upstream project and its layered RH OSP product have plans
to adopt this OCF RA for its use. That guarantees the future
maintanance and support for it.

How it works
==========
Documentation is kept by its original upstream location:
https://www.rabbitmq.com/pacemaker.html#auto-pacemaker

Future Plans
============
Once it is there, the package builds for RDO will catch up
changes for that OCF RA and make sure it's CI'ed, also in TripleO
and OSP.

Status Quo
==========
Until the adoption completes, I'm planning to test changes proposed
by this new location in my fork, with github actions, like [0].
The CI runs on pre-build images [1] and vagrant scripts [1] that
I maintain for the (more or less) recent Pacemaker and RabbitMQ
builds. The test coverage includes a simple cluster assemble smoke
test and a sofisticated Jepsen testcase that verifies auto-healing
of the cluster resource in Pacemaker managed by this OCF RA.

[0] https://github.com/bogdando/rabbitmq-server/runs/3757495446
[1] https://hub.docker.com/r/bogdando/rabbitmq-cluster-ocf
[2] https://github.com/bogdando/rabbitmq-cluster-ocf-vagrant

Signed-off-by: Bogdan Dobrelya <bdobreli@redhat.com>


bogdando:
Now it works as expected, see an [example commit from history](https://github.com/ClusterLabs/resource-agents/commit/3a9253aa5f95f23839b4b70a6e3ae1412745368e)

bogdando:
Please note that this PR includes commits all the way back to 2015 intentionally, to preserve authored changes and link authors, whenever possible, as well

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+                else
+                    ocf_log info "${LH} disconnecting node '${nodename}' failed."
+                fi
+
+                # unjoin node
+                # when the rabbit node went down, its status
+                # remains 'running' for a while, so few retries are required
+                local tries=0
+                until [ $tries -eq 5 ]; do
+                    tries=$((tries+1))
```
oalbrigt, line 1137: I think `$(( ))` is bash specific, so you should probably use expr instead.

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+    local ts_end
+    local pf_end
+    local pid
+    local LH="${LL} start_beam_process():"
+
+    # remove old PID-file if it exists
+    if [ -f "${OCF_RESKEY_pid_file}" ] ; then
+        ocf_log warn "${LH} found old PID-file '${OCF_RESKEY_pid_file}'."
+        pid=$(cat ${OCF_RESKEY_pid_file})
+        if [ "${pid}" -a -d "/proc/${pid}" ] ; then
```
oalbrigt, line 1243: We prefer not using -a/-o to avoid issues like this:
https://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+        new_score=$MIN_MASTER_SCORE
+    else
+        new_score=$BEST_MASTER_SCORE
+        for node in $(get_alive_pacemaker_nodes_but $THIS_PCMK_NODE)
+        do
+            node_start_time=$(get_node_start_time $node)
+            node_score=$(get_node_master_score $node)
+
+            ocf_log info "${LH} comparing us (start time: $our_start_time, score: $new_score) with $node (start time: $node_start_time, score: $node_score)"
+            if [ $node_start_time -ne 0 -a $node_score -ne 0 -a $node_start_time -lt $our_start_time ]; then
```
oalbrigt, line 1640: We should use && here as well.

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+                new_score=$((node_score - 10 < new_score ? node_score - 10 : new_score ))
+            elif [ $node_start_time -ne 0 -a $node_score -ne 0 -a $node_start_time -eq $our_start_time ]; then
```
oalbrigt, line 1642: We should use && here as well.

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+        return $OCF_SUCCESS
+    fi
+}
+
+node_health_check_legacy() {
+    local rc_alive
+    local timeout_alive
+    su_rabbit_cmd "${OCF_RESKEY_ctl} list_channels > /dev/null 2>&1"
+    rc_alive=$?
+    [ $rc_alive -eq 137 -o $rc_alive -eq 124 ] && ocf_log err "${LH} 'rabbitmqctl list_channels' timed out, per-node explanation: $(enhanced_list_channels)"
```
oalbrigt, line 1789: We should use || here.

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+        return $OCF_ERR_GENERIC
+
+    elif [ $timeout_alarms -eq 0 ]; then
+        if [ $rc_alarms -ne 0 ]; then
+            ocf_log err "${LH} rabbitmqctl get_alarms exited with errors."
+            rc=$OCF_ERR_GENERIC
+
+        elif [ -n "${alarms}" ]; then
+            for node in ${alarms}; do
+                name=`echo ${node} | perl -n -e "m/memory,'(?<n>\S+)+'/ && print \"$+{n}\n\""`
```
oalbrigt, line 1827: We should use sed or awk here.

oalbrigt (changes requested):
Looks good in general.

I've added some comments of what I think should be improved.

You should also check what parts of the logging to INFO should stay INFO, and which can be changed to DEBUG (to avoid spamming the logs too much).

Consider switching license to the same as the other agents, unless you got a good reason for using the current license:
https://github.com/ClusterLabs/resource-agents/blob/master/doc/dev-guides/ra-dev-guide.asc#licensing

bogdando:
@jeckersb PTAL

bogdando:
@oalbrigt thank you for reviewing! Do you think it would make sence to land this first, then addressing your changes by small patches, which I could test on my own, as individual changes?

heartbeat/rabbitmq-server-ha.ocf
```diff
@@ -0,0 +1,2435 @@
...
+meta_data() {
+    # The EXTENDED_OCF_PARAMS parameter below does not exist by default
+    # and hence converted to an empty string unless overridden. It
+    # could be used by an extention script to add new parameters. For
+    # example see https://review.openstack.org/#/c/249180/10
+
+    cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="${OCF_RESKEY_binary}">
```
oalbrigt, line 125: This should be the name of the agent (same as it's filename),

heartbeat/rabbitmq-server-ha.ocf
bogdando: Good point, I'd fix that in follow-up. Since that might regress the agent code and requires extra run of functional testing, I'd ran it in my forked repo with github actions once I point that to the "new home".

oalbrigt (approved):

bogdando:
It's good to go. I tested it with [github workflows set up for my fork](https://github.com/bogdando/rabbitmq-server/runs/4105799406?check_suite_focus=true), it passed smoke and jepsen without regressions (in my test suite jepsen always reports failures because messages are expectedly lost, so that's the expected result)

PR #1699, opened 2021-10-06
ra-dev-guide: update agent/OCF version info

oalbrigt:
Based on: https://github.com/ClusterLabs/OCF-spec/blob/master/ra/1.0/ra-metadata-example.xml#L3-L7


doc/dev-guides/ra-dev-guide.asc
```diff
@@ -180,7 +180,9 @@ OCF Resource Agent Developers Guide.
 --------------------------------------------------------------------------
 
 The +resource-agent+ element, of which there must only be one per
-resource agent, defines the resource agent +name+ and +version+.
+resource agent, defines the resource agent +name+ and +version+. The
++version+ element specifies the OCF version standard the metadta complies
```
kgaillot, line 184: metadata

PR #1700, opened 2021-10-07
all agents/ocf.py: follow OCF standard for version and OCF version in metadata

oalbrigt:


tomjelinek:
Looks good to me, thanks!

PR #1702, opened 2021-10-08
Added the ability to unset utilization values automatically

tparkercbn:
This is useful in the case they were added automatically and need to be removed.  

Sample useage:
params \
    autoset_utilization_cpu=false \
    unset_utilization_cpu=true \
    autoset_utilization_hv_memory=false \
    unset_utilization_hv_memory=true

If autoset is set to true that will be higher priortiy than unset so you can't actually do both.


heartbeat/VirtualDomain
```diff
@@ -389,20 +422,36 @@ set_util_attr() {
...
+
+	outp=$(crm_resource --resource=$OCF_RESOURCE_INSTANCE --utilization --delete-parameter=$attr 2>&1) ||
+	ocf_log warn "crm_resource failed to unset utilization attribute $attr: $outp"
+}
+
 update_utilization() {
 	local dom_cpu dom_mem
 
 	if ocf_is_true "$OCF_RESKEY_autoset_utilization_cpu"; then
 	   dom_cpu=$(LANG=C virsh $VIRSH_OPTIONS dominfo ${DOMAIN_NAME} 2>/dev/null | awk '/CPU\(s\)/{print $2}')
```
oalbrigt, line 437: Can you update these sections? They seem to combine tabs with spaces.

tparkercbn: You are correct.  I missed that when I replaced the other spaces.  New commit coming.

Issue #1706, opened 2021-11-04
heartbeat/pgsql - PGSQL.lock is never deleted, because OCF_RESKEY_CRM_meta_notify_slave_uname is never empty


dbk-rabel:
Hi.

As far as I understood, PGSQL.lock should be deleted automatically on demoting or stopping the pgsql resource on the master. But it does not.
This is the case, since in https://github.com/ClusterLabs/resource-agents/blob/master/heartbeat/pgsql#L871 it is tested if OCF_RESKEY_CRM_meta_notify_slave_uname ist empty, which should not be the case.

Example: Two node cluster. crm resource stop ...

This seems to be unwanted behaviour. If I am incorrect, could someone explain to me?

Thanks a lot!

Yours
  David

dbk-rabel:
I close this issue. See PR #1707  for more information.

Issue #808, opened 2016-05-13
Postgres Master-role migration 


kunalija:
Hi guys,

I have a postgres master/slave setup on SLES11SP3. For maintenance purpose I need a procedure to move the master role to the other node. I've tried the command 'crm_resource --resource msPostgresql --ban --master --host node1 --lifetime "P60S"'. Master role is actually moved to node2 but node1 stops with status disconnected (PGSQL.lock file remains in tmp directory). I have to manually delete the lock file and do a resource cleanup. Node1 should remain as slave after the master migration. Is it a bug or expected behaviour?
PS What is the recommended version of pgsql ra agent, the one shipped with SLES or the latest from git? 

Version: 1.1.11-3ca8c3b
Postgres 9.1

Thanks

krig:
Well, first of all SLES11SP3 is getting quite old. Besides that, it sounds like this may be a problem of too short timeouts: Try increasing your operation timeouts and see if that helps.

kunalija:
These are my current values. Tried timeout with 120s and different "interval" values, same result.

 primitive pgsql ocf:heartbeat:pgsql \
     params \
        pgctl="/usr/bin/pg_ctl" \
        psql="/usr/bin/psql" \
        pgdata="/var/lib/pgsql/data/" \
        start_opt="-p 5432" \
        rep_mode="sync" \
        node_list="db1 db2" \
        restore_command="/var/lib/pgsql/data/restore_wal %f %p" \
        primary_conninfo_opt="keepalives_idle=60 keepalives_interval=5 keepalives_count=5" \
        master_ip="192.168.5.100" \
        restart_on_promote="true" \
		op start   timeout="60s" interval="0s"  on-fail="restart" \
		op monitor timeout="60s" interval="4s" on-fail="restart" \
		op monitor timeout="60s" interval="3s"  on-fail="restart" role="Master" \
		op promote timeout="60s" interval="0s"  on-fail="restart" \
		op demote  timeout="60s" interval="0s"  on-fail="stop" \
		op stop    timeout="60s" interval="0s"  on-fail="block" \
		op notify  timeout="60s" interval="0s"

 primitive vip-master ocf:heartbeat:IPaddr2 \
     params \
         ip="192.168.5.100" \
         nic="eth0" \
         cidr_netmask="24" \
     op start   timeout="60s" interval="0s"  on-fail="stop" \
     op monitor timeout="60s" interval="10s" on-fail="restart" \
[...]

PR #1707, opened 2021-11-04
Update heartbeat/pgsql

dbk-rabel:
#1706

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -868,7 +868,7 @@ pgsql_real_stop() {
     # Remove postmaster.pid if it exists
     rm -f $PIDFILE
 
-    if  [ "$1" = "master" -a "$OCF_RESKEY_CRM_meta_notify_slave_uname" = " " ]; then
+    if  [ "$1" = "master" ]; then
         ocf_log info "Removing $PGSQL_LOCK."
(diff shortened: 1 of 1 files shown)


HideoYamauchi:
Hi dbk-rabel,

Perhaps it has been discussed in the issue or PR before, but the LOCK file should only be deleted when a single Master stops.
So I think the current code is correct.

Best Regards,
Hideo Yamauchi.

dbk-rabel:
Hi HideoYamauchi,

thanks a lot for your answer!

But I don't seem to understand that. When the master shuts down and there are still Slaves to take over, it should be safe to remove the LOCK file, so the master can start again later. Even more, if the master is not even stopped, but just demoted to slave, while another slave takes over the master role.

It seems, I got something wrong here, but I'm not able to figure it out. It would be very helpful, if you could explain to me in more detail.

The background for me is actually: How to safely reboot the master node, without any manual interaction?

Yours
  David

HideoYamauchi:
Hi dbk-rabel,

The discussion of the next issue may be helpful to you.
 - https://github.com/ClusterLabs/resource-agents/issues/808

Best Regards,
Hideo Yamauchi.

dbk-rabel:
Yes, that helped. Thank you!

PR #1715, opened 2021-11-18
asterisk: fix sipsak check during start-action

x0r3n:
To avoid the case where the sipsak check runs before the sip starts at the start action

### DIFF
--- a/heartbeat/asterisk
+++ b/heartbeat/asterisk
@@ -307,22 +307,23 @@ asterisk_monitor() {
     #   This can also happen if sipsak is run too early after asterisk
     #   start.
    
-   #To avoid the case where the sipsak check runs before the sip starts at the start action
+    #To avoid the case where the sipsak check runs before the sip starts at the start action
+    SIPCHECK=`sipsak -s $OCF_RESKEY_monitor_sipuri`
     if [ -n "$OCF_RESKEY_monitor_sipuri" ]; then   
-        ocf_run sipsak -s "$OCF_RESKEY_monitor_sipuri"
+        ocf_run $SIPCHECK
         rc=$?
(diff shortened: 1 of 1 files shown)


heartbeat/asterisk
```diff
@@ -306,13 +306,24 @@ asterisk_monitor() {
     #   simply no answer (timeout).
     #   This can also happen if sipsak is run too early after asterisk
     #   start.
-    if [ -n "$OCF_RESKEY_monitor_sipuri" ]; then
+   
+   #To avoid the case where the sipsak check runs before the sip starts at the start action
+    if [ -n "$OCF_RESKEY_monitor_sipuri" ]; then   
         ocf_run sipsak -s "$OCF_RESKEY_monitor_sipuri"
```
oalbrigt, line 312: Can you define the command in `cmd=` or similar name and run that with ocf_run in both instances?

That should make it easier to see that they're the same and update in the future.

heartbeat/asterisk
```diff
@@ -306,13 +306,24 @@ asterisk_monitor() {
...
+            sleep 1
+            ocf_run sipsak -s "$OCF_RESKEY_monitor_sipuri"
+            rc=$?
+          done
+        else
         case "$rc" in
           1|2) return $OCF_ERR_GENERIC;;
           3)   return $OCF_NOT_RUNNING;;
         esac
+        fi
```
oalbrigt, line 326: You should indent the case block one more time to make it easier to see that it's part of the else.

heartbeat/asterisk
x0r3n: sure

oalbrigt:
ok to test

oalbrigt (approved):
LGTM.

PR #1721, opened 2021-12-01
Filesystem: unmount bind mounts before unmount file system

ganghe:
When unmount a file system, we should unmount any bind mounts on
this device firstly. Otherwise, the device still has file system
mounts, this may cause the deactivation of the device to fail.

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -359,6 +359,27 @@ list_submounts() {
 	list_mounts | grep "${TAB}${1}/" | cut -d"$TAB" -f2 | sort -r
 }
 
+# Lists all bind mounts of a given file system,
+# excluding the path itself.
+list_bindmounts() {
+	if is_bind_mount; then
+		# skip bind mount
+		# we should not umount the original file system via a bind mount
+		return
+	fi
+
+	match_string="${TAB}${1}${TAB}"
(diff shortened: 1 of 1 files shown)


ganghe:
Hello Guys,

I'd like to submit a patch, which can help to umount bind mounts before umount this file system.
The requirement is from our customer side, please refer to
https://lists.clusterlabs.org/pipermail/developers/2021-April/002362.html

Thanks
Gang

oalbrigt:
ok to test

ganghe:
> ok to test

Hello Oyvind,

Could you help to take a code review? this patch was asked from the customer side, in fact, they is using Filesystem RA like this way, they hope this enhancement can be merged into the upstream. 

Thanks
Gang

oalbrigt:
LGTM.

heartbeat/Filesystem
```diff
@@ -359,6 +359,27 @@ list_submounts() {
 	list_mounts | grep "${TAB}${1}/" | cut -d"$TAB" -f2 | sort -r
 }
 
+# Lists all bind mounts of a given file system,
+# excluding the path itself.
+list_bindmounts() {
+	if is_bind_mount; then
+		# skip bind mount
+		# we should not umount the original file system via a bind mount
```
zzhou1, line 367: Good catch! This will protect the existing bind mount use cases!

PR #1724, opened 2021-12-07
Route: return OCF_NOT_RUNNING for probe action when interface or route doesnt exist

oalbrigt:


robert-scheck:
Without this patch, the Route resource agent fails, if the non-default IPv4 FIB table is not yet initialized, even it would be the resource's job to add a route to the non-default IPv4 FIB (which initialized it at the same time, but still the check whether the route exists fails). And with this patch, it works properly.

Example (after an IPv4 and an IPv6 address has been added to ens10):

```
$ ip -4 route show table 1
Error: ipv4: FIB table does not exist.
Dump terminated
$ 
```

The failure above leads to `OCF_ERR_GENERIC` and this is exactly where this patch jumps in.

IPv6 works, because the default IPv6 link-local address initializes the IPv6 FIB table already.

```
$ ip -6 route show table 1
$ 
```

PR #1733, opened 2022-01-10
Azure Events: handle events locally and use location constraint to su…

MSSedusch:
…pport clusters using Azure Availability Zones


varkoly:
Why is the 2019-08-01 version still being used instead of the current 2020-07-01?

MSSedusch:
will check internally if this PR is still needed. If this is the case, I would duplicate the resource agent to avoid breaking customers who are already using the Events resource agent

PR #1734, opened 2022-01-17
openstack-*: add support for multiple setup options

oalbrigt:
- clouds.yaml
- parameters (based on fence_openstack naming)


fabbione:
ack from me, tho I will only be able to properly test them once they are merged.

PR #1740, opened 2022-02-02
Filesystem: Add support for Amazon EFS mount helper

nrwahl2:
`mount.efs`, the mount helper for Amazon Elastic File System (EFS)
provided by `amazon-efs-utils` [1], is a wrapper for `mount.nfs4`. It offers
a number of AWS-specific mount options and some security improvements
like encryption of data in transit.

Here we add support by treating an `fstype=efs` like `fstype=nfs4` for the
most part.

This PR also does the following:
- fixes a check that's been causing us to skip `fstype_supported()` since it was added 8 years ago.
- informs the user when we have to use a different name in order to check fstype support. I'm logging to info instead of debug because we only call `fstype_supported()` during start operations and this will make the exit reason less confusing if we fail.

[1] https://docs.aws.amazon.com/efs/latest/ug/efs-mount-helper.html

Resolves: RHBZ#2049319

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -341,7 +341,7 @@ determine_blockdevice() {
 	# Get the current real device name, if possible.
 	# (specified devname could be -L or -U...)
 	case "$FSTYPE" in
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM.

PR #1743, opened 2022-02-07
rabbitmq-server-ha: Use ocf_is_true, if applicable

bogdando:
ocf_is_true is better, as users may use other true/false values as well

Signed-off-by: Bogdan Dobrelya <bdobreli@redhat.com>


oalbrigt:
ok to test

PR #1745, opened 2022-02-17
IPsrcaddr: Better error message when no matching route found

nrwahl2:
If `OCF_RESKEY_destination` is not explicitly set and `ip route list`
can't find a route matching the specifications, the `NETWORK` variable
doesn't get set. This causes a certain failure of the start operation,
because there is no `PREFIX` argument to `ip route replace` (syntax
error). It may also cause unexpected behavior for stop operations (but
not in all cases). During a monitor, this event can only happen if
something has changed outside the cluster's control, and so is cause
for warning there.

Exit `OCF_ERR_ARGS` for start, log debug for probe, log warning for all
other ops.

Resolves: RHBZ#1654862


oalbrigt:
LGTM. Thanks.

PR #1748, opened 2022-03-02
corosync-qnetd: Add resource agent

jfriesse:
Mostly for better monitor operation using corosync-qnetd-tool. As qnetd
is (almost) stateless only directory which has to be copied (once)
across the nodes is nss db directory (usually
/etc/corosync/qnetd/nssdb).

Signed-off-by: Jan Friesse <jfriesse@redhat.com>

### DIFF
--- a/heartbeat/corosync-qnetd
+++ b/heartbeat/corosync-qnetd
@@ -1,0 +1,353 @@
+#!/bin/sh
+#
+# Copyright (C) 2022 Red Hat, Inc.  All rights reserved.
+#
+# Authors: Jan Friesse <jfriesse@redhat.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of version 2 of the GNU General Public License as
+# published by the Free Software Foundation.
+#
+# This program is distributed in the hope that it would be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# Further, this software is distributed without any warranty that it is
+# free of the rightful claim of any third person regarding infringement
+# or the like.  Any license provided herein, whether implied or
+# otherwise, applies only to this software file.  Patent licenses, if
+# any, provided herein do not apply to combinations of this program with
+# other software, or any other product whatsoever.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+
+# Initialization:
+: "${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}"
+. "${OCF_FUNCTIONS_DIR}/ocf-shellfuncs"
+
+# Use runuser if available for SELinux.
+if [ -x "/sbin/runuser" ]; then
+    SU="runuser"
+else
+    SU="su"
+fi
+
+# Attempt to detect a default binary
+OCF_RESKEY_binary_default=$(which corosync-qnetd 2> /dev/null)
+if [ "${OCF_RESKEY_binary_default}" = "" ]; then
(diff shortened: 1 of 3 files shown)


heartbeat/corosync-qnetd
```diff
@@ -0,0 +1,347 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write the Free Software Foundation,
+# Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
+#
+
+# Initialization:
+: "${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}"
+. "${OCF_FUNCTIONS_DIR}/ocf-shellfuncs"
+
+# Use runuser if available for SELinux.
```
oalbrigt, line 31: You dont use this, so you can remove it.

heartbeat/corosync-qnetd
```diff
@@ -0,0 +1,347 @@
...
+    </longdesc>
+    <shortdesc lang="en">corosync-qnetd pid file</shortdesc>
+    <content type="string" default="${OCF_RESKEY_pid_default}" />
+</parameter>
+
+<parameter name="port">
+    <longdesc lang="en">
+        TCP port to listen on. Default port is 5403.
+    </longdesc>
+    <shortdesc lang="en">TCP portto listen on</shortdesc>
```
oalbrigt, line 128: Typo.

heartbeat/corosync-qnetd
```diff
@@ -0,0 +1,347 @@
...
+    fi
+
+    pid_dir=$(dirname "${OCF_RESKEY_pid}")
+    sock_dir=$(dirname "${OCF_RESKEY_ipc_sock}")
+
+    for d in "$pid_dir" "$sock_dir";do
+        if [ ! -d "$d" ];then
+            mkdir -p "$d"
+            chmod 0770 "$d"
+            chown "${OCF_RESKEY_user}:${OCF_RESKEY_user}" "$d"
```
oalbrigt, line 206: We should probably add a group parameter as well for cases where user != group.

heartbeat/corosync-qnetd
```diff
@@ -0,0 +1,347 @@
...
+        fi
+    done
+
+    if [ -n "${OCF_RESKEY_nss_db_dir}" ];then
```
oalbrigt, line 210: It's probably better to set `params=""` and do `params="$params ..."` in the if's here.

heartbeat/corosync-qnetd
```diff
@@ -0,0 +1,347 @@
...
+        $param_ip \
+        $param_port \
+        ${OCF_RESKEY_qnetd_opts}"
+
+    while :; do
+        corosync_qnetd_monitor "debug"
+        rc="$?"
+
+        if [ "$rc" -eq "${OCF_SUCCESS}" ]; then
+            break ;
```
oalbrigt, line 235: You can remove the ;

oalbrigt (changes requested):

heartbeat/corosync-qnetd
jfriesse: great catch, fixed

heartbeat/corosync-qnetd
jfriesse: upstram init script has only user and I "copy/paste" that part... I've added group param with default = user_default

heartbeat/corosync-qnetd
jfriesse: I've moved everything to params="$params ...." (even the "hardcoded" sock/pid) so resulting ocf_run stanza is much shorter.

heartbeat/corosync-qnetd
jfriesse: done

jfriesse:
Thank you for the review. I've force pushed the updated version (hopefully everything fixed) with support for instance(s) and `$SU` (instead of old `${SU}`)

oalbrigt (approved):

PR #1749, opened 2022-03-02
ocf-shellfuncs: quote pid in ocf_pidfile_status

jfriesse:
Usage of pidfile variable was not quoted so it was not possible to call
ocf_pidfile_status with path containing space. Solution is to add quotes
to -e test and cat command. Also use $() syntax for cat command and
unify indentation.

Signed-off-by: Jan Friesse <jfriesse@redhat.com>


oalbrigt:
retest this please

oalbrigt:
LGTM. Thanks.

PR #1753, opened 2022-03-10
LVM-activate:change lvm_status return value from ocf_not_running to ocf_err_generic

XingWei-Liu:
hi  ，I found a problem when building a cluster  , When the cluster uses LVM-activate and the on-fail=fence parameter is set, if  the vg is lost, the return value of the monitor function is ocf_not_running  that the cluster cannot trigger the fence action

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -790,7 +790,11 @@ lvm_status() {
 	fi
 
 	if [ $dm_count -eq 0 ]; then
-		return $OCF_ERR_GENERIC
+		if ocf_is_probe ;then
+			return $OCF_NOT_RUNNING
+		else
+			return $OCF_ERR_GENERIC
+		fi
 	fi
 
 	case "$OCF_CHECK_LEVEL" in
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -791,9 +791,9 @@ lvm_status() {
 
 	if [ $dm_count -eq 0 ]; then
 		if ocf_is_probe ;then
-			return $OCF_NOT_RUNNING
-		else
 			return $OCF_ERR_GENERIC
+		else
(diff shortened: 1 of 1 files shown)


wenningerk:
test this please

heartbeat/LVM-activate
```diff
@@ -790,7 +790,7 @@ lvm_status() {
 	fi
 
 	if [ $dm_count -eq 0 ]; then
-		return $OCF_NOT_RUNNING
```
oalbrigt, line 793: We should probably do an `if ocf_is_probe; then` here to return NOT_RUNNING for probe-actions.

XingWei-Liu: ok，It's fixed, can you please check it?

XingWei-Liu: It's fixed, can you please check it?

oalbrigt: It should be reversed. ERR_GENERIC in the else-clause.

XingWei-Liu: I have some questions  ,doesn't  "ocf_is_probe " mean cluster is using monitor ?

XingWei-Liu: in commits  , i just reversed it before ⊙︿⊙

oalbrigt: > I have some questions ,doesn't "ocf_is_probe " mean cluster is using monitor ?

Correct. In this agent the monitor-action is in lvm_status(). Probe is a special monitor where it only checks whether it's running or not (e.g. for disabled resources).

XingWei-Liu: ok got it，It's fixed

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1755, opened 2022-03-23
ipsec: fix check string against another string not the before defined variable

jfrede:


heartbeat/ipsec
```diff
@@ -150,10 +150,14 @@ ipsec_monitor() {
 
 	ipsec status | grep "$OCF_RESKEY_tunnel" | grep -q unoriented
 	state=$?
+
+	# TODO  I think we need also to check if output of ipsec status is empty!
+	#       Then the tunnel is not running.
```
oalbrigt, line 155: If it is empty it wont match unoriented.

jfrede: Yes but if it is empty the check reports that the tunnel is active which is not true.

oalbrigt: Oh.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1762, opened 2022-04-06
ethmonitor,ovsmonitor,mariadb: check for bc binary

vvidic:
bc might not be installed by default on a minimal Debian system.


oalbrigt:
Thanks.

PR #1764, opened 2022-04-20
podman: add podman parameter error judgment

XingWei-Liu:
hi, I found a problem when using podman. If the parameters are incorrectly filled when creating a podman resource, podman will not return the corresponding error message as exit_reason, so some modifications have been made in podman to add podman parameter error judgment.

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -376,6 +376,11 @@ run_new_container()
 		# with the same name, even though podman itself thinks there
 		# is no such container. If so, purge the storage layer to try
 		# to clean the corruption and try again.
+                if [ -n $(echo "$out" | grep -q "unknown*flag") ]; then
+                        ocf_exit_reason "$out"
+                        return $rc
+                fi
+
 		ocf_log warn "Internal podman error while creating new container $CONTAINER. Retrying."
 		ocf_run podman rm --storage $CONTAINER
 		ocf_run podman run $opts $image $cmd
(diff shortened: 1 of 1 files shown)


heartbeat/podman
```diff
@@ -376,6 +376,11 @@ run_new_container()
 		# with the same name, even though podman itself thinks there
 		# is no such container. If so, purge the storage layer to try
 		# to clean the corruption and try again.
+                if [ -n $(echo "$out" | grep -q "unknown*flag") ]; then
```
bogdando, line 379: you could do as well w/o a subshell:
```
if echo "$out" | grep -q "unknown*flag"; then

Comment (XingWei-Liu) [reply heartbeat/podman]: i try this before   , but its not work  on my pc ,its there some problem in my system?

Comment (oalbrigt) [reply heartbeat/podman:379]: Seems like you need to use `.*` instead of `*`.

Comment (XingWei-Liu) [reply heartbeat/podman]: my fault  ，its work，thanks ,i commit again can you hellp to review it

File: heartbeat/podman [hunk 2]
```diff
@@ -376,6 +376,11 @@ run_new_container()
 		# with the same name, even though podman itself thinks there
 		# is no such container. If so, purge the storage layer to try
 		# to clean the corruption and try again.
+                if echo "$out" | grep -q "unknown.*flag"; then
+                        ocf_exit_reason "$out"
+                        return $rc
+                fi
```
oalbrigt, line 382: Can you update the code to use tabs instead of spaces for indentation, so it follows the rest of the agent?

XingWei-Liu: ok , wait a minute, let me replace it

heartbeat/podman
XingWei-Liu: its done

oalbrigt:
ok to test

XingWei-Liu:
test failed ，why, where i can  see the error log

oalbrigt:
It's not related to the patch, so I'm in contact with the admin to hear if I should rerun it, or just ignore it for now.

oalbrigt:
Thanks

XingWei-Liu:
its my pleasure

PR #1771, opened 2022-05-25
aws-vpc-move-ip: Allow to set the interface label

petrpavlu:
Add a parameter to specify an interface label to distinguish the IP address managed by aws-vpc-move-ip, similarly as can be done with IPaddr2. This allows to easily recognize the address from other addresses assigned to a given interface.

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -43,6 +43,7 @@ OCF_RESKEY_address_default=""
 OCF_RESKEY_routing_table_default=""
 OCF_RESKEY_routing_table_role_default=""
 OCF_RESKEY_interface_default="eth0"
+OCF_RESKEY_iflabel_default=""
 OCF_RESKEY_monapi_default="false"
 OCF_RESKEY_lookup_type_default="InstanceId"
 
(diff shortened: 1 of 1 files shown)


wenningerk:
test this please

wenningerk:
sry seems as if I got purged from the list :-(

heartbeat/aws-vpc-move-ip
```diff
@@ -363,7 +385,13 @@ ec2ip_get_and_configure() {
...
 	# Reconfigure the local ip address
 	ec2ip_drop
-	cmd="ip addr add ${OCF_RESKEY_ip}/32 dev $OCF_RESKEY_interface"
+
+	extra_opts=""
+	if [ -n "$OCF_RESKEY_iflabel" ]; then
+		extra_opts="$extra_opts label $OCF_RESKEY_interface:$OCF_RESKEY_iflabel"
+	fi
+
+	cmd="ip addr add ${OCF_RESKEY_ip}/32 dev $OCF_RESKEY_interface$extra_opts"
```
oalbrigt, line 394: This will expand to `$OCF_RESKEY_interface$OCF_RESKEY_interface$extra_opts`, and you probably want to use it in the other actions as well.

So I think a better solution is to set `interface_opt` like we do with region, and use that instead of `$OCF_RESKEY_interface` in the rest of the code:
https://github.com/ClusterLabs/resource-agents/blob/273c014a839a85ae6cd0e84f42ddced4339f37f0/heartbeat/aws-vpc-move-ip#L62=

petrpavlu: With `iflabel` set, the line should expand into `... dev $OCF_RESKEY_interface label $OCF_RESKEY_interface:$OCF_RESKEY_iflabel`, e.g. with `interface=eth0` and `iflabel=myvip`: `... dev eth0 label eth0:myvip`.

It should be sufficient to use the label only for the `ip address add` command. My idea was to match IPaddr2 which does the same.

petrpavlu: Updated patch adds the mentioned space.

oalbrigt:
Thanks.

PR #1776, opened 2022-06-01
CTDB: disable realtime scheduling to avoid "Unable to set scheduler to SCHED_FIFO (Operation not permitted)" error

oalbrigt:


oalbrigt:
We're considering other options to solve this issue.

oalbrigt:
Closing in favor of: https://github.com/ClusterLabs/resource-agents/pull/1796

PR #1783, opened 2022-06-23
LVM-activate: Fix return codes

nrwahl2:
Splitting this commit out from #1600.

OCF_ERR_ARGS should be used when the configuration isn't valid for the
**local** node, and so the resource should not attempt to start again
locally until the issue is corrected.

OCF_ERR_CONFIGURED should be used when the configuration isn't valid on
**any** node, and so the resource should not attempt to start again
anywhere until the issue is corrected.

One remaining gray area:
  - Should lvmlockd/clvmd improperly running (or improperly not running)
be an OCF_ERR_GENERIC or OCF_ERR_ARGS? The fact that it's a state issue
rather than a config issue suggests OCF_ERR_GENERIC. The fact that it
won't be fixed without user intervention suggests OCF_ERR_ARGS.

More info about return code meanings:
  - https://clusterlabs.org/pacemaker/doc/en-US/Pacemaker/2.0/html-single/Pacemaker_Administration/#_how_are_ocf_return_codes_interpreted
  - https://github.com/ClusterLabs/OCF-spec/pull/21/commits/60718974 (WIP)

Resolves: RHBZ#1905820

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -333,8 +333,7 @@ config_verify()
 	real=$(lvmconfig "$name" | cut -d'=' -f2)
 	if [ "$real" != "$expect" ]; then
 		ocf_exit_reason "config item $name: expect=$expect but real=$real"
-		exit $OCF_ERR_CONFIGURED
-
+		exit $OCF_ERR_ARGS
 	fi
 
 	return $OCF_SUCCESS
@@ -366,12 +365,12 @@ lvmlockd_check()
 		fi
 
 		ocf_exit_reason "lvmlockd daemon is not running!"
-		exit $OCF_ERR_CONFIGURED
+		exit $OCF_ERR_GENERIC
 	fi
 
 	if pgrep clvmd >/dev/null 2>&1 ; then
 		ocf_exit_reason "clvmd daemon is running unexpectedly."
-		exit $OCF_ERR_CONFIGURED
+		exit $OCF_ERR_GENERIC
 	fi
 
 	return $OCF_SUCCESS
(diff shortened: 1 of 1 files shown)


oalbrigt:
> One remaining gray area:
> 
>     * Should lvmlockd/clvmd improperly running (or improperly not running)
>       be an OCF_ERR_GENERIC or OCF_ERR_ARGS? The fact that it's a state issue
>       rather than a config issue suggests OCF_ERR_GENERIC. The fact that it
>       won't be fixed without user intervention suggests OCF_ERR_ARGS.

I would say _GENERIC. _ARGS would make sense if it can only happen due to incorrect configuration of the resource or lvm.conf. Also _ARGS would make it not try to restart, so _GENERIC let's it recover if it's just a user error or it died.

nrwahl2:
> Also _ARGS would make it not try to restart, so _GENERIC let's it recover if it's just a user error or it died.

The `clvmd_check` and `lvmlockd_check` functions are invoked only during a start operation (through `validate-all`). So as long as `start-failure-is-fatal=true`, the behavior would be identical for `_ARGS` and `_GENERIC`.

In the case that `start-failure-is-fatal=false`
  - I can see "the process died (and may recover)" as a reason to retry startup, so I agree with using `_GENERIC` for "improperly not running.
  - I can't see "the process is improperly running (and needs to be manually stopped)" as a reason to retry startup, so it might make sense to use `_ARGS` here.

I'm on the fence about the "improperly running" case. But there is no practical significance, only semantic significance. I'm pushing an update that uses `_GENERIC` for both.

oalbrigt:
Thanks.

PR #1789, opened 2022-07-14
openstack-agents: set domain parameter's default to Default and fix missing parameter name in ocf_exit_reason

oalbrigt:


fabbione (approved):
ack

PR #1791, opened 2022-07-19
Make storage_mon -h exit just after printing help messages.

MasaoFujii:
Previously, when -h or an invalid option was specified, storage_mon
printed the help messages, proceeded processing and then could
throw an error. This was not the behavior that, e.g., users who want
to specify -h option to see the help messages are expecting. To fix
this issue, this commit changes storage_mon so that it exits just
after printing the help messages when -h or an invalid option is
specified.

### DIFF
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -28,7 +28,7 @@ static void usage(char *name, FILE *f)
 	fprintf(f, "      --timeout <n>   max time to wait for a device test to come back. in seconds (default %d)\n", DEFAULT_TIMEOUT);
 	fprintf(f, "      --inject-errors-percent <n> Generate EIO errors <n>%% of the time (for testing only)\n");
 	fprintf(f, "      --verbose        emit extra output to stdout\n");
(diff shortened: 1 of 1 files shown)


tools/storage_mon.c
```diff
@@ -28,7 +28,7 @@ static void usage(char *name, FILE *f)
 	fprintf(f, "      --timeout <n>   max time to wait for a device test to come back. in seconds (default %d)\n", DEFAULT_TIMEOUT);
 	fprintf(f, "      --inject-errors-percent <n> Generate EIO errors <n>%% of the time (for testing only)\n");
 	fprintf(f, "      --verbose        emit extra output to stdout\n");
-	fprintf(f, "      --help           print this messages\n");
+	fprintf(f, "      --help           print this messages, then exit\n");
```
oalbrigt, line 31: No need to update this, but you can change it to "print this message", as messages is a typo.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1792, opened 2022-07-19
storage_mon: Fix bug in checking of number of specified scores.

MasaoFujii:
Previously specifying the maximum allowed number (MAX_DEVICES, currently 25)
of devices and scores as arguments could cause storage_mon to fail unexpectedly
with the error message "too many scores, max is 25". This issue happened
because storage_mon checked whether the number of specified scores
exceeded the upper limit by using the local variable "device_count" indicating
the number of specified devices (not scores). So after the maximum number
of devices arguments were interpreted, the appearance of next score argument
caused the error even when the number of interpreted scores arguments had
not exceeded the maximum.

This patch fixes storage_mon so that it uses the local variable "score_count"
indicating the number of specified scores, to check whether arguments for
scores are specified more than the upper limit.

### DIFF
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -154,7 +154,7 @@ int main(int argc, char *argv[])
 				}
 				break;
 			case 's':
-				if (device_count < MAX_DEVICES) {
+				if (score_count < MAX_DEVICES) {
 					int score = atoi(optarg);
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1797, opened 2022-08-10
storage_mon: Use the O_DIRECT flag in open() to eliminate cache effects

inouekazu:
Use of O_DIRECT in open() eliminates the possibility of false negatives resulting from buffer cache reads.

please refer to https://lists.clusterlabs.org/pipermail/users/2022-August/030459.html for background.

### DIFF
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -38,7 +38,6 @@ static int open_device(const char *device, int verbose)
 	uint64_t devsize;
 	off_t seek_spot;
 
-#if defined(__linux__) || defined(__FreeBSD__)
 	device_fd = open(device, O_RDONLY|O_DIRECT);
 	if (device_fd >= 0) {
 		return device_fd;
@@ -46,7 +45,6 @@ static int open_device(const char *device, int verbose)
 		fprintf(stderr, "Failed to open %s: %s\n", device, strerror(errno));
 		return -1;
 	}
-#endif
 
 	device_fd = open(device, O_RDONLY);
 	if (device_fd < 0) {
@@ -100,7 +98,11 @@ static void *test_device(const char *device, int verbose, int inject_error_perce
 		exit(-1);
 	}
 
+#ifdef __FreeBSD__
+	ioctl(device_fd, DIOCGSECTORSIZE, &sec_size);
+#else
 	ioctl(device_fd, BLKSSZGET, &sec_size);
+#endif
 	if (sec_size == 0) {
 		fprintf(stderr, "Failed to stat %s: %s\n", device, strerror(errno));
 		goto error;
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -63,7 +63,7 @@ static void *test_device(const char *device, int verbose, int inject_error_perce
 	res = ioctl(device_fd, BLKGETSIZE64, &devsize);
 #endif
 	if (res < 0) {
-		fprintf(stderr, "Failed to stat %s: %s\n", device, strerror(errno));
+		fprintf(stderr, "Failed to get device size for %s: %s\n", device, strerror(errno));
 		goto error;
 	}
 	if (verbose) {
@@ -93,7 +93,7 @@ static void *test_device(const char *device, int verbose, int inject_error_perce
 		res = ioctl(device_fd, BLKSSZGET, &sec_size);
 #endif
 		if (res < 0) {
-			fprintf(stderr, "Failed to stat %s: %s\n", device, strerror(errno));
+			fprintf(stderr, "Failed to get block device sector size for %s: %s\n", device, strerror(errno));
 			goto error;
 		}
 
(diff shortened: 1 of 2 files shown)


fabbione:
test this please

chrissie-c:
I agree with Ken (in the above thread) that there should be a fallback if O_DIRECT fails. There are filesystems and options even on Linux that don't always support O_DIRECT and they shouldn't be ecxluded or errored

inouekazu:
- https://github.com/ClusterLabs/resource-agents/pull/1797/commits/ce4e632f29ed6b86b82a959eac5844655baed153
   - According to https://ci.kronosnet.org/job/resource-agents-build-all-voting/1530/, build targets are FreeBSD and RHEL etc., so I deleted ```#if defined(__linux__) || defined(__FreeBSD__)```.
   - Fixed FreeBSD build failure.
 - The open() manual says it returns EINVAL for filesystems that do not support O_DIRECT. Are there any issues that the implementation below does not address?
   - https://github.com/ClusterLabs/resource-agents/pull/1797/files#diff-91e236cc5ccce26f5a934e98aef5f1082251a8676e999b5364add5603538483dR44

chrissie-c (changes requested):
There are few things I'm not clear/happy about still.

Is there a reason that the random lseek doesn't happen with O_DIRECT? I must admit I haven't looked into it in detail but it seems odd that the (hopefully) majority test case only checks the start of the device.

Also, I don't understand why the error checking for the ioctl is done on the sec_size value rather than the ioctl return. If that's fine then a comment saying why would be helpful. Maybe we need both?

In the case where O_DIRECT is not being used, we can skip the alignment calls (ioctl & posix_memalign) - they are just another potential cause of failure which we can do without - we're looking for disk errors here so the fewer syscalls that can fail the better.

wenningerk:
> 
> Also, I don't understand why the error checking for the ioctl is done on the sec_size value rather than the ioctl return. If that's fine then a comment saying why would be helpful. Maybe we need both?
> 
Some kind of shortcut for 'ioctl has to succeed and not return bogus'.
But some comment would of course make sense.
Same code-pattern btw. has been in sbd since ever.

inouekazu:
- https://github.com/ClusterLabs/resource-agents/pull/1797/commits/7a0aaa0dfdebeab3fae9fe9ddc412c3d1f610273
   - reflected the https://github.com/ClusterLabs/resource-agents/pull/1797#pullrequestreview-1080084034.

chrissie-c:
retest this please

chrissie-c:
retest this please

tools/storage_mon.c
```diff
@@ -72,65 +77,58 @@ static int open_device(const char *device, int verbose)
...
 #else
-	ioctl(device_fd, BLKSSZGET, &sec_size);
+		res = ioctl(device_fd, BLKSSZGET, &sec_size);
 #endif
-	if (sec_size == 0) {
-		fprintf(stderr, "Failed to stat %s: %s\n", device, strerror(errno));
-		goto error;
-	}
+		if (res < 0) {
+			fprintf(stderr, "Failed to stat %s: %s\n", device, strerror(errno));
```
wenningerk, line 96: why don't we simply say that getting the block-size did fail?

tools/storage_mon.c
```diff
@@ -72,65 +77,58 @@ static int open_device(const char *device, int verbose)
...
+			goto error;
+		}
 
-	if (posix_memalign(&buffer, sysconf(_SC_PAGESIZE), sec_size) != 0) {
-		fprintf(stderr, "Failed to allocate aligned memory: %s\n", strerror(errno));
-		goto error;
-	}
+		if (posix_memalign(&buffer, sysconf(_SC_PAGESIZE), sec_size) != 0) {
+			fprintf(stderr, "Failed to allocate aligned memory: %s\n", strerror(errno));
```
wenningerk, line 101: What I've read alignment should be more block-size. That makes sense thinking of a contiguous
block of memory cut into block-size chunks that you would like to be able to pass in any combination.
Usually it will probably be 512 bytes block-size vs. 4k pagesize and thus Ok.
Block-size of many disks being 4k nowadays I would expect that to become larger in future.
(or already in special cases like maybe raids, ssds with their large erase-block-size, shingled-write disks, ...)
max(block-size, pagesize) might be a way to satisfy everybody as much as possible.

wenningerk: sbd is using block-size - at least since I've reworked that code-part a bit but iirc that was there since ever.

wenningerk: aah - crap you aren't using aio here. So maybe forget what I've said.

tools/storage_mon.c
inouekazu: I agree. https://github.com/ClusterLabs/resource-agents/pull/1797/commits/db97e055a17526cec056c595844a9d8851e3ee19

oalbrigt:
Thanks.

PR #1808, opened 2022-09-22
LVM-activate: adding new parameter auto_drop_lock

ldzhong:
If VG is activated exclusively in the cluster and the lvmlockd daemon is killed, the EX locks for LV will become orphaned in the lockspace. Then LV will fail to be activated when cluster trying to bring it up again. With this new parameter enabled, we drop the orphaned locks first left in previous lockspace before activation.

### DIFF
--- a/heartbeat/LVM-activate
+++ b/heartbeat/LVM-activate
@@ -50,6 +50,7 @@ OCF_RESKEY_vg_access_mode_default=""
 OCF_RESKEY_activation_mode_default="exclusive"
 OCF_RESKEY_tag_default="pacemaker"
 OCF_RESKEY_partial_activation_default="false"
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

ldzhong:
Thanks for the review. The ideal resolution to this problem would be the lvmlockd --adopt option, but as you said the lvmlockd daemon should be restarted while persistent LV locks are still in place, and it doesn't apply in cluster environment as cluster manager normally will bring down the VG resource first before restarting lvmlockd. I agree with the second point here, which obviously will cause regression. It seems the agent is also not a perfect place to fix this problem.
I think this pull request can be closed for now. I'll figure out if there's any other resolution. Thanks for your time.

PR #1811, opened 2022-09-27
dev-guide: Correct default branch name

kpfleming:


fabbione:
ok to test

fabbione:
thanks for the patch

Issue #1809, opened 2022-09-22
storage_mon: child processes remain when stuck in read() etc


inouekazu:
When an I/O unresponsive failure occurs, the child process of storage_mon remains.
```
# ./storage_mon --device /dev/mapper/mpathc --score 1 --verbose
Testing device /dev/mapper/mpathc
/dev/mapper/mpathc: opened with O_DIRECT, size=536870912000
/dev/mapper/mpathc: reading from pos 727385088
Thread for device /dev/mapper/mpathc did not complete in time
Final score is 1
```
```
# ps -ef | grep storage_mon
root      710237  697460  0 13:17 pts/1    00:00:00 ./storage_mon --device /dev/mapper/mpathc --score 1 --verbose
root      710238  710237  0 13:17 pts/1    00:00:00 ./storage_mon --device /dev/mapper/mpathc --score 1 --verbose
```
```
# ps -ef | grep storage_mon
root      710238       1  0 13:17 pts/1    00:00:00 ./storage_mon --device /dev/mapper/mpathc --score 1 --verbose
```
I got `strace -ff -tt -T -o storage_mon ./storage_mon --device /dev/mapper/mpathc --score 1`.
 * __parent process__ exits,
   ```
   # cat storage_mon.629992
   14:45:20.173791 execve("./storage_mon", ["./storage_mon", "--device", "/dev/mapper/mpathc", "--score", "1", "--verbose"], 0x7fff94c9e6a0 /* 26 vars */) = 0 <0.000605>
   (snip)
   14:45:20.193591 clone(child_stack=NULL, flags=CLONE_CHILD_CLEARTID|CLONE_CHILD_SETTID|SIGCHLD, child_tidptr=0x7fba97913250) = 629993 <0.000246>
   14:45:20.194002 wait4(629993, 0x7fff320ea9a8, WNOHANG|WSTOPPED|WCONTINUED, NULL) = 0 <0.000023>
   14:45:20.194104 nanosleep({tv_sec=0, tv_nsec=100000000}, NULL) = 0 <0.100166>
   (snip)
   14:45:29.962911 wait4(629993, 0x7fff320ea9a8, WNOHANG|WSTOPPED|WCONTINUED, NULL) = 0 <0.000127>
   14:45:29.963242 nanosleep({tv_sec=0, tv_nsec=100000000}, NULL) = 0 <0.100199>
[...]

PR #1812, opened 2022-10-13
storage-mon: storage_mon updates attribute, waits for child to finish

inouekazu:
This patch is intended to avoid the issue (https://github.com/ClusterLabs/resource-agents/issues/1809) of an increasing number of child processes that do not finish.

### DIFF
--- a/heartbeat/storage-mon.in
+++ b/heartbeat/storage-mon.in
@@ -49,7 +49,6 @@
 
 #
 STORAGEMON=$HA_BIN/storage_mon
-ATTRDUP=/usr/sbin/attrd_updater
 
 OCF_RESKEY_CRM_meta_interval_default="0"
 OCF_RESKEY_io_timeout_default="10"
@@ -209,12 +208,8 @@ storage-mon_monitor() {
 	fi
 	$STORAGEMON $cmdline
 	if [ $? -ne 0 ]; then
-		status="red"
-	else
-		status="green"
+		return $OCF_ERR_GENERIC
 	fi
-
-	"$ATTRDUP" -n "#health-${OCF_RESOURCE_INSTANCE}" -U "$status" -d "5s"
 	return $OCF_SUCCESS
 }
 
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -159,6 +159,10 @@ int main(int argc, char *argv[])
 	time_t start_time;
 	size_t i;
 	int final_score = 0;
+	char attrname[512] = "#health-";
+	const char *str;
+	pid_t pid;
+	int wstatus;
 	int opt, option_index;
 	int verbose = 0;
 	int inject_error_percent = 0;
@@ -252,23 +256,20 @@ int main(int argc, char *argv[])
 		}
 	}
 
-	/* See if they have finished */
+	/* See if threads have finished until timeout */
 	clock_gettime(CLOCK_REALTIME, &ts);
 	start_time = ts.tv_sec;
 
 	while ((finished_count < device_count) && ((start_time + timeout) > ts.tv_sec)) {
 		for (i=0; i<device_count; i++) {
-			int wstatus;
-			pid_t w;
-
 			if (test_forks[i] > 0) {
-				w = waitpid(test_forks[i], &wstatus, WUNTRACED | WNOHANG | WCONTINUED);
-				if (w < 0) {
+				pid = waitpid(test_forks[i], &wstatus, WUNTRACED | WNOHANG | WCONTINUED);
+				if (pid < 0) {
 					fprintf(stderr, "waitpid on %s failed: %s\n", devices[i], strerror(errno));
 					return -1;
 				}
 
-				if (w == test_forks[i]) {
+				if (pid == test_forks[i]) {
 					if (WIFEXITED(wstatus)) {
 						if (WEXITSTATUS(wstatus) != 0) {
 							syslog(LOG_ERR, "Error reading from device %s", devices[i]);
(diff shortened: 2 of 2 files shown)


wenningerk:
I do understand how you moved updating the status-attribute from the RA to the storage_mon-tool to give an update to the cluster without having to return from the RA (and the storage_mon-tool as a precondition to that).
But does that really prevent creation of a lot of defunct-processes as pacemaker is gonna shoot the RA after timeout.
Of course selecting a large timeout you can still keep the rate of defunct-processes under control somehow.
Don't remember how pacemaker will handle the case where timeout > monitor-rate though.

inouekazu:
If an I/O unresponsive failure occurs with the following settings:
```
pcs property set node-health-strategy=only-green
pcs resource create storage-mon ocf:heartbeat:storage-mon drives="/dev/mapper/mpathc"
pcs resource clone storage-mon storage-mon-clone
pcs resource meta storage-mon-clone allow-unhealthy-nodes=true
pcs stonith create fence1-ipmilan fence_ipmilan (snip)
pcs stonith create fence2-ipmilan fence_ipmilan (snip)
```
1. storage-mon monitor is executed,
```
UID          PID    PPID  C STIME TTY          TIME CMD
root        5526    5164  0 12:18 ?        00:00:00 /usr/bin/bash /usr/lib/ocf/resource.d/heartbeat/storage-mon monitor
root        5529    5526  0 12:18 ?        00:00:00 /usr/libexec/heartbeat/storage_mon --device /dev/mapper/mpathc --score 1 --timeout 10
root        5530    5529  0 12:18 ?        00:00:00 /usr/libexec/heartbeat/storage_mon --device /dev/mapper/mpathc --score 1 --timeout 10
```
2. and timeout occurs.
```
Node List:
  * Node dl380g8a: online (health is RED)
  * Online: [ dl380g8b ]

Full List of Resources:
  * Clone Set: storage-mon-clone [storage-mon]:
    * storage-mon       (ocf::heartbeat:storage-mon):    FAILED dl380g8a
    * Started: [ dl380g8b ]
  * fence1-ipmilan      (stonith:fence_ipmilan):         Started dl380g8b
  * fence2-ipmilan      (stonith:fence_ipmilan):         Stopped

Migration Summary:
  * Node: dl380g8a:
    * storage-mon: migration-threshold=1000000 fail-count=1 last-failure='Fri Oct 14 12:20:36 2022'

Failed Resource Actions:
  * storage-mon_monitor_30000 on dl380g8a 'error' (1): call=17, status='Timed Out', exitreason='Resource agent did not complete within 2m', last-rc-change='Fri Oct 14 12:20:36 2022', queued=0ms, exec=0ms
```
3. Next, stop is executed,
```
UID          PID    PPID  C STIME TTY          TIME CMD
root        5530       1  0 12:18 ?        00:00:00 /usr/libexec/heartbeat/storage_mon --device /dev/mapper/mpathc --score 1 --timeout 10
root        5641    5164  0 12:20 ?        00:00:00 /usr/bin/bash /usr/lib/ocf/resource.d/heartbeat/storage-mon stop
root        5644    5641  0 12:20 ?        00:00:00 /usr/libexec/heartbeat/storage_mon --device /dev/mapper/mpathc --score 1 --timeout 10
root        5645    5644  0 12:20 ?        00:00:00 /usr/libexec/heartbeat/storage_mon --device /dev/mapper/mpathc --score 1 --timeout 10
```
4. and timeout also occurs. As a result, this node is fenced.
```
Node List:
  * Node dl380g8a: UNCLEAN (online) (health is RED)
  * Online: [ dl380g8b ]

Full List of Resources:
  * Clone Set: storage-mon-clone [storage-mon]:
    * storage-mon       (ocf::heartbeat:storage-mon):    FAILED dl380g8a
    * Started: [ dl380g8b ]
  * fence1-ipmilan      (stonith:fence_ipmilan):         Started dl380g8b
  * fence2-ipmilan      (stonith:fence_ipmilan):         Stopped

Migration Summary:
  * Node: dl380g8a:
    * storage-mon: migration-threshold=1000000 fail-count=1000000 last-failure='Fri Oct 14 12:22:37 2022'

Failed Resource Actions:
  * storage-mon_stop_0 on dl380g8a 'error' (1): call=22, status='Timed Out', exitreason='Resource agent did not complete within 2m', last-rc-change='Fri Oct 14 12:20:37 2022', queued=0ms, exec=120004ms

Pending Fencing Actions:
  * reboot of dl380g8a pending: client=pacemaker-controld.2805184, origin=dl380g8b
```
I think that it is not avoidable for the process to be stuck, but it is avoidable for the process to keep increasing.

wenningerk:
True! I guess there is no way around these processes kept hanging. And I didn't say I knew one or you should find one ;-)
Just wanted to state that if running into timeouts and being called repeatedly by pacemaker we will still see the hanging
processes piling up still - although maybe with a little bit more control than before.
If you are running the health-resource in a way that a timeouting monitor and subsequent stop will lead to node-fencing
you are right that there won't be much repeating done by pacemaker. Unless maybe that happens already during start.
All depending on how the resource is configured to react on failures of course.
In case of these health-resources, question is, if we really want them to fail and trigger immediate fencing of a node.
Idea of health, as I see it, is to still keep running what is running happily on a node - even if unhealthy - while not starting
any new resources and migrating away other stuff at some point.

Side I was approaching this topic from was verification of devices used by sbd poison-pill (in fence_sbd - fence-agent).
In this case you definitely don't want immediate fencing of the node as everything that isn't working
for now is that the node isn't able to fence another node (or not even that as long as e.g. just one out of 3 disks is
failing leaving behind a hanging process). If the node is foreseen to be a target for poison-pills as
well sbd-daemon will in parallel check availability of the disk(s) and trigger suicide once the node isn't able to
properly access a quorate number of disks anymore.
Hope that this example explains cases where it would be interesting to check for a disk, maybe creating a hanging
process, while there is still the need for the node to survive and of course we still would like for those processes not
to pile up.
So I thought we could gather ideas how to tackle these cases.

Things I was thinking of would be to return (with a negative monitor result or not) before a timeout but keeping
the process ids of still running processes memorized somehow/where as to properly react with the next monitor
call. Other approach would be to keep a daemon running that keeps this data and possibly re-trigger tests and
have the actual RA to just communicate with that daemon. iirc the latter pattern has been discussed with certain
kinds of resources earlier already (with the plugin-interface of the legacy fenced-implementation it was even
possible to run such a daemon as part of fenced).
In the above case of sbd poison-pill devices it of course comes handy that there usually is already a daemon
running so that we could easily instead of directly checking a device, ask that already running daemon, that
is anyway doing these checks periodically (at quite a high frequency that can hardly be achieved with
pacemaker resource monitoring btw.).

Excuse my excursion into sbd but I think we have some common issues/patterns here and I thought
it might add some substance to the discussion.
Btw. since OCF 1.1 we have the possibility to have a resource running in a degraded state instead
of having to fail if something isn't right as it should be. That might be interesting for health-resources -
at least under certain circumstances - to show that something isn't right in cluster status (more
fine-grained than just seeing the whole node being e.g. yellow).

inouekazu:
> Things I was thinking of would be to return (with a negative monitor result or not) before a timeout but keeping
> the process ids of still running processes memorized somehow/where as to properly react with the next monitor
> call.

Great, this will allow us to continue monitoring without increasing the number of processes.

> Other approach would be to keep a daemon running that keeps this data and possibly re-trigger tests and
> have the actual RA to just communicate with that daemon.

Although this approach has a proven track record [1], it was a major change, so it was only considered.

I will also consider with the members involved in pm_diskd [1].

[1] https://github.com/linux-ha-japan/pm_diskd : Japanese users used this agent with pacemaker-1.x to monitor disks.

PR #1814, opened 2022-10-31
storage-mon: daemonize storage_mon to deal with I/O hangs

inouekazu:
This patch is intended to avoid the issue (#1809) of an increasing number of child processes that do not finish.
 * Based on discussion in https://github.com/ClusterLabs/resource-agents/pull/1812, changed storage_mon to run as a daemon.

### DIFF
--- a/heartbeat/storage-mon.in
+++ b/heartbeat/storage-mon.in
@@ -49,19 +49,20 @@
 
 #
 STORAGEMON=$HA_BIN/storage_mon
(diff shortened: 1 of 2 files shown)


HideoYamauchi:
Hi All,

No one seems to have any comments, but we think this fix is a good one.
what do other people think?

Best Regards,
Hideo Yamauchi.

oalbrigt:
We're having some discussions about this, and will come back to you next week.

PR #1815, opened 2022-11-03
iface-macvlan: new resource agent

vafgoettlich:
We needed a Resource-Agent for adding / removing macvlan devices. This is because we are using a Gateway in one place which only supports forwarding to a destination if you keep the MAC-Address for a specific IP.
I found other people searching the Web with the need of a macvlan RA, so I had the Idea to publish this. It is currently limited to bridge type because I only have tested it with type bridge.

### DIFF
--- a/heartbeat/iface-macvlan
+++ b/heartbeat/iface-macvlan
@@ -97,7 +97,7 @@ macvlan_meta_data() {
       <content type="string"/>
     </parameter>
 
-    <parameter name="mode" unique="1">
+    <parameter name="mode" unique="0">
       <longdesc lang="en">
         Define the name of the MACVLAN mode (currently only bridge is supported).
       </longdesc>
@@ -107,7 +107,7 @@ macvlan_meta_data() {
       <content type="string" default="${OCF_RESKEY_mode_default}" />
     </parameter>
 
-    <parameter name="mac">
+    <parameter name="mac" unique="1">
      <longdesc lang="en">
       Set the interface MAC address explicitly.
      </longdesc>
@@ -158,13 +158,13 @@ macvlan_validate() {
 	check_binary $IP2UTIL
 
 	if [ -z "$OCF_RESKEY_device" ]; then
-		ocf_log err "Invalid OCF_RESKEY_device: value cannot be empty"
-		return 1
+		ocf_log err "Invalid device: value cannot be empty"
+		return $OCF_ERR_CONFIGURED
 	fi
 
 	if [ -z "$OCF_RESKEY_name" ]; then
-		ocf_log err "Invalid OCF_RESKEY_name: value cannot be empty"
-		return 1
+		ocf_log err "Invalid name: value cannot be empty"
+		return $OCF_ERR_CONFIGURED
 	fi
 
 	# the echo .. is the equivalent of strlen in bash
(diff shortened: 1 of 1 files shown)


heartbeat/iface-macvlan
```diff
@@ -0,0 +1,363 @@
...
+		return 1
+	fi
+	return 0
+}
+
+macvlan_validate() {
+	check_binary $IP2UTIL
+
+	if [ -z "$OCF_RESKEY_device" ]; then
+		ocf_log err "Invalid OCF_RESKEY_device: value cannot be empty"
```
oalbrigt, line 161: You should remove OCF_RESKEY_ in the errors, as that's appended by Pacemaker when you set e.g. device for the resource. OCF_RESKEY_ names are only visible to the users if they manually test it outside of Pacemaker.

And you should `return $OCF_ERR_CONFIGURED`. See <https://github.com/ClusterLabs/resource-agents/blob/main/doc/dev-guides/ra-dev-guide.asc#return-codes> for description of the return codes.

heartbeat/iface-macvlan
```diff
@@ -0,0 +1,363 @@
...
+		return 1
+	fi
+
+	return 0
+}
+
+macvlan_check() {
+	if [ -e "/sys/class/net/$OCF_RESKEY_name" ]; then
+		if [ ! -e "$HA_RSCTMP/iface-macvlan.$OCF_RESKEY_name" ]; then
+			return $OCF_ERR_GENERIC
```
oalbrigt, line 201: You might want to use OCF_ERR_CONFIGURED or _INSTALLED in some occurences in this function as well (depending on whether it would be caused by misconfiguration or lack of e.g. cli util).

oalbrigt (changes requested):
Looks good in general, so I've added some comments on some minor things to improve.

heartbeat/iface-macvlan
```diff
@@ -0,0 +1,363 @@
...
+      <longdesc lang="en">
+        Define the MACVLAN NAME. It has to be a valid interface name (max 15 characters).
+      </longdesc>
+      <shortdesc lang="en">
+        Define the MACVLAN NAME.
+      </shortdesc>
+      <content type="string"/>
+    </parameter>
+
+    <parameter name="mode" unique="1">
```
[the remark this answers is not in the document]
vafgoettlich: Thanks, I fixed this.

heartbeat/iface-macvlan
vafgoettlich: I have checked that and now the name of the variable is reported correctly and the return codes are changed.

heartbeat/iface-macvlan
vafgoettlich: I only found one place where I have been sure to use another return code than OCF_ERR_GENERIC. I have already changed that.

heartbeat/iface-macvlan
```diff
@@ -0,0 +1,363 @@
...
+      <longdesc lang="en">
+        Define the interface where MACVLAN should be attached.
+      </longdesc>
+      <shortdesc lang="en">
+        Network interface.
+      </shortdesc>
+      <content type="string"/>
+    </parameter>
+
+    <parameter name="name" unique="0" required="1">
```
oalbrigt, line 90: I guess this should be unique too?

vafgoettlich: You are right, I have fixed it, thanks.

oalbrigt (approved):

oalbrigt:
ok to test

oalbrigt:
PR to add to Makefiles: https://github.com/ClusterLabs/resource-agents/pull/1816

PR #1819, opened 2022-11-16
docker/podman/zabbixserver:replace error log messages with calls to o…

XingWei-Liu:
hi, i change  some  log messages  from "ocf_log err" to "ocf_exit_reason"


oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1825, opened 2022-12-05
SAPInstance: be more resilient against broken kill.sap files

fmherschel:
The code using pid could break, if kill.sap includes unexpected content. Now the scanner checks more focused, if the third element is a number and only let the number pass the scanner.


oalbrigt:
LGTM.

varkoly (approved):
OK

PR #1827, opened 2022-12-13
Please don't suggest /tmp/mysql-proxy.sock, that is bad practice.

varkoly:
Use one of the directories where this socket can be provided securely


fabbione:
test this please

fabbione (approved):
makes sense.

thanks!
Fabio

PR #1829, opened 2022-12-17
podman: add info log when returning OCF_NOT_RUNNING

freedge:
there are multiple cases where OCF_NOT_RUNNING is returned and unfortunately, there are rare cases where the container is, in fact, running.

We add an info log so we have something to look at in case the heartbeat fails for seemingly no reason.

### DIFF
--- a/heartbeat/podman
+++ b/heartbeat/podman
@@ -207,6 +207,7 @@ monitor_cmd_exec()
 	# 255: podman 2+: container not running
 	case "$rc" in
 		125|126|255)
+			ocf_log info "monitor cmd failed (rc=$rc), output: $out, assume container is not running"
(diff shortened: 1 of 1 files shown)


freedge:
that was the error I was looking for fyi

    Dec 20 08:53:56  podman(rabbitmq-bundle-podman-1)[435653]:    INFO: monitor cmd failed (rc=255), output: Error: an exec session with ID 207b204101475831fa87c972f13dfff8872e4f33005444fb8a572f6ffc49077f already exists: exec session already exists

david-hill:
Could we add a retry here instead ?  If it's a race condition that we sometimes hit, re-trying should pass the second time and avoid a costly fencing event.

PR #1838, opened 2023-01-24
db2: avoid eval with unsanitized values

varkoly:
I'm not sure if you can get characters in here that are problematic, but better not to use unsanitized string than be sorry


oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1842, opened 2023-02-16
nfsserver: fix "server scope" to live with additional drop-in files

zzhou1:
Symptom:

The distro may have additional drop-in files to override `ExecStart=`, eg. openSUSE. In that case, the result drop-in file is messed up with several ExecStart= generated. For example,

cat
/run/systemd/system/nfs-server.service.d/51-resource-agents-unshare-uts.conf

[Service]
EnvironmentFile=/run/sysconfig/nfs-server-scope

ExecStart=
ExecStart=/usr/bin/unshare --uts /bin/sh -c 'hostname ${NFS_SERVER_SCOPE}; exec "$@"' -- /usr/sbin/rpc.nfsd ExecStart=/usr/bin/unshare --uts /bin/sh -c 'hostname ${NFS_SERVER_SCOPE}; exec "$@"' -- 
ExecStart=-/usr/bin/unshare --uts /bin/sh -c 'hostname ${NFS_SERVER_SCOPE}; exec "$@"' -- /usr/sbin/rpc.nfsd $NFSD_OPTIONS

Indeed, only the last ExecStart= is needed.

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -711,7 +711,9 @@ inject_unshare_uts_name_into_systemd_units ()
 		test -d "$dir" || mkdir -p "$dir"
 		test -e "$dropin" && rm -f "$dropin"
 
-		edited_exec_start=$(systemctl cat $svc | sed -ne "s#^ExecStart=\\([-+:!@]*\\)\\(.*\\)#ExecStart=\\1/usr/bin/unshare --uts /bin/sh -c 'hostname \${NFS_SERVER_SCOPE}; exec \"\$@\"' -- \\2#p")
+		# NOTE: additional ExecStart= might exist in the drop-in files, eg. openSUSE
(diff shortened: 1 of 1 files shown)


zzhou1:
@lge

zzhou1:
> The added / might cause issues if there are anyone not using full path in their ExecStart=.

hmm, you are right. Let me remove "/" prefix.

oalbrigt:
You can remove the `.`. It already does .*, which covers it.

zzhou1:
Though remove it is fine, I'm thinking to leave that "." to reinforce at least one character to filter out the empty one.

oalbrigt:
You should use .+ for that (you might have to use .\+.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1853, opened 2023-03-23
ZFS: add non-blocking check if pool was imported

doma2203:
The function  `zpool_is_imported` can be improved to use ZFS kstats instead of blocking `zpool list` calls for newer releases of ZFS.

### DIFF
--- a/heartbeat/ZFS
+++ b/heartbeat/ZFS
@@ -89,7 +89,7 @@ zpool_is_imported () {
         return `[ -d  /proc/spl/kstat/zfs/"${OCF_RESKEY_pool}" ]`
     else
         # If ZFS kstats do not exists, fallback to the standard check
-		return zpool list -H "$OCF_RESKEY_pool" > /dev/null
+		return `zpool list -H "$OCF_RESKEY_pool" > /dev/null`
 	fi
 }
 
--- a/heartbeat/ZFS
+++ b/heartbeat/ZFS
@@ -83,7 +83,14 @@ END
 }
 
 zpool_is_imported () {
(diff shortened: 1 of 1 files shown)


heartbeat/ZFS
```diff
@@ -83,7 +83,14 @@ END
 }
 
 zpool_is_imported () {
-	zpool list -H "$OCF_RESKEY_pool" > /dev/null
+	# Check if ZFS kstats exists
```
oalbrigt, line 86: You should Tab indentation like the rest of the agent.

heartbeat/ZFS
```diff
@@ -83,7 +83,14 @@ END
...
+    if [ -d  /proc/spl/kstat/zfs/ ] ; then
+		# Check the existence of kstats for the pool. If the stats exists, the pool was imported.
+        return `[ -d  /proc/spl/kstat/zfs/"${OCF_RESKEY_pool}" ]`
```
oalbrigt, line 89: I think it's better to not spawn a new process for this:
```
[ -d  /proc/spl/kstat/zfs/"${OCF_RESKEY_pool}" ]
return $?
```

oalbrigt (changes requested):

heartbeat/ZFS
doma2203: Ok. Maybe we can also omit `$?` in this case - we'll return the status of the last command anyway.

doma2203: Yes, it's a better idea.

oalbrigt (approved):

oalbrigt:
retest this please

oalbrigt:
Thanks.

PR #1858, opened 2023-03-30
IPv6addr: expect ping/pong delay

### LINKED ISSUE #1855: Occasional false positive "down" reports from IPv6addr "monitor" action
The "monitor" action of IPv6addr is sending an ICMPv6 echo request to the given local address: https://github.com/ClusterLabs/resource-agents/blob/d6b954890b496fcdd8a76d7c2dd44a36fa0ad42c/heartbeat/IPv6addr.c#L630

and is the expecting to receive the respective echo response immediately, without delay, using `MSG_DONTWAIT`: https://github.com/ClusterLabs/resource-agents/blob/d6b954890b496fcdd8a76d7c2dd44a36fa0ad42c/heartbeat/IPv6addr.c#L647

This works fine most of the time, but occasionally under heavy network load, the echo response is not immediately available, and the `recvmsg` fails with `EAGAIN`, leading to a false positive down event ("not running") on the resource:

```
Mar 28 18:53:59 sp1 pacemaker-schedulerd[41843]:  warning: Unexpected result (not running) was recorded for monitor of p_vip_neth0_v6_1 on sp1 at Mar 28 00:02:38 2023 
```

rfuchs:
Under heavy network load, the echo response to an echo request that was just sent may not immediately be available for reading, with recvmsg(MSG_DONTWAIT) failing with EAGAIN. This leads to occasional false positive "not running" events.

This wraps the recvmsg() within a poll() loop with a short timeout (10 ms) and retries reading the echo response up to 3 times, in case poll() was interrupted by some other event (e.g. EINTR).

Closes #1855

### DIFF
--- a/heartbeat/IPv6addr.c
+++ b/heartbeat/IPv6addr.c
@@ -104,6 +104,7 @@
 #include <syslog.h>
 #include <signal.h>
 #include <errno.h>
+#include <poll.h>
 #include <clplumbing/cl_log.h>
 
 
@@ -606,6 +607,8 @@ is_addr6_available(struct in6_addr* addr6)
 	struct iovec			iov;
 	u_char				packet[MINPACKSIZE];
 	struct msghdr			msg;
+	int				i;
+	struct pollfd			pfd;
 
 	if ((icmp_sock = socket(AF_INET6, SOCK_RAW, IPPROTO_ICMPV6)) == -1) {
 		return -1;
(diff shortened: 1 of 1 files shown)


oalbrigt:
ok to test

rfuchs:
Any opinions on this? FTR we're actively hitting this bug in some of our deployments.

oalbrigt:
Thanks.

PR #1860, opened 2023-03-30
nfsserver: prepare the layout for the default /var/lib/nfs

zzhou1:
Some deployment might deploy "/var/lib/nfs" as the fresh mount point without explicitly providing "nfs_shared_infodir=" for a different place. This directory could be empty in the very beginning. In such case, we could leverage prepare_directory() to populate nfsserver information directories and files layout transparantly.


heartbeat/nfsserver
```diff
@@ -454,10 +454,6 @@ nfsserver_monitor ()
 
 prepare_directory ()
 {
-	if [ -z "$fp" ]; then
```
oalbrigt, line 457: We shouldnt remove this, or it will try to populate `/`.

You can add `if ! ocf_is_probe && [ -z "$fp" ]` to `nfsserver_validate()` to avoid this.

zzhou1: aha, you are right. My brain is messed in the late evening to copy&paste the code. I intended to set the empty $fp to /var/lib/nfs instead.

My impression of nfsserver_validate() is it is called for every monitor operation. Might be no need to check so frequently.

heartbeat/nfsserver
```diff
@@ -472,8 +472,9 @@ prepare_directory ()
 	[ -f "$fp/xtab" ] || touch "$fp/xtab"
 	[ -f "$fp/rmtab" ] || touch "$fp/rmtab"
 
-	dd if=/dev/urandom of=$fp/$STATD_DIR/state bs=1 count=4 >/dev/null 2>&1
-	[ -n "`id -u rpcuser 2>/dev/null`" -a "`id -g rpcuser 2>/dev/null`" ] && chown rpcuser.rpcuser "$fp/$STATD_DIR/state"
+	fs="$fp/$STATD_DIR/state"
+	[ -f "$fs" ] || dd if=/dev/urandom of=$fs bs=1 count=4 >/dev/null 2>&1
```
zzhou1, line 476: Meanwhile, my use case is to preserve "/var/lib/nfs" in a shared storage or drbd disk for failover. Hence, intend to avoid change "/var/lib/nfs/state" whenever start nfsserver.

Not sure if this as the big concerns to other use cases. This might need further improvement. I wonder.

CC. @dpartrid

zzhou1: > Meanwhile, my use case is to preserve "/var/lib/nfs" in a shared storage or drbd disk for failover. Hence, intend to avoid change "/var/lib/nfs/state" whenever start nfsserver.
> 
> Not sure if this as the big concerns to other use cases. This might need further improvement. I wonder.
> 

Go back to [the root commit 0434581](https://github.com/ClusterLabs/resource-agents/commit/0434581), I see the purpose, hence I revert my previous code change on "/var/lib/nfs/state". It is used for NFSv2 and NFSv3 actually. They are stateless protocol back then.

oalbrigt:
ok to test

oalbrigt:
Thanks.

PR #1861, opened 2023-04-10
exportfs: make the "fsid=" parameter optional

zzhou1:
Based on feedback [1] from the kernel developer @neilbrown regarding the NFS clustering use case, it has been determined that the fsid= parameter is now considered optional and safe to omit.

[1] https://bugzilla.suse.com/show_bug.cgi?id=1201271#c49 """
Since some time in 2007 NFS has used the UUID of a filesystem as the primary identifier for that filesystem, rather than using the device number.  So from that time there should have been reduced need for the "fsid=" option.  Probably there are some filesystems that this didn't work for.  btrfs has been problematic at time, particularly when subvols are exported.  But for quite some years this has all "just worked" at least for the major filesystems (ext4 xfs btrfs). [...] I would suggest getting rid of the use of fsid= altogether. [...] I'm confident that it was no longer an issue in SLE-12 and similarly not in SLE-15. """

### DIFF
--- a/heartbeat/exportfs
+++ b/heartbeat/exportfs
@@ -82,7 +82,7 @@ The directory or directories to export.
 <content type="string" />
 </parameter>
 
-<parameter name="fsid" unique="0" required="1">
+<parameter name="fsid" unique="0" required="0">
 <longdesc lang="en">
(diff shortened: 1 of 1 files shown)


oalbrigt:
Thanks.

PR #1863, opened 2023-04-26
Use -LS instead of -Ls as parameter to get the Transition Summary

varkoly:
The fact is, the option "-s" is basically used for showing resource allocation scores. So arguably it was not really appropriate/necessary for "-s" to additionally display "Transition Summary". Now -S have to be used.


oalbrigt:
ok to test

oalbrigt:
You need to rebase to solve the conflicts after I fixed the return values: https://github.com/ClusterLabs/resource-agents/pull/1854

varkoly:
Creating new PR

PR #1868, opened 2023-05-16
Filesystem: add signals parameter to set signals to send processes during stop-action

oalbrigt:


oalbrigt:
Closing in favor of https://github.com/ClusterLabs/resource-agents/pull/1878

PR #1871, opened 2023-05-26
Delay: increase stop, status and monitor timeouts to 40s to avoid failing with default values

oalbrigt:


oalbrigt:
retest this please

PR #1872, opened 2023-06-06
spec: Migrate to SPDX license

jfriesse:
Both Fedora and openSUSE now recommends to use SPDX shortname format for License.


oalbrigt:
Thanks.

PR #633, opened 2015-06-29
Use runuser intead of su command for SELinux enforcing mode

hashnao:
It is not recommended that using su command in order to change UID because it has several problems as introduced in the following blog, http://danwalsh.livejournal.com/55588.html. In addition, the current pgsql resource agent never fails to run in SELinux enforcing mode.
This patch uses runuser command instead of su in order to properly run the pgsql resource agent in SELinux enforcing mode.

PR #1880, opened 2023-07-18
postfix: Update permission check from su to runuser/su

Tnthr:
In the postfix resource agent, changed the check for a writable config directory to use sudo rather than su. Using su is prone to authentication failures while sudo works well.

### DIFF
--- a/heartbeat/postfix
+++ b/heartbeat/postfix
@@ -44,11 +44,11 @@ USAGE="Usage: $0 {start|stop|reload|monitor|validate-all|meta-data}";
 
 ##########################################################################
 
-# Check availability of the runuser command, otherwise use sudo
+# Check availability of the runuser command, otherwise use su
 if [ -x /sbin/runuser ]; then
-    SUDO=runuser
+    SU=runuser
 else
-    SUDO=sudo
+    SU=su
 fi
 
 usage() {
@@ -297,7 +297,7 @@ postfix_validate_all()
         if ocf_is_true $status_support; then
             user=`postconf $OPTION_CONFIG_DIR -h mail_owner 2>/dev/null`
             for dir in $data_dir; do
-                if ! $SUDO -u $user -- test -w $dir; then
+                if ! $SU -s /bin/sh -l $user -c "test -w $dir"; then
                     if ocf_is_probe; then
                         ocf_log info "Directory '$dir' is not writable by user '$user' during probe."
                     else
--- a/heartbeat/postfix
+++ b/heartbeat/postfix
@@ -297,7 +297,7 @@ postfix_validate_all()
         if ocf_is_true $status_support; then
             user=`postconf $OPTION_CONFIG_DIR -h mail_owner 2>/dev/null`
             for dir in $data_dir; do
-                if ! $SU -s /bin/sh -l $user -c "test -w $dir"; then
+                if ! $SU -s /bin/sh - $user -c "test -w $dir"; then
                     if ocf_is_probe; then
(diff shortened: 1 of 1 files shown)


oalbrigt:
su has been working fine for years in other agents. You might try adding the logic to use runuser for selinux though.

https://github.com/ClusterLabs/resource-agents/pull/633/files

Tnthr:
It does look like runuser may be the more appropriate command. I can work on integrating something similar to #633 into this RA as well.

For further clarification though. I had my postfix running well for about a year but with a distro upgrade from Ubuntu 22.04 to 22.10 I started seeing the issues with postfix. I'm still looking around to see the exact cause but it must be a default security setting that was changed with the new release. I don't have SELinux running either. I imagine as people eventually upgrade distro releases (assuming this might be a Ubuntu specific change) they may begin to see the same issue.

With that in mind I will modify this to add a check for runuser and use sudo as a fallback.

Tnthr:
Sorry for the clutter here. I'm still learning the finer details of git. Attempting to squash my three commits has now netted 6 commits. Anyway...

I changed the runuser check back to the -x as recommended assuming most distros have the binary in the same folder or at least symlinked.

heartbeat/postfix
```diff
@@ -44,6 +44,13 @@ USAGE="Usage: $0 {start|stop|reload|monitor|validate-all|meta-data}";
 
 ##########################################################################
 
+# Check availability of the runuser command, otherwise use sudo
+if [ -x /sbin/runuser ]; then
+    SUDO=runuser
+else
+    SUDO=sudo
```
oalbrigt, line 51: I think we should stay with su, as that's what all the agents are using (except pingd).

So if there are issues with new su on Ubuntu you should report the issue to them, and if there's an easy fix we can use it in our agents (if it doesnt break for older versions or similar).

oalbrigt:
No worries about the clutter. People make mistakes, and additional mistakes expected for people who are new to git 😄

heartbeat/postfix
Tnthr: Making a revision now to change it back to su.

As it turns out, if the user (postfix) doesn't exist in /etc/shadow, su will give an authentication failure every time no matter what. Sudo would work in that case anyway. But that shouldn't be an issue if you create your users correctly.

heartbeat/postfix
```diff
@@ -290,7 +297,7 @@ postfix_validate_all()
         if ocf_is_true $status_support; then
             user=`postconf $OPTION_CONFIG_DIR -h mail_owner 2>/dev/null`
             for dir in $data_dir; do
-                if ! su -s /bin/sh - $user -c "test -w $dir"; then
+                if ! $SU -s /bin/sh -l $user -c "test -w $dir"; then
```
oalbrigt, line 300: It should still use `-` and not `-l` here. No need to change parts that already work.

heartbeat/postfix
oalbrigt: Interresting. Yeah, that shouldnt be an issue for most people.

heartbeat/postfix
Tnthr: Minor change and not really necessary you’re right. My logic was to make it clearer to understand at a glance. The man page recommends using the full —login rather than the shortcut - so I went for a middle ground

oalbrigt: Yeah. I would rather have it the way it has been to avoid any issues with backporting or similar (e.g. *BSD su).

oalbrigt:
Thanks.

PR #1887, opened 2023-09-06
mysql-common: improve error message

XingWei-Liu:
hi , there are situations when MySQL fails to start and the error is sent to standard error/standard out and is not display on the terminal. then i add some log message in ocf_exit_reason . hope it helps

### DIFF
--- a/heartbeat/mysql-common.sh
+++ b/heartbeat/mysql-common.sh
@@ -254,7 +254,7 @@ mysql_common_start()
     while [ $start_wait = 1 ]; do
         if ! ps $pid > /dev/null 2>&1; then
             wait $pid
-            ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?), please check your installation"
+            ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?), please check your installation, log message you can check $OCF_RESKEY_log"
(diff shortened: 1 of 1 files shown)


heartbeat/mysql-common.sh
```diff
@@ -254,7 +254,7 @@ mysql_common_start()
     while [ $start_wait = 1 ]; do
         if ! ps $pid > /dev/null 2>&1; then
             wait $pid
-            ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?), please check your installation"
+            ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?), please check your installation, log message you can check $OCF_RESKEY_log"
```
oalbrigt, line 257: You can shorten it to `ocf_exit_reason "MySQL server failed to start (pid=$pid) (rc=$?). Check $OCF_RESKEY_log for details"`

XingWei-Liu: ok  i already change it

oalbrigt: Can you update the , to . as well? That way the capitalization is correct.

XingWei-Liu: ok  my fault  , its done , its there anther question

oalbrigt:
Thanks.

PR #1888, opened 2023-09-08
fix Delay resource monitor timeout

mengfansheng-git:
ocf:heartbeat:Delay RA fails monitor operations when using default settings


oalbrigt:
This patch should address the issue: https://github.com/ClusterLabs/resource-agents/pull/1871

mengfansheng-git:
截图_选择区域_20230908170941
截图_选择区域_20230908171002
The timeout is extended to 40 seconds and an error is reported during the monitor

PR #1898, opened 2023-11-01
findif.sh: fix loopback handling

vvidic:
`tools/ocft/IPaddr2` fails the loopback test because of the missing `table local` parameter:

```
$ ip -o -f inet route list match 127.0.0.3 scope host

$ ip -o -f inet route list match 127.0.0.3 table local scope host
local 127.0.0.0/8 dev lo proto kernel src 127.0.0.1
```

Also rename the function because it is called only in for the special loopback address case.

### DIFF
--- a/heartbeat/findif.sh
+++ b/heartbeat/findif.sh
@@ -29,10 +29,10 @@ prefixcheck() {
   fi
   return 0
 }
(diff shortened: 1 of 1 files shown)


oalbrigt:
Thanks.

PR #1902, opened 2023-11-20
storage_mon/findif: fix handler out of scope leak, unitialized value and check that netmaskbits != EOS

oalbrigt:


oalbrigt:
retest this please

PR #1903, opened 2023-11-22
storage_mon: use memset() to fix "uninitialized value" covscan error, as qb_ipcc_recv() will always set a message (according to honzaf)

oalbrigt:


tools/storage_mon.c
```diff
@@ -686,11 +687,7 @@ storage_mon_client(void)
 	/* greater than 0	: monitoring error. 		*/
 	/* -1			: communication system error.	*/
 	/* -2                   : Not all checks completed for first device in daemon mode. */ 
-	if (strnlen(response.message, 1)) {
-		rc = atoi(response.message);
-	} else {
-		rc = -1;
-	}
+	rc = atoi(response.message);
```
wenningerk, line 690: That alone would change the behavior right?
empty string now gives you rc == 0 instead of rc == -1.

oalbrigt: The message should always be set by `qb_ipcc_recv()`. The if/else was added yesterday to try to solve the possible issue detected by covscan.

wenningerk: aah ...

oalbrigt: There is a minimal chance it might happen, so I'm adding back the if/else, as the memset() has solved the new covscan issue with strnlen().

Thanks @wenningerk for pointing this out.

PR #1905, opened 2023-12-19
Filesystem: modify the check for the exit code of the child process.

SatomiOSAWA:
Hi All,

I found something a bit off with the check for the exit code of the child process.
It's in the stop operation.
According to TLDP,  the exit code when a process is signaled is expected to be "128 + signal number", 
not "127 + child exit code".
(FYI: https://tldp.org/LDP/abs/html/exitcodes.html)
So, I've done some small tweaks.
I hope this will help.

Best Regards,
SatomiOSAWA

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -704,8 +704,8 @@ timeout_child() {
 	wait $pid
 	ret=$?
 
-	# ret would be 127 + child exit code if the timeout expired
-	[ $ret -lt 128 ] && kill -s KILL $killer
+	# ret would be 128 + signal number if the timeout expired
+	[ $ret -lt 129 ] && kill -s KILL $killer
 	return $ret
 }
 fs_stop_loop() {


oalbrigt:
From the wait manpage:
If  id  specifies  a non-existent  process or job, the return status is 127.  If wait is interrupted by a signal, the return status will be greater than 128, as described under SIGNALS in bash(1).  Otherwise, the return status is the exit status of the last process or job waited for.

SatomiOSAWA:
Hi, @oalbrigt 

OH!! I understand now.
I'm so sorry for bothering you.
Now I think the code is fine as it is.
However, I still believe it's better to make adjustments to just the comment section.

Thank you.


Best Regards,
Satomi OSAWA

oalbrigt:
Updating the comment makes sense, so I'll look into that later this week.

SatomiOSAWA:
Hi, @oalbrigt

Thank you so much!

Satomi OSAWA

oalbrigt:
You're welcome.

PR #1909, opened 2024-01-16
LVM-activate: Avoid false positive for VG activation

arslan9292:
When lvm.conf file has `volume_list` parameter configured and the cluster is managing the shared storage using `system_id_source`, then the activation of the LV fails to happen. However it is reported as a success.

The fixes will avoid starting of `LVM-activate` resource when the cluster is configured with both `system_id_source` and `volume_list`.


teigland:
In LVM, volume_list has no relationship with system_id, and in it's perfectly fine to use them both together.  So, it's possible that this change could break a legitimate setup.  Could this patch check for specific values in volume_list that would indicate it's been mistakenly configured for tagging with LVM-activate?

teigland:
After more info I think this is a good solution.

PR #1910, opened 2024-01-24
galera: allow joiner to report non-Primary during initial IST

dciabrin:
It seems that with recent galera versions, when a galera node joins a cluster, there is a small time window where the node is connected to the primary component of the galera cluster, but it might still be preparing its IST. During this time, it can report itself as being 'not ready' and in 'non-primary' state.

Update the galera resource agent to allow the node to be in non-primary state, but only if running a "promote" operation. Any network partition during the promotion will be caught by the promote timeout.

In reworking the promotion code, we move the check for primary partition into the "galera_monitor" function. The check works as before for regular "monitor" or "probe" operations.

Related-Bug: rhbz#2255414

### DIFF
--- a/heartbeat/galera.in
+++ b/heartbeat/galera.in
@@ -822,6 +822,11 @@ galera_promote()
         return $rc
     fi
 
+    # At this point, the mysql pidfile is created on disk and the
+    # mysql server is reacheable via its UNIX socket. If we are a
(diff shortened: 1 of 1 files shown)


oalbrigt:
add to whitelist

oalbrigt:
Thanks.

PR #1918, opened 2024-02-29
Add WAS8 ocf script to support WAS8 server

15620230229:


oalbrigt:
Doesnt this have a lot of code in common with WAS or WAS6? If so I'd prefer to have the differences in case's or if's where e.g. `${0##*/} = WAS8` and make WAS8 a link.

PR #1919, opened 2024-03-07
Doc: Delay: Drop old comments

nrwahl2:
`stopdelay` and `mondelay` do not default to `startdelay`. Should have been done with fe8a807d.

Also remove comments that are redundant with the metadata, and remove descriptions of this as a "test" resource. This has valid uses in some production environments.


oalbrigt:
Thanks.

PR #1921, opened 2024-03-14
VirtualDomain: 'virsh list --all' can query the domain when the domain stops.

twtlpla:
'virsh list --all' can query the domain when the domain stops.


oalbrigt:
We do not want to change this logic. It has been working fine for years.

PR #1922, opened 2024-03-14
azure-lb: support IPv6 with Azure load balancer

varkoly:
Using the  "TCP6-LISTENER" causes socat to listen in a dual-stack manner like nc.


varkoly:
I have not touched the netcat part only the socat.

oalbrigt:
Thanks.

PR #1938, opened 2024-05-21
outscale: new resource agent

stephrobert:
This new resource agent adds functionality to manage a outscale elastic IP. It supports being configured as a primitive or as a master/slave set.

### DIFF
--- a/heartbeat/Makefile.am
+++ b/heartbeat/Makefile.am
@@ -153,6 +153,7 @@ ocf_SCRIPTS	      = AoEtarget		\
 			oraasm			\
 			oracle			\
 			oralsnr			\
+			osceip			\
 			ovsmonitor		\
 			pgagent			\
 			pgsql			\
--- a/heartbeat/osceip
+++ b/heartbeat/osceip
@@ -1,5 +1,4 @@
 #!/bin/sh
-set -x
 #
 #
 #    Manage Outscale Elastic IP with Pacemaker
@@ -94,14 +93,6 @@ Valid OSC-CLI profile name (see ~/.osc/config and 'osc configure')
 reserved elastic ip for vm instance
 </longdesc>
 <shortdesc lang="en">reserved elastic ip for vm instance</shortdesc>
-<content type="string" default="" />
-</parameter>
-
-<parameter name="allocation_id" unique="1" required="1">
-<longdesc lang="en">
-reserved allocation id for vm instance
-</longdesc>
-<shortdesc lang="en">reserved allocation id for vm instance</shortdesc>
 <content type="string" default="" />
 </parameter>
 
@@ -205,7 +196,6 @@ if [ -n "${OCF_RESKEY_region}" ]; then
 fi
 ELASTIC_IP="${OCF_RESKEY_elastic_ip}"
 INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
-ALLOCATION_ID="${OCF_RESKEY_allocation_id}"
 
 case $__OCF_ACTION in
     start)
--- a/heartbeat/osceip
+++ b/heartbeat/osceip
@@ -47,10 +47,14 @@
 OCF_RESKEY_osccli_default="/usr/local/bin/osc-cli"
 OCF_RESKEY_profile_default="default"
 OCF_RESKEY_region_default=""
+OCF_RESKEY_curl_retries_default="3"
+OCF_RESKEY_curl_sleep_default="1"
 
 : ${OCF_RESKEY_osccli=${OCF_RESKEY_osccli_default}}
 : ${OCF_RESKEY_profile=${OCF_RESKEY_profile_default}}
 : ${OCF_RESKEY_region=${OCF_RESKEY_region_default}}
+: ${OCF_RESKEY_curl_retries=${OCF_RESKEY_curl_retries_default}}
+: ${OCF_RESKEY_curl_sleep=${OCF_RESKEY_curl_sleep_default}}
 
 meta_data() {
     cat <<END
@@ -147,32 +151,6 @@ Expects to have a fully populated OCF RA-compliant environment set.
 END
 }
 
-curl_retry()
-{
-	local retries=$1 sleep=$2 opts=$3 url=$4
-	local tries=$(($retries + 1))
-	local args="--fail $opts $url"
-	local result rc
-
-	for try in $(seq $tries); do
-		ocf_log debug "curl $args try $try of $tries"
-		result=$(echo "$args" | xargs curl 2>&1)
-		rc=$?
-
-		ocf_log debug "result: $result"
-		[ $rc -eq 0 ] && break
-		sleep $sleep
-	done
-
-	if [ $rc -ne 0 ]; then
-		ocf_exit_reason "curl $args failed $tries tries"
-		exit $OCF_ERR_GENERIC
-	fi
-
-	echo "$result"
-	return $rc
-}
-
 osceip_start() {
     osceip_monitor && return $OCF_SUCCESS
 
@@ -220,6 +198,12 @@ osceip_validate() {
         return $OCF_ERR_GENERIC
     fi
 
+    if [ -z "${ELASTIC_IP}"] ; then
+        ocf_exit_reason "elastic_ip is a mandatory paramater. "
+        return $OCF_ERR_CONFIGURED
+    fi
+
+
     return $OCF_SUCCESS
 }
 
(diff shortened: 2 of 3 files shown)


heartbeat/osceip
```diff
@@ -0,0 +1,243 @@
...
+osceip_usage() {
+    cat <<END
+usage: $0 {start|stop|monitor|migrate_to|migrate_from|validate|validate-all|meta-data}
+
+Expects to have a fully populated OCF RA-compliant environment set.
+END
+}
+
+osceip_start() {
+    osceip_monitor || return $OCF_SUCCESS
```
oalbrigt, line 133: This should be `&&` to exit early if it's already started.

heartbeat/osceip
```diff
@@ -0,0 +1,243 @@
...
+    if [ $RET -ne 0 ]; then
+        return $OCF_NOT_RUNNING
+    fi
+
+    ocf_log info "elastic_ip has been successfully brought up (${ELASTIC_IP})"
+    return $OCF_SUCCESS
+}
+
+osceip_stop() {
+    osceip_monitor && return $OCF_SUCCESS
```
oalbrigt, line 148: This should be `||`.

heartbeat/osceip
```diff
@@ -0,0 +1,243 @@
...
+}
+
+osceip_start() {
+    osceip_monitor || return $OCF_SUCCESS
+
+    $oscCLI_CMD api LinkPublicIp --profile "${OCF_RESKEY_profile}" --PublicIp "${ELASTIC_IP}" --VmId "${INSTANCE_ID}" --AllowRelink True
+    RET=$?
+
+
+    if [ $RET -ne 0 ]; then
```
oalbrigt, line 139: You should add some more checks where osc-cli is run to return OCF_ERR_GENERIC or other error codes when osc-cli fails:
https://github.com/ClusterLabs/resource-agents/blob/main/doc/dev-guides/ra-dev-guide.asc#return-codes

heartbeat/osceip
```diff
@@ -0,0 +1,243 @@
...
+        meta_data
+        exit $OCF_SUCCESS
+        ;;
+    usage|help)
+        osceip_usage
+        exit $OCF_SUCCESS
+        ;;
+esac
+
+oscCLI_CMD="${OCF_RESKEY_osccli}"
```
oalbrigt, line 193: You can add --profile here as well (when it's set). And you could add a check validate() to fail if osc-cli requires it.

heartbeat/osceip
```diff
@@ -0,0 +1,233 @@
...
+if [ -n "${OCF_RESKEY_region}" ]; then
+	oscCLI_CMD="$oscCLI_CMD --region ${OCF_RESKEY_region}"
+fi
+ELASTIC_IP="${OCF_RESKEY_elastic_ip}"
+INSTANCE_ID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
```
oalbrigt, line 198: You should use curl_retry() here, and add the parameters for it, so it wont fail if there's a minor hickup in the network or API servers. See the aws-* agents for how to do that.

heartbeat/osceip
```diff
@@ -0,0 +1,233 @@
...
+
+case $__OCF_ACTION in
+    start)
+        osceip_validate
```
oalbrigt, line 202: Add ` || exit $?` to fail when it fails to validate.

oalbrigt (changes requested):
I've added some comments of what to improve, and you also need to add the agents in the Makefile.am files as mentioned here:
https://github.com/ClusterLabs/resource-agents/blob/main/doc/dev-guides/ra-dev-guide.asc#submitting-resource-agents

heartbeat/osceip
stephrobert: I have a strange behavior with curl_retry

./osceip: 222: curl_retry: not found
INSTANCE_ID=$(curl_retry "${OCF_RESKEY_curl_retries}" "${OCF_RESKEY_curl_sleep}" -s" "http://169.254.169.254/latest/meta-data/instance-id")

even if  i have :

oalbrigt: Ah. It has only been added recently:
https://github.com/ClusterLabs/resource-agents/commit/fc0657b936f6a58f741e33f851b22f82bc68bffa

You can set `OCF_ROOT=. OCF_FUNCTIONS_DIR=.` when trying to run it manually (after running `make`).

heartbeat/osceip
stephrobert: fixed

heartbeat/osceip
stephrobert: fixed

heartbeat/osceip
stephrobert: fixed

heartbeat/osceip
stephrobert: fixed

heartbeat/osceip
```diff
@@ -0,0 +1,277 @@
...
+
+osceip_usage() {
+    cat <<END
+usage: $0 {start|stop|monitor|migrate_to|migrate_from|validate|validate-all|meta-data}
+
+Expects to have a fully populated OCF RA-compliant environment set.
+END
+}
+
+curl_retry()
```
oalbrigt, line 150: We dont want curl_retry() here, as it's in ocf-shellfuncs.

stephrobert: Unable to use ocf-shellfunc's curl_retry (curl_retry not found). I tested the awseip agent which is the only agent using it and I have the same problem. This problem appears when using the agent and not in standalone mode. I am on an ubuntu 22.04

stephrobert: May 22 07:49:00.915 hap-b.test.local pacemaker-execd     [733] (log_op_output)  notice: elastic_stop_0[8386] error output [ /usr/lib/ocf/resource.d/heartbeat/osceip: 220: curl_retry: not found ]

heartbeat/osceip
stephrobert: profile is not mandatory. It has a default value

heartbeat/osceip
stephrobert: I add a check on elastic-ip parameter

heartbeat/osceip
oalbrigt: Ah, yeah. You'll have to patch /usr/lib/ocf/lib/heartbeat/ocf-shellfuncs in that case.

You should be able to do it by running the following commands:
```
cd  /usr/lib/ocf/lib/heartbeat
curl -LO https://github.com/ClusterLabs/resource-agents/pull/1936/commits/fc0657b936f6a58f741e33f851b22f82bc68bffa.patch
patch -p2 -F2 < fc0657b936f6a58f741e33f851b22f82bc68bffa.patch
rm fc0657b936f6a58f741e33f851b22f82bc68bffa.patch
```

stephrobert: Is it possible to install resource agents from source ?

oalbrigt: Yes, it is.

You might need to install some -dev packages for some libraries, which it will complain about are missing when you run ./configure:
```
./autogen.sh

# rerun after installing lib<name>-dev it was complaining about until it completes without errors
./configure
make
make install
```

stephrobert: fixed

heartbeat/Makefile.am
```diff
@@ -153,6 +153,7 @@ ocf_SCRIPTS	      = AoEtarget		\
 			oraasm			\
 			oracle			\
 			oralsnr			\
+			osceip			\
```
oalbrigt, line 156: You also need to add it to the Makefile.am in doc/man.

stephrobert: Done

heartbeat/osceip
```diff
@@ -0,0 +1,261 @@
...
+
+osceip_validate() {
+    check_binary "${OCF_RESKEY_osccli}"
+
+    if [ -z "${INSTANCE_ID}" ]; then
+        ocf_exit_reason "instance_id not found. Is this a vm instance?"
+        return $OCF_ERR_GENERIC
+    fi
+
+    if [ -z "${ELASTIC_IP}"] ; then
```
oalbrigt, line 201: You need a space before `]` here (detected by CI).

oalbrigt (approved):

stephrobert:
It's me who thanks you

Le mer. 22 mai 2024 à 14:11, Oyvind Albrigtsen ***@***.***> a
écrit :

PR #1940, opened 2024-05-27
Mid: storage-mon: Add logging to daemon mode.

HideoYamauchi:
Hi All,

Disk error logs that are output in normal mode are not output in daemon mode.
Modify it to output the same log as normal mode.

Best Regards,
Hideo Yamauchi.

### DIFF
--- a/tools/storage_mon.c
+++ b/tools/storage_mon.c
@@ -308,6 +308,8 @@ static int32_t sigchld_handler(int32_t sig, void *data)
 						/* so add the final_score from the exit code of the terminated child process. 	*/
 						if (qb_loop_timer_is_running(storage_mon_poll_handle, expire_handle)) { 
 							if (WEXITSTATUS(status) !=0) {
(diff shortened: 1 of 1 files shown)


tools/storage_mon.c
```diff
@@ -308,6 +308,8 @@ static int32_t sigchld_handler(int32_t sig, void *data)
 						/* so add the final_score from the exit code of the terminated child process. 	*/
 						if (qb_loop_timer_is_running(storage_mon_poll_handle, expire_handle)) { 
 							if (WEXITSTATUS(status) !=0) {
+								syslog(LOG_ERR, "Error reading from device %s", devices[index]);
```
oalbrigt, line 311: index is the child_pid in this case, and not the index of the devices array.

oalbrigt: Ah great. Just making sure it wasnt by accident.

oalbrigt:
Thanks.

PR #1949, opened 2024-06-13
Filesystem/pgsql: silence spurious WARNING and ERROR log messages

lge:
* pgsql: pgsql_status: silence ERROR: command failed: runuser
* Filesystem: for "probe", silence "Expected /dev/... to exist"

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -485,7 +485,7 @@ runasowner() {
             "-q")
                 quietrun="-q"
                 shift 1;;
-            "info"|"warn"|"err")
+            "info"|"warn"|"err"|"debug")
                 loglevel="-$1"
                 shift 1;;
             *)
@@ -920,7 +920,7 @@ pgsql_status() {
      if [ -f $PIDFILE ]
      then
          PID=`head -n 1 $PIDFILE`
-         runasowner "kill -s 0 $PID >/dev/null 2>&1"
+         runasowner -q debug "kill -s 0 $PID"
          return $?
      fi
 
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -1012,7 +1012,7 @@ set_blockdevice_var() {
 		blockdevice=yes
 		;;
 	*)
-		if [ ! -b "$DEVICE"  -a ! -d "$DEVICE" -a "X$OP" != Xstart ] ; then
+		if [ ! -b "$DEVICE"  -a ! -d "$DEVICE" -a "X$OP" != Xstart ] && ! ocf_is_probe ; then
 			ocf_log warn "Couldn't find device [$DEVICE]. Expected /dev/??? to exist"
 		fi
 		if [ ! -d "$DEVICE" ]; then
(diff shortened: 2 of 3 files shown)


heartbeat/pgsql
```diff
@@ -920,7 +920,10 @@ pgsql_status() {
      if [ -f $PIDFILE ]
      then
          PID=`head -n 1 $PIDFILE`
-         runasowner "kill -s 0 $PID >/dev/null 2>&1"
+         # "failure" to "ping" the pid may be "expected" during shutdown,
+         # and is not critical for a stale pid file.
+         # No need to have ocf_run log "ERROR command failed"
+         $SU $OCF_RESKEY_pgdba -c "kill -s 0 $PID >/dev/null 2>&1"
```
oalbrigt, line 926: `runasowner -q` should do this.

lge: afaics, the "-q" only silences the output of a _successful_ command.  
"failed" (exit code != 0) commands are always logged. That's why I drop the ocf_run.

oalbrigt: You're right. I think we could extend runasowner() and ocf_run() in ocf-shellfuncs() to support debug-level logging of errors.

lge: We certainly could, but that is unrelated.

For commands that are used in some `if` condition, exit code != 0 is not a "failure" to be noticed, but an expected outcome, and should not be logged at all. If you want that level, you can still ask for a "trace".

lge: in this specific case, in `pgsql_real_stop()`
```sh
    # Stop PostgreSQL, do not wait for clients to disconnect
    if [ $stop_escalate -gt 0 ]; then
            runasowner "$OCF_RESKEY_pgctl -W -D $OCF_RESKEY_pgdata stop -m fast"
    fi

    # stop waiting
    count=0
    while [ $count -lt $stop_escalate ]
    do
        if ! pgsql_status
        then
            #PostgreSQL stopped
            break;
        fi
        count=`expr $count + 1`
        sleep 1
    done
```

So for _EVERY_ regular stop, the last line logged will be
`ERROR command failed: runuser postgress ... kill -s 0 ...`

That is just ~stupid~ "bad logging discipline".

oalbrigt: Correct.

So if we log it to debug it will only get output when debugging is enabled, which could be helpful when trying to find why something fails.

oalbrigt: I've added debug-level logging now, so you can use `runasowner debug` to only log the error in debug-mode:
https://github.com/ClusterLabs/resource-agents/pull/1950

lge: Though it still is not an error. It is an expected result.

oalbrigt: It's fine to log in debug-mode though, and I also prefer to use these functions when they're already used to make the code simpler to change in the future.

lge:
Updated to include the first commit from #1950 and folded the second commit (pgsql runasowner debug support) into this one.

oalbrigt:
Thanks.

PR #1957, opened 2024-07-09
powervs-subnet: improvements on virtual server status collection, and subnet - port handling

ehaefele:
improvements on virtual server status collection, and subnet - port handling.


oalbrigt:
LGTM. Thanks.

PR #1965, opened 2024-08-14
findif.sh: ignore unreachable, blackhole, and prohibit routes

### LINKED ISSUE #1963: IPaddr2/findif: new "fail when matching more than 1 route" breaks in my environment
In my setup, the following Pacemaker resource (using crmsh syntax) breaks:

```
primitive p_vip_hafw_ip6 IPaddr2 \
        params ip="2001:db8::1" cidr_netmask=128 nic=lo \
        op monitor interval=15 timeout=100
```

I suspect what is triggering this breakage is a covering unreachable route for the prefix, e.g.:

```
unreachable 2001:db8::/64 dev lo proto bird metric 32 pref medium
```

My syslog contains messages such as:

```
2024-08-03T21:05:58.920320+01:00 hicks IPaddr2(p_vip_hafw_ip6)[856437]: ERROR: More than 1 routes match 2001:db8::1/128. Unable to decide which route to use.
2024-08-03T21:05:58.922592+01:00 hicks IPaddr2(p_vip_hafw_ip6)[856437]: WARNING: [findif] failed
```

Downgrading from 4.15.1 to 4.14.0 resolves the issue.

oalbrigt:
Fixes #1963.

### DIFF
--- a/heartbeat/findif.sh
+++ b/heartbeat/findif.sh
@@ -218,9 +218,9 @@ findif()
   fi
   if [ -n "$nic" ] ; then
     # NIC supports more than two.
-    routematch=$(ip -o -f $family route list match $match $proto $scope | grep "dev $nic " | sed -e 's,^\([0-9.]\+\) ,\1/32 ,;s,^\([0-9a-f:]\+\) ,\1/128 ,' | sort -t/ -k2,2nr)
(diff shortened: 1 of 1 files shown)


bootc:
If you're doing it this way I think you'll also want to ignore `prohibit` and `blackhole` routes.

I'm not entirely clear why this code does a `ip route list match` rather than simply a `ip route get` to get the route that the kernel would use. Would that not be better? Yes it works differently and it's not necessarily obvious if a route uses the default route in this case, but does it matter for this usage?

oalbrigt:
We can ignore those as well. I dont know the reason why it was used in the first place, but it's what has been used forever.

PR #1969, opened 2024-09-01
Add nic detect for IPaddr and IPaddr2

twtlpla:
If the network adapter is faulty (down or otherwise), the IP address cannot be accessed even if the ip resource is working properly, so the network adapter detection is added


oalbrigt:
You should use the ethmonitor agent for advanced monitoring of the NICs.

PR #1970, opened 2024-09-04
powervs-subnet: Enable access via private endpoint for IBM IAM

ehaefele:
Private networks in IBM PowerVS now can communicate with private endpoints without a proxy.
Allow a private endpoint also for IBM IAM access.


oalbrigt:
Thanks.

PR #1971, opened 2024-09-04
apache/http-mon.sh: change curl opts to match wget

athos-ribeiro:
curl was recently moved to be de default http client, deprecating wget, the former preferred one. However, the curl options being used differ from the ones set for wget, changing the behavior of the agents using the monitor when using one or the other client.


oalbrigt:
Thanks.

PR #1977, opened 2024-09-18
Filesystem fix stop try umount first

lge:
See individual commit messages.

* fix stop: try umount first, and only if that fails try to find users of that file system
* in the "safe" mode of get_pids(), avoid a potential "Argument list too long" problem

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -732,6 +732,11 @@ fs_stop() {
 	local SUB="$1" timeout=$2 grace_time ret
 	grace_time=$((timeout/2))
 
+	# Just walking /proc may take "a long time", even if we don't find any users of this FS.
+	# If dependencies are properly configured, umount should just work.
+	# Only if that fails, try to find and kill processes that still use it.
+	try_umount "" "$SUB" && return $OCF_SUCCESS
+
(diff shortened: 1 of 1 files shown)


oalbrigt (approved):
LGTM.

SatomiOSAWA:
Hi, @lge @oalbrigt 

Impressive work!
Thank you for your consideration.
I understand that the issue with [RHBZ:2207567](https://bugzilla.redhat.com/show_bug.cgi?id=2207567) is serious.
However, as Mr. Lars noted, if the dependencies are properly configured,
it seems unlikely that unmounting first would cause any problems.
I have come to recognize the importance of adding proper comments once again.
I believe many users will appreciate this improvement.

oalbrigt:
Thanks.

oalbrigt:
Feel free to make a patch when you got the time.

We should keep it non-bash specific, as the agent can also be used on *BSD.

lge:
> Feel free to make a patch when you got the time.
> 
> We should keep it non-bash specific, as the agent can also be used on *BSD.

https://github.com/ClusterLabs/resource-agents/pull/1978 (hope that is sufficiently *BSD compatible ...)

PR #1980, opened 2024-09-20
IPaddr2: add proto-parameter to be able to match a specific route

### LINKED ISSUE #1976: IPaddr2/findif:  more than 1 matching routes error
Similar to [#1963], I'm facing a problem using IPv6 addresses with IPAddr2 since the 4.15 release of Resource Agents due to error 'More than 1 routes match nn:nn:nn:nn/128. Unable to decide which route to use.'

In my case, the issue is different because there really are multiple routes that match:

```
ip -o -f inet6 route list match 2222:3333:4444::a00:8/128 | grep -v "^\(unreachable\|prohibit\|blackhole\)" | grep "dev br0 " | sed -e 's,^\([0-9.]\+\) ,\1/32 ,;s,^\([0-9a-f:]\+\) ,\1/128 ,' | sort -t/ -k2,2nr | grep -v "^default"
2222:3333:4444::/64 dev br0 proto kernel metric 256 pref medium
2222:3333:4444::/64 dev br0 proto ra metric 1024 expires 6870sec pref medium
2222:3333:4444::/48 via fe80::b2f2:8ff:fedf:bb27 dev br0 proto ra metric 1024 expires 1470sec pref medium
```

Two of these routes are learnt via Router Announcements with different network masks (/48 and /64), one is 'kernel' assigned. Based on the routes in the routing table, I can see why findif complains that there are 3 matching routes.

However, from the commit notes, I can't see the relevance of this check, which was not part of the prior 4.14 release. If I comment out the relevant section of code as follows, 'findif' continues without any issue and the IP address is added to the correct interface.

```
#  if [ $(echo "$routematch" | wc -l) -gt 1 ]; then
#    ocf_exit_reason "More than 1 routes match $match. Unable to decide which route to use."
#    return $OCF_ERR_GENERIC
#  fi
```

It's possible I'm adding IPv6 addresses incorrectly; I've seen other examples where they are added to the loopback interface. This 'works' though the IP address is then inaccessible across the network so I'm likely misunderstanding how this should work.

Other than hacking out this check, which is presumably there for a reason, any suggestions on how I can should use IPaddr2 to add an IPv6 address to a Pacemaker clusters?

Thanks.

oalbrigt:
Fixes #1976

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -72,8 +72,8 @@
 OCF_RESKEY_ip_default=""
 OCF_RESKEY_cidr_netmask_default=""
 OCF_RESKEY_broadcast_default=""
+OCF_RESKEY_proto_default=""
 OCF_RESKEY_iflabel_default=""
-OCF_RESKEY_cidr_netmask_default=""
 OCF_RESKEY_lvs_support_default=false
 OCF_RESKEY_lvs_ipv6_addrlabel_default=true
 OCF_RESKEY_lvs_ipv6_addrlabel_value_default=99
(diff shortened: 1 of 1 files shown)


oalbrigt:
retest this please

HideoYamauchi:
Hi Oyvind,

I think this modification adds a proto parameter and attempts to process it with findif().
However, since findif()'s proto is a local parameter, I don't think you'll be able to pass the proto parameter successfully with this modification.

Best Regards,
Hideo Yamauchi.

oalbrigt:
@HideoYamauchi It is being used, as findif.sh is sourced from IPaddr2: https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/IPaddr2#L69

HideoYamauchi:
Hi @oalbrigt 

> @HideoYamauchi It is being used, as findif.sh is sourced from IPaddr2: https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/IPaddr2#L69

I understand that too.
But the proto variable of findif() is local, so the proto assignment added in this fix is ​​not passed to findif(), is it?

```
findif()
{
  local match="$OCF_RESKEY_ip"
  local family
  local proto
  (snip)
```

Best Regards,
Hideo Yamauchi.

oalbrigt:
Nice catch.

Patch to fix it: https://github.com/ClusterLabs/resource-agents/pull/1981

PR #1982, opened 2024-10-07
IPaddr2: add a check to make sure the label got applied

karcaw:
We have found an instance where sometimes on Rocky8 the ipv4 label does not get set on a specific IP, and we needed to add this check to make sure the label is set properly before other services start.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -1003,11 +1003,11 @@ ip_served() {
 
 	if [ -z "$IP_CIP" ]; then
 		for i in $cur_nic; do
-                        # check address label
-                        if [ ! -z $IFLABEL ] && [ -z "$IP2UTIL -o -f $FAMILY addr show $nic label $IFLABEL" ]; then
-                           echo partial3
-                           return 0
-                        fi
+			# check address label
+			if [ ! -z $IFLABEL ] && [ -z "$IP2UTIL -o -f $FAMILY addr show $nic label $IFLABEL" ]; then
+				echo partial3
+				return 0
+			fi
 			# only mark as served when on the same interfaces as $NIC
 			[ "$i" = "$NIC" ] || continue
 			echo "ok"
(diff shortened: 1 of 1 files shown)


heartbeat/IPaddr2
```diff
@@ -1002,6 +1003,11 @@ ip_served() {
 
 	if [ -z "$IP_CIP" ]; then
 		for i in $cur_nic; do
+                        # check address label
+                        if [ ! -z $IFLABEL -a -z "`$IP2UTIL -o -f $FAMILY addr show $nic label $IFLABEL`" ]; then
```
oalbrigt, line 1007: We avoid -a/-o where possible: https://web.archive.org/web/20230306130858/https://wiki.bash-hackers.org/commands/classictest#why_you_should_avoid_using_-a_and_-o

Can you update it to `[ ! -z $IFLABEL ] && [ -z "`$IP2UTIL -o -f $FAMILY addr show $nic label $IFLABEL`" ]`?

karcaw: Done!

heartbeat/IPaddr2
```diff
@@ -1002,6 +1003,11 @@ ip_served() {
 
 	if [ -z "$IP_CIP" ]; then
 		for i in $cur_nic; do
+                        # check address label
```
oalbrigt, line 1006: I just noticed you're using spaces for indentation, so that needs to be updated to tabs to follow the rest of the code.

karcaw: done

heartbeat/IPaddr2
oalbrigt: Oops. I shouldnt have used single backquotes for formatting, which made them cancelled out in the comment.

Can you add the backquotes or $( ) around the command again?

oalbrigt:
Do you have a reproducer for us to be able to test this issue?

heartbeat/IPaddr2
oalbrigt: We ended up with some additional improvements based on your fix, so no need to update this PR.

I'll just make a new PR, and give you credit there.

oalbrigt:
Closing in favor of https://github.com/ClusterLabs/resource-agents/pull/1985 which has additional fixes to fail during start-action to avoid having to wait for the first monitor-action to fail.

PR #1983, opened 2024-10-08
ocf-shellfuncs: systemd_drop_in only if needed

tacerus:
Avoid dbus overload upon many simultaneous "daemon-reload" invocations (when a resource agent using systemd_drop_in() is called multiple times as part of parallel resource operations in Pacemaker) by skipping the file creation and reload if the expected data already exists.

Whilst at it, align the indentation of the heredoc with the other parts of the function.


oalbrigt:
Thanks.

PR #1985, opened 2024-10-11
IPaddr2: improve fail logic and check ip_status after adding IP

oalbrigt:
* check that the label got applied
* return OCF_ERR_GENERIC to avoid false-positive when IP was manually added before starting the resource
* check ip_status after adding IP to fail without having to wait for the first monitor-action

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -586,7 +586,7 @@ ip_init() {
 			exit $rc
 		fi
 	fi
-	
+
 	SENDARPPIDFILE="$SENDARPPIDDIR/send_arp-$OCF_RESKEY_ip"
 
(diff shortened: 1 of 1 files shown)


oalbrigt:
Based on https://github.com/ClusterLabs/resource-agents/pull/1982

karcaw:
the best way we have to reproduce this is to have one label and IP on an Infiniband ipoib interface, and then try to add a second IP and label on the same interface.  the second one seems to not get a label at times.  This does not happen on our ethernet interfaces.  this is in an active/active failover pair, and so we have two IPs and in a failover situation, we just don't get the label.  a dependent resource is looking for that label for binding and fails.

oalbrigt:
Thanks for the instructions on how to reproduce the issue.

PR #1988, opened 2024-10-28
Filesystem: add new azure filesystem type "azfns"

HappyTobi:
Add new filesystem type for Azure.

 - New azure filesystem type: ["azfns"](https://github.com/Azure/AZNFS-mount)

The new implementation enables the mounting of the Azure 'azfns' filesystem type with the right tools and type.


oalbrigt:
Thanks.

PR #1999, opened 2024-11-26
spec: fix sbin dependencies for ELN

yselkowitz:
Like in Fedora 41, ELN (the future RHEL 11) also defaults to dnf5 and therefore only can use /usr/bin and /usr/sbin file dependencies.

https://src.fedoraproject.org/rpms/resource-agents/pull-request/7
https://src.fedoraproject.org/rpms/resource-agents/pull-request/9


oalbrigt:
Duplicate of https://github.com/ClusterLabs/resource-agents/pull/1998.

PR #2006, opened 2025-01-12
ocf-shellfuncs: fixed bash systax error added at a25f08

sergey-safarov:
ocf-shellfuncs: fixed bash systax error added at a25f08cf98d784894df9c52960eff5ccef059393


oalbrigt:
Thanks.

PR #2007, opened 2025-01-15
nfsserver: stop nfsdcld after rpc-gssd due to issue in Debian 12

oalbrigt:
Tested on RHEL8/RHEL9 without any issues with and without nfsv4_only.


oalbrigt:
The issue is caused by the agent using sysv init when systemd is being used on the system.

Issue #2003, opened 2024-12-24
Some resource agents still use egrep/fgrep.


SatomiOSAWA:
Hi all!

I noticed that egrep command is issuing warnings when trying to create a ping resource on RHEL 10.0 beta, 
and I have submitted a PR to address it. 
https://github.com/ClusterLabs/pacemaker/pull/3775

It seems that there are some RAs and test tools in resource-agents which use egrep/fgrep as well.
```
[root@rh10-b03 resource-agents]# find . -name '*' -type f | xargs grep -e 'egrep' -e 'fgrep'
./configure.ac:AC_PATH_PROGS(EGREP, egrep)
./configure.ac:if export -p | fgrep " CFLAGS=" > /dev/null; then
./heartbeat/IPaddr2:    if egrep -q "(^|,)${IP_INC_NO}(,|$)" $IP_CIP_FILE ; then
./heartbeat/ManageVE.in:`echo "$OCF_RESKEY_veid" | egrep -q '^[[:digit:]]+$'`
./heartbeat/SAPInstance:      if [ `echo "$SERVICE" | egrep -c "$SEARCH"` -eq 1 ]
./heartbeat/VirtualDomain:              elif echo "$migrate_opts" | fgrep -qs -- "--migrateuri="; then
./heartbeat/VirtualDomain:      DOMAIN_NAME=`egrep '[[:space:]]*<name>.*</name>[[:space:]]*$' ${OCF_RESKEY_config} 2>/dev/null | sed -e 's/[[:space:]]*<name>\(.*\)<\/name>[[:space:]]*$/\1/'`
./heartbeat/WAS:  netstat -ltnp  2>/dev/null| egrep -i "($pat) .*LISTEN" | sed 's%.*LISTEN *%%'
./heartbeat/WAS6:  netstat -ltnp  2>/dev/null| egrep -i "($pat) .*LISTEN" | sed 's%.*LISTEN *%%'
./heartbeat/docker:     docker images | awk '{print $1 ":" $2}' | egrep -q -s "^(docker.io\/|${SERVER_NAME}\/)?${IMAGE_NAME}:${IMAGE_TAG}\$"
./heartbeat/eDir88.in:        RET=$($NDSTRACE --config-file "$NDSCONF" -c modules | egrep -i '^vrdim.*Running' | awk '{print $1}')
./heartbeat/mysql-proxy:    echo $log_level | egrep -q "^(error|warning|info|message|debug|)$"
[...]

PR #2011, opened 2025-01-21
All agents: use detected grep -E/-F instead of egrep/fgrep when detected

oalbrigt:
Fixes https://github.com/ClusterLabs/resource-agents/issues/2003.


SatomiOSAWA:
Thank you so much for fixing them all!

PR #2016, opened 2025-01-23
storage-mon: replace dashes with underscores in functions

oalbrigt:
Dashes in function names produce "`storage-mon_usage': not a valid identifier" error when run with sh -x.


HideoYamauchi:
It's an old problem.
Thanks for discovering it.

oalbrigt:
Yeah. It's from the initial commit for the agent I think, but luckily it doesnt seem to have caused any issues.

PR #2017, opened 2025-01-24
storage-mon: check if daemon is already running during start-action

oalbrigt:


HideoYamauchi:
Hi @oalbrigt 

This seems like a correct fix.

It might be more helpful to output something like "storage_mon is already running. PID=xxx" iby ocf_log.

Best Regards,
Hideo Yamauchi.

PR #2018, opened 2025-01-27
storage-mon: log "storage_mon is already running" in start-action

oalbrigt:
As suggested by @HideoYamauchi in https://github.com/ClusterLabs/resource-agents/pull/2017.


HideoYamauchi:
@oalbrigt 

Thanks!

PR #2024, opened 2025-02-20
portblock: use ocf_log() for logging

oalbrigt:


lge:
Just for the record: the `echo` statemenst in the `Say*()` functions where not "logging", but compatibility to the "heartbeat" resource agent conventions, which expected "OK" or "running" on stdout for "status" requests (and, fun fact, had to learn the hard way that the "not" in "not.*running" or "not.*OK" has to be considered as well ;-) yes I'm around long enough to remember that...).
That's why they are worded the way they are.

I'm totally find to convert those to logging statements, as has been done here.

oalbrigt:
Nice. Thanks for the info.

PR #2030, opened 2025-03-29
azure-events-az: change variable name allNodesStopped to allResourcesNowStopped

susegrant:
Same as c45ecb0 but with minor change to word order in log about nodes with resources still running.


susegrant:
This commit also changes the variable name allNodesStopped to allResourcesNowStopped which was the main change I submitted for c45ecb0.  I noticed the minor grammatical error in the log after that original submission and decided to include it with this change.  The original issue is that the code incorrectly refers to resources as nodes in the variable name allNodesStopped, which holds a boolean value from a test of whether all resources on a particular node are in stopped state.

The log message correction is the less important change with this commit.

Can this please be renamed again to "azure-events-az: Change variable name allNodesStopped to allResourcesNowStopped"?

oalbrigt:
Done.

oalbrigt:
Thanks.

PR #2033, opened 2025-04-03
azure-events-az: use attr_healthstate variable in crm_attribute call in putNodeOnline()

susegrant:
Change putNodeOnline block to use value in previously defined attr_healthstate variable instead of "#health-azure" string literal, making it consistent with the earlier putNodeStandby block


oalbrigt:
Thanks.

PR #2034, opened 2025-04-03
azure-events-az: Add params defaults to RA description

susegrant:
Set params explicitly in RA description using default values to give users a sane starting point.


oalbrigt:
Thanks.

PR #2035, opened 2025-04-04
azure-events-az: Remove # from start of attr_healthstate attribute name

susegrant:
Fix to improve visibility of azure health and increment version number.  See commit 2d52958 for more details.


oalbrigt:
The # is supposed to be there:
https://clusterlabs.org/projects/pacemaker/doc/2.1/Pacemaker_Explained/singlehtml/#node-health-attributes

Issue #1800, opened 2022-09-05
Resource agent "mdraid" need "md_dev=auto" setting.


mlaguilharre:
We have thousands resource agent Raid1 that contains each several soft raid device. They are managed thanks to the parameters raidconf="/etc/mdadm.conf" raiddev=auto. The /etc/mdamd.conf contains several entries for each soft raid device:

eg:
```
# /etc/mdadm.conf
ARRAY /dev/md/test-system metadata=1.2 name=node1:test-system UUID=ba616f8a:78da4bee:618dbdf6:cf9889f1
ARRAY /dev/md/test-data1 metadata=1.2 name=node1:test-data001 UUID=b805c2c2:b88442c3:b23bfd5f:b052b45b
ARRAY /dev/md/test-data2 metadata=1.2 name=:node1:test-data2 UUID=13fc4c7a:97067415:85dca6fe:20cc2f8

```
With latest pacemaker version, "Raid1" Resource Agent is obsoleted by "mdraid", unfortunately there is no equivalent parameter in "mdraid" for raiddev=auto.

If we need to migrate to mdraid, it needs to create a Resource Agent for each Soft Raid device which is cumbersome.

zhaohem:
In my view, the 'auto' function should be supported in mdraid RA.

oalbrigt:
I think that's a per distro decision. Pacemaker should support both Raid1 and mdadm.

I've asked the dev of the agent whether we intend to add this feature to mdadm.

oalbrigt:
There's no plans to add this feature to the mdraid agent.

zhaohem:
Thanks for the info. let's keep opening this issue. If I have time, I will submit a patch. I estimate the time is some months later.

Damenly:
@oalbrigt Hi, I've implemented a version to solve the issue. Would you help review the PR?

PR #2037, opened 2025-04-07
mdraid: Add 'auto' and multiple devices support to parameter md_dev

Damenly:
This commit allows the value 'auto' or multiple devices separated by space are set in parameter md_dev.

When 'auto' is assigned, mdraid RA will try to get all md arrays from mdraid_conf then assemble them automatically.
If multiple devices are assigned e.g. '/dev/md2 /dev/md3', the RA will try to assemble them.

The single device syntax is same as the original.

To keep the code and logic simple, global varaiable RAIDDEVS is introduced to record md devices. All entry functions mdraid_(stop|start|mintor) are renamed to mdraid_(stop|start|mintor)_one which are called by new function forall(). In forall(), md_dev and devs are used by mdraid_(stop|start|mintor)_one during iterating RAIDDEVS.
The original mdraid_stop_one is renamed to mdraid_stop_array for naming consistency.

Related: https://github.com/ClusterLabs/resource-agents/issues/1800

### DIFF
--- a/heartbeat/mdraid
+++ b/heartbeat/mdraid
@@ -96,6 +96,8 @@ preferably be a clustered raid1 or raid10 array created
 with --bitmap=clustered, assuming its resource will
 be cloned (i.e., active-active access).
 
+One or more block devices to use, space separated. Alternatively,
+set to "auto" to manage all devices specified in mdadm_conf.
 Be sure to disable auto-assembly for the resource-managed arrays!
 </longdesc>
 <shortdesc lang="en">MD block device</shortdesc>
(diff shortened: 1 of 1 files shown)


Damenly:
Hmmm. It seems Jenkins is broken.

heartbeat/mdraid
```diff
@@ -163,6 +165,37 @@ resource_is_cloned() {
...
+
+	for mddev in $RAIDDEVS; do
+		# prepare md_dev and devs
+		md_dev="$mddev"
+
+		# Required by start|stop|monitor
+		devs="`list_devices_for_mddev`"
+		if [ `echo $devs|wc -l` -eq 0 ]; then
+			ocf_exit_reason "No component device(s) found for MD RAID array $md_dev"
+			rc=$(($rc | $OCF_ERR_CONFIGURED))
```
oalbrigt, line 188: This should be `$OCF_ERR_CONFIGURED`, so Pacemaker can react accordingly.

Damenly: Got it. As OCF_ERR_CONFIGURED is a fatal error, do we need just return here? Or let the loop continue?

Damenly: Okay. Just pushed with the update.

oalbrigt:
We might need some more checks, so I will leave this open for 1-2 weeks to let the developer of the agent review it.

Damenly:
Hi. Any comment from the developers of the agent?

varkoly:
> Hi. Any comment from the developers of the agent?

Sure.  The change is pretty straightforward, and I've tested the following cases based on SLE 16 Beta3:

1. Set md_dev to 'auto'
1.1  If some of the md arrays listed in $mdadm_conf are not cluster-md, unless force_clones is true, the start of the RA will return $OCF_ERR_CONFIGURED
1.2 If all of the md arrays are in cluster-md type, they will be assembled one by one
1.3 If some of the md arrays listed in $mdadm_conf are not able to be assembled, $OCF_ERR_GENERIC will be returned, like how single device parameter behaves
2. Set md_dev to multi deivces like '/dev/md3 /dev/md127', mostly like 'auto' is given :
2.1 If some of these md arrays listed in $mdadm_conf are not cluster-md, unless force_clones is true, the start of the RA will return $OCF_ERR_CONFIGURED
2.2 If all of these md arrays are in cluster-md type, they will be assembled one by one
2.3 If some of the md arrays listed in $mdadm_conf are not able to be assembled, $OCF_ERR_GENERIC will be returned, like how single device parameter behaves
2.4 If some of given md devices are not in $mdadm_conf, the RA will fail to start as $OCF_ERR_GENERIC.
3. Set md_dev to single device. It works as before the change.
4. Set md_dev to other meaningless strings, the start returns $OCF_ERR_ARGS as the original.

mauelsha:
Ok with me, as it's optionally configurable (keyword 'auto') and doesn't change single array RA semantics.

oalbrigt:
retest this please

Damenly:
Ping.  Can this PR be merged?

PR #2044, opened 2025-05-07
azure-events-az: Update wording in log line

susegrant:
Update log line to improve semantic accuracy.  The word "clean" is used incorrectly to describe whether all resources have been "stopped" on the node.


oalbrigt:
Thanks.

PR #2049, opened 2025-05-20
podman-etcd: fix listen-peer-urls binding

clobrano:
This change ensures learner etcd listens on all interfaces for peer connections, resolving accessibility issues.

Fix: OCPBUGS-56447


oalbrigt:
retest this please

PR #2054, opened 2025-06-06
Allow cloned crypt primitives

ryan-ronnander:
When using DRBD (a cloned resource) on top of encrypted luks volumes, there is a need to have the underlying luks crypt resources cloned as well.


oalbrigt:
The crypt agent must not allow any cloned resource because the dm-crypt target is not cluster-aware hence causing data corruption.

PR #2061, opened 2025-07-18
WIP: Allow for control-plane node replacement in podman-etcd

jaypoulz:
This is a collection of tweaks intended to adjust the start-up properties of podman-etcd to allow for control-plane node replacement.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1078,9 +1078,23 @@ compare_revision()
 	revision=$(attribute_node_revision get)
 	peer_revision=$(attribute_node_revision_peer)
 
-	if [ "$revision" = "" ] || [ "$revision" = "null" ] || [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
+	if [ "$revision" = "" ] || [ "$revision" = "null" ] && [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
 		ocf_log err "could not compare revisions: $NODENAME local revision: $revision, peer revision: $peer_revision"
 		return "$OCF_ERR_GENERIC"
+	fi
+
+	# When only the local revision exists, it can start normally. This ensures that it can start during node replacement events.
+	if [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
+		ocf_log info "only local revision exists: $NODENAME local revision: $revision, peer revision: $peer_revision"
+		echo "only_local"
+		return "$OCF_SUCCESS"
+	fi
+
+	# When only the peer revision exists, the local node must join as a learner. This ensures that it can start during node replacement events.
+	if [ "$revision" = "" ] || [ "$revision" = "null" ]; then
+		ocf_log info "only peer revision exists: $NODENAME local revision: $revision, peer revision: $peer_revision"
+		echo "only_peer"
+		return "$OCF_SUCCESS"
 	fi
 
 	if [ "$revision" -gt "$peer_revision" ]; then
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -1078,11 +1078,25 @@ compare_revision()
 	revision=$(attribute_node_revision get)
 	peer_revision=$(attribute_node_revision_peer)
 
-	if [ "$revision" = "" ] || [ "$revision" = "null" ] || [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
+	if [ "$revision" = "" ] || [ "$revision" = "null" ] && [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
 		ocf_log err "could not compare revisions: $NODENAME local revision: $revision, peer revision: $peer_revision"
 		return "$OCF_ERR_GENERIC"
 	fi
 
+	# When only the local revision exists, it can start normally. This ensures that it can start during node replacement events.
```
clobrano, line 1086: I think this is too risky, as it be part of any normal startup, not only when an admin operates a node replacement.
There is an alternative already, the admin can set the property "force_new_cluster" on the node will ignore the revisions.

heartbeat/podman-etcd
```diff
@@ -1078,11 +1078,25 @@ compare_revision()
...
+	if [ "$peer_revision" = "" ] || [ "$peer_revision" = "null" ]; then
+		ocf_log info "only local revision exists: $NODENAME local revision: $revision, peer revision: $peer_revision"
+		echo "only_local"
+		return "$OCF_SUCCESS"
+	fi
+
+	# When only the peer revision exists, the local node must join as a learner. This ensures that it can start during node replacement events.
```
clobrano, line 1093: This is less risky, as the local revision comes directly from disk, so this might occur only if the node has no etcd data at all.

heartbeat/podman-etcd
```diff
@@ -1194,29 +1208,31 @@ podman_start()
...
+					if [ "$revision_compare_result" != "older" ] || [ "$revision_compare_result" != "only_peer" ]; then
 						# If our revision is the same as or newer than the peer's last saved
 						# revision, and the peer agent isn't currently starting, we can
 						# restore e-quorum by forcing a new cluster.
 						set_force_new_cluster
 					else
-						ocf_log err "local revision is older and peer is not starting: cannot start"
-						ocf_exit_reason "local revision is older and peer is not starting: cannot start"
+						ocf_log err "local revision is older or empty and peer is not starting: cannot start"
+						ocf_exit_reason "local revision is older or emptyand peer is not starting: cannot start"
```
clobrano, line 1218: Missing a space between "and"

heartbeat/podman-etcd
```diff
@@ -1194,29 +1208,31 @@ podman_start()
...
 					if [ "$revision_compare_result" = "newer" ]; then
 						set_force_new_cluster
-					elif [ "$revision_compare_result" = "older" ]; then
+					elif [ "$revision_compare_result" = "older" ] || [ "$revision_compare_result" = "only_peer" ]; then
 						ocf_log info "$NODENAME shall join as learner"
 						JOIN_AS_LEARNER=true
 					else
 						if [ "$(attribute_node_cluster_id get)" = "$(attribute_node_cluster_id_peer)" ]; then
 							ocf_log info "same cluster_id and revision: start normal"
+						else if [ "$revision_compare_result" = "only_local" ] && [ "$(attribute_node_cluster_id get)" != ""] && [ "$(attribute_node_cluster_id get)" != "null" ]; then
```
clobrano, line 1232: A couple of minor notes on this
* you can use elif here
* there's a missing space before a closing bracket "]"

Moreover, this seems wrong to me. We have 2 starting resources with different cluster-ids, they cannot start normally.

clobrano (changes requested):

PR #2062, opened 2025-07-23
[WIP] OCPBUGS-57456: podman-etcd should keep the container for crash debugging

clobrano:
This change modifies the `podman-etcd` resource agent to conditionally reuse existing containers, preventing their removal and preserving historical logs.

As preliminary change, it replaces inline configuration arguments with external files:
* environment variables passed to podman via file /var/lib/etcd/podman-env.yaml
* etcd configuration variable passed via file /var/lib/etcd/etcd-config.yaml

This allows configuration changes to be applied via container restart rather than requiring full container recreation.

The agent can then
* Compare Cluster Etcd Operator (CEO) `etcd-pod.yaml` manifest to decide whether to reuse an existing container or create a new one.
* Retain containers after stops to ensure logs are always available for debugging.


clobrano:
The existing container is deleted right before starting a new one. I could improve it to keep the last one only

clobrano:
As this PR also introduces configuration files for podman (env.yaml) and etcd (config.yaml), I also need to backup those files together with the podman container, otherwise we will be unable to check what data etcd is started with.


/hold

heartbeat/podman-etcd
```diff
@@ -453,6 +478,67 @@ prepare_env() {
...
+
+generate_etcd_configuration() {
+	if is_force_new_cluster; then
+		#OCF_RESKEY_run_cmd="$OCF_RESKEY_run_cmd --force-new-cluster"
+		FORCE_NEW_CLUSTER=true
+	else
+		FORCE_NEW_CLUSTER=false
+	fi
+
+	cat > "$OCF_RESKEY_podman_env_file" << EOF
```
jaypoulz: For the collection of commented-out options, are these things we're leaving here because we expect to enable them down the line?

jaypoulz (approved):
I really like the new file-based configuration options. It think it makes it easier to follow along to where the configuration options are being sourced and propagated. Excited to also have these logs preserved.

heartbeat/podman-etcd
clobrano: Those were leftovers I forgot to remove. However I noticed that some env variables (e.g. etcd_data) were not correctly used, and, thinking more about it, the env file for podman is not necessary anyway, as the only thing that can really change is the etcd command line. This to say that I need to push a version without podman env file :)

heartbeat/podman-etcd
```diff
@@ -532,7 +564,19 @@ listen-peer-urls: "https://${LISTEN_PEER_URLS}:2380"
...
 experimental-initial-corrupt-check: true
+experimental-max-learners: 1
+etcd-socket-reuse-address: $ETCD_SOCKET_REUSE_ADDRESS
+experimental-warning-apply-duration: $(convert_duration_in_nanoseconds "$ETCD_EXPERIMENTAL_WARNING_APPLY_DURATION")
+experimental-watch-progress-notify-interval: $(convert_duration_in_nanoseconds "$ETCD_EXPERIMENTAL_WATCH_PROGRESS_NOTIFY_INTERVAL")
 EOF
+
+	{
+		echo "cipher-suites:"
+		IFS=',' read -ra cipher_array <<< "$ETCD_CIPHER_SUITES"
```
oalbrigt: `<<<` is bash-specific, so you should avoid it if you can. If not we'll have to rename it to .in, update the hashbang to `#!@BASH_SHELL@` and add it to "configure.ac" to auto-update.

clobrano: oh, I'll try to find an alternative

heartbeat/podman-etcd
```diff
@@ -309,42 +312,71 @@ container_exists()
...
+	fi
+
+	# Archive current container
+	if ocf_is_true "$is_archive_container_name_available"; then
+		ocf_log info "attempting to archive '$CONTAINER' container as '$CONTAINER-previous' for debugging purpose"
+		if ocf_run podman rename "$CONTAINER" "$CONTAINER-previous"; then
+			# Archive corresponding etcd configuration file
+			etcd_configuration_file_previous=${OCF_RESKEY_etcd_configuration_file//".yaml"/"-previous.yaml"}
+			if ocf_run cp "$OCF_RESKEY_etcd_configuration_file" "$etcd_configuration_file_previous"; then
+				return "$OCF_SUCCESS"
```
jaypoulz: I'm confused about this return - aren't we also planning to archive the old logs? Is it sufficient to just have the old yaml?

jaypoulz: Also, don't we need to remove the container still?

heartbeat/podman-etcd
```diff
@@ -453,6 +521,64 @@ prepare_env() {
 	LISTEN_METRICS_URLS="0.0.0.0"
 }
 
+
+generate_etcd_configuration() {
+	if is_force_new_cluster; then
+		# The embedded newline is required for correct YAML formatting.
+		FORCE_NEW_CLUSTER_CONFIG="force-new-cluster: true
+force-new-cluster-bump-amount: 1000000000"
```
jaypoulz, line 563: what is this for?

heartbeat/podman-etcd
```diff
@@ -1119,6 +1245,58 @@ ensure_pod_manifest_exists()
...
+	fi
+
+	# Decide whether to reuse the existing container or create a new one based
+	# on etcd pod manifest changes explicitly ignoring the revision, as the content
+	# might be the same even if the revision number has changed.
+
+	# NOTE: the pod manifest contains some values referred to its REVISION that we want to ignore
+	jq_filter='del(.metadata.labels.revision) | .spec.containers[] |= ( .env |= map(select( .name != "ETCD_STATIC_POD_VERSION" ))) | .spec.volumes |= map( select( .name != "resource-dir" ))'
+
+	ocf_run diff -s <(jq "$jq_filter" "$OCF_RESKEY_pod_manifest") <(jq "$jq_filter" "$OCF_RESKEY_pod_manifest_copy")
```
jaypoulz: does this print out the diff so we can see it in the runtime log?

heartbeat/podman-etcd
```diff
@@ -1206,6 +1393,7 @@ podman_start()
 					fi
 					;;
 				2)
+					# TODO: can we start "normally", regardless the revisions, if the container-id is the same on both nodes?
```
jaypoulz, line 1473: If the revision we're referring to is the revision of the static pod yaml generated by CEO, I'm pretty sure the answer is yes.

jaypoulz (approved):
/lgtm

heartbeat/podman-etcd
clobrano: yes

heartbeat/podman-etcd
clobrano: I agree, but since this was a different story, I decided it was best to at least leave a message for the future :)

heartbeat/podman-etcd
clobrano: Which logs? Aren't the logs inside the container?

heartbeat/podman-etcd
clobrano: This is something we left out initially. It's part of the restore-quorum script in CEO. The script is more advanced, calculating the bump automatically. I had a simpler change for testing, which I forgot to complete.

https://github.com/openshift/cluster-etcd-operator/commit/7a5c8910796e6314d0af6ff715eeb1ef522b25ba

heartbeat/podman-etcd
oalbrigt: Wouldnt it be a lot easier to mount a log volume?

jaypoulz: I mean the output of `podman logs {container}`
I'm pretty sure they are stored in `/var/log/container` by default, but they are deleted when then container is deleted.

jaypoulz: Wait I lost the context here. You're saying that this line: https://github.com/ClusterLabs/resource-agents/pull/2062/files#diff-2600c60d13bcbda50a7ac6542e535b71029aabd73ddd86e188ea997508e1e6e7R345

Renames the container, which in turn will ensure that podman saves its log record at the previous location. That seems right, ignore the comment above :)

jaypoulz: Somehow I saw the `cp` command but missing the `rename` command above it

clobrano: I saw `etcd` logs also in pacemaker's journactl. I thought we wanted to keep the whole container for later inspection.

clobrano:
Updated with the following changes

* Replace bash-specific here-string with POSIX-compatible tr command for cipher suite parsing
* Add function documentation for archive_current_container explaining fallback behavior
* Improve variable naming: is_archive_container_name_available → can_archive_container
* Add input validation to convert_duration_in_nanoseconds function
* Enhance error messages with function context and better descriptions
* Add guard clause to prevent empty cipher-suites section in config file

PR #2063, opened 2025-07-24
Medium: portblock: Add multi-state support with nftables and iptables

zypek:
### Summary

This patch extends the `portblock` resource agent to support multi-state (promotable) functionality. The primary motivation is to allow `portblock` to integrate cleanly with promotable resources in Pacemaker, and coordinate primary/secondary network isolation logic via iptables or nftables.

---

### Changes Introduced

- Added `start`, `stop`, `monitor`, `promote`, and `demote` actions supporting the OCF multi-state interface
- Implemented `check_ms_mode_and_warn_action()` to handle `OCF_RESKEY_CRM_meta_notify_active_*` environment hints
- Ensured compatibility with both iptables and nftables backends
- Added logic to prevent unnecessary promotion when not in multi-state mode
- Improved logging and error handling

---

### Testing Performed

- ✅ Tested as a **primitive resource** in Pacemaker — works as expected
- ✅ Tested as a **multi-state clone** — proper promotion/demotion behaviour verified
- ✅ Verified rule creation and deletion on both iptables and nftables systems
- ✅ Ensured idempotency and graceful demotion during node shutdown/failover

---

### Notes

- The agent remains backward-compatible with existing deployments using `portblock` in non-promotable mode
- This change does not alter any existing behaviour unless used in a multi-state clone

---

Please let me know if you'd prefer this logic split across separate files or if further modularisation is preferred.

Thanks for your consideration!

### DIFF
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -86,8 +86,8 @@ usage()
 	NOTE: iptables is Linux-specific.
 
 	An additional feature in the portblock RA is the tickle ACK function
-	enabled by specifying the tickle_dir parameter. The tickle ACK 
-	triggers the clients to faster reconnect their TCP connections to the 
+	enabled by specifying the tickle_dir parameter. The tickle ACK
+	triggers the clients to faster reconnect their TCP connections to the
 	fail-overed server.
 
 	Please note that this feature is often used for the floating IP fail-
@@ -95,7 +95,7 @@ usage()
 	It doesn't support the cluster alias IP scenario.
 
 	When using the tickle ACK function, in addition to the normal usage
-	of portblock RA, the parameter tickle_dir must be specified in the 
+	of portblock RA, the parameter tickle_dir must be specified in the
 	action=unblock instance of the portblock resources.
 	For example, you may stack resources like below:
 		portblock action=block
@@ -103,18 +103,18 @@ usage()
 		portblock action=unblock tickle_dir=/tickle/state/dir
 
 	If you want to tickle all the TCP connections which connected to _one_
-	floating IP but different ports, no matter how many portblock resources 
-	you have defined, you should enable tickles for _one_ portblock 
+	floating IP but different ports, no matter how many portblock resources
+	you have defined, you should enable tickles for _one_ portblock
 	resource(action=unblock) only.
-	
-	The tickle_dir is a location which stores the established TCP 
-	connections. It can be a shared directory(which is cluster-visible to 
+
+	The tickle_dir is a location which stores the established TCP
+	connections. It can be a shared directory(which is cluster-visible to
 	all nodes) or a local directory.
 	If you use the shared directory, you needn't do any other things.
 	If you use the local directory, you must also specify the sync_script
 	paramater. We recommend you to use csync2 as the sync_script.
-	For example, if you use the local directory /tmp/tickle as tickle_dir, 
-	you could setup the csync2 as the csync2 documentation says and 
+	For example, if you use the local directory /tmp/tickle as tickle_dir,
+	you could setup the csync2 as the csync2 documentation says and
 	configure your /etc/csync2/csync2.cfg like:
 		group ticklegroup {
 		  host node1;
@@ -202,7 +202,7 @@ The IP address used to be blocked/unblocked.
 
 <parameter name="tickle_dir" unique="0" required="0">
 <longdesc lang="en">
-The shared or local directory (_must_ be absolute path) which 
+The shared or local directory (_must_ be absolute path) which
 stores the established TCP connections.
 </longdesc>
 <shortdesc lang="en">Tickle directory</shortdesc>
@@ -244,7 +244,6 @@ If "both" is used, both the incoming and outgoing ports are blocked.
 </resource-agent>
 END
 }
-
 
 #
 #	Because this is the normal usage, we consider "block"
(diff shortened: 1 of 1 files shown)


heartbeat/portblock
```diff
@@ -1,9 +1,10 @@
 #!/bin/sh
 #
-# portblock: iptables temporary portblocking control 
+# portblock: iptables/nftables temporary portblocking control
 #
-# Author:	Sun Jiang Dong (initial version)
+# Author:       Sun Jiang Dong (initial version)
```
wenningerk, line 5: guess these whitespace corrections are useful but maybe split them off into a separate commit prior to your additions.

heartbeat/portblock
```diff
@@ -137,17 +149,25 @@ meta_data() {
...
-Resource script for portblock. It is used to temporarily block ports 
-using iptables. In addition, it may allow for faster TCP reconnects
+Resource script for portblock. It is used to block ports
+using iptables or nftables. In addition, it may allow for faster TCP reconnects
 for clients on failover. Use that if there are long lived TCP
 connections to an HA service. This feature is enabled by setting the
 tickle_dir parameter and only in concert with action set to unblock.
 Note that the tickle ACK function is new as of version 3.0.2 and
-hasn't yet seen widespread use.
+hasn't yet seen widespread use. In multi-state mode, the promote action unblocks the ports on the Promoted/Master node and blocks the ports on the Unpromoted/Slaves node(s).
```
wenningerk, line 158: we should try keeping lines to 80 chars if there is no really good reason not to do so.

heartbeat/portblock
```diff
@@ -273,15 +294,178 @@ active_grep_pat()
...
+    [ "$4" = "s" ] && ip=$src || ip=$dst
+    echo "^\s\+ip $4addr ${ip} $1 $4port $2 ct state { established, related, new } drop$"
+  else
+    echo "^DROP${w}${prot}${w}--${w}${src}${w}${dst}${w}multiport${w}${4}ports${w}${2}$"
+  fi
+}
+
+# Update state file atomically
+stateful_update() {
+    local tempfile
```
wenningerk, line 308: Do we have to take care of atomicness here?
Would've expected pacemaker to do the required serialization already.
Otoh deletion of a file might alter behavior compared to other resource-agent that seem to do that differently. Don't see anybody doing inotify-wait on the state-file but that would definitely not work after a deletion and rename.

heartbeat/portblock
```diff
@@ -273,15 +294,178 @@ active_grep_pat()
...
+    }
+    mv -f "$tempfile" "${OCF_RESKEY_state}" || {
+        ocf_log err "Failed to move state file";
+        rm -f "$tempfile";
+        return $OCF_ERR_GENERIC;
+    }
+}
+
+# Handle state file location
+if [ -z "$OCF_RESKEY_state" ]; then
```
wenningerk, line 326: This logic is obsolete as ${OCF_RESOURCE_INSTANCE} should already take care of potentially multiple instances.

heartbeat/portblock
```diff
@@ -372,43 +556,136 @@ SayInactive()
...
+PortStatus() {
   local rc
   rc=$OCF_ERR_GENERIC
   is_active=0
+
+  ocf_log info "PortStatus: Starting monitor check for $1 ports $2 on $3 direction $4 action $5"
+  ocf_log info "PortStatus: FIREWALL=$FIREWALL"
+
   if [ "$4" = "in" ] || [ "$4" = "both" ]; then
+    ocf_log info "PortStatus: Calling chain_isactive for INPUT chain"
```
wenningerk, line 574: info might be a bit too verbose here.
maybe info for actual results and some debug level for showing how that was gathered.

heartbeat/portblock
```diff
@@ -372,43 +556,136 @@ SayInactive()
...
     case $5 in
       block)
+        # We want to block and rules exist - SUCCESS
         SayActive $*
         rc=$OCF_SUCCESS
         ;;
+      unblock)
+        # We want to unblock but rules still exist - NOT RUNNING
+        SayInactive $*
+        rc=$OCF_NOT_RUNNING
```
wenningerk, line 660: That doesn't feel correct somehow. But have to check in the context.

heartbeat/portblock
```diff
@@ -584,55 +1111,99 @@ IptablesValidateAll()
   return $OCF_SUCCESS
 }
 
-if
-  ( [ $# -ne 1 ] )
-then
+# Detect firewall tool - improved version
+detect_firewall_tool() {
```
wenningerk, line 1115: if this package is updated on a setup that up to now would have used iptables though nftables would've been present we might introduce behavioral change for the case that the iptables (wrapper I think) behaves differently. Upstream this might be Ok but as a courtesy to downstream adopters we might provide a simple way as to influence the logic so that it stays with iptables as long as the executable is available.
My suggestion would be 'auto' and 'safe-auto'  as that downstream people can set the default to 'safe-auto'.
For detect_firewall_tool this would mean a safe parameter that modifies the logic accordingly.

heartbeat/portblock
```diff
@@ -642,31 +1213,40 @@ if [ -n "$OCF_RESKEY_tickle_dir" ] ; then
 	elif have_binary netstat ; then
 		ss_or_netstat="netstat -tn"
 	else
-		ocf_log err "Neither ss nor netstat found, but needed to record estblished connections."
```
wenningerk, line 645: Typos and whitespace can go to the same commit.

wenningerk (commented):
Mainly a few general comments for now.
Will have to review the actual code more thoroughfully.

wenningerk:
What failed in CI was exclusively power9 builders. So I guess an issue with those. Kicked it off again.

PR #2068, opened 2025-08-28
OCPBUGS-60977: podman-etcd: wrap ipv6 address in brackets for attribute_node_ip

eggfoobar:
When trying to determine the node ip address we need to make sure we account for ipv6 and dualstack deployments, and accordingly wrap ipv6 in brackets so it correctly resolves. We also need to prefer ipv6 over ipv4 in dualstack deployments, so when querying for the ip address, we favor ipv6.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -347,21 +347,41 @@ remove_container()
 	return $rc
 }
 
+# Correctly wraps an ipv6 in [] for url otherwise use return normal ipv4 address.
+ip_url() {
+	local ip_addr=$1
+	local value
+	if echo "$ip_addr" | grep -q ":" ; then
+		value="[$ip_addr]"
+	else
+		value="$ip_addr"
+	fi
+	echo "https://$value"
+}
+
 attribute_node_ip()
 {
 	local action="$1"
 	local attribute="node_ip"
-	local value
+	local ip_addr name
 
-	if ! value=$(ip -brief addr show "$OCF_RESKEY_nic" | awk '{gsub("/.*", "", $3); print $3}'); then
-		rc=$?
-		ocf_log err "could not get node ip, error code: $rc"
-		return "$rc"
+	for node in $(echo "$OCF_RESKEY_node_ip_map" | sed "s/\s//g;s/;/ /g"); do
+		name=$(echo "$node" | cut -d: -f1)
+		# ignore other nodes
+		if [ "$name" != "$NODENAME" ]; then
+			continue
+		fi
+		ip_addr=$(echo "$node" | cut -d: -f2-) # Grab everything after the first : this covers ipv4/ipv6
+	done
+
+	if [ -z "$ip_addr" ]; then
+		ocf_log err "ip address was empty when querying (getent ahosts) for hostname: $(hostname -f)"
+		return 1
 	fi
 
 	case "$action" in
 		get)
-			echo "$value"
+			echo "$ip_addr"
 			;;
 		update)
 			if ! crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value"; then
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -352,13 +352,24 @@ attribute_node_ip()
 	local action="$1"
 	local attribute="node_ip"
 	local value
+	local ip_addr
 
-	if ! value=$(ip -brief addr show "$OCF_RESKEY_nic" | awk '{gsub("/.*", "", $3); print $3}'); then
+	# Get the first IP address from the ahosts database, this will get us a list of address
+	# with IPV6 being first, so we just need to grab the first element and exit, since we prefer
+	# IPV6 during DualStack deployments.
+	if ! ip_addr=$(getent ahosts $(hostname -f) | awk '{print $1; exit}'); then
```
clobrano: Cool! :+1:

heartbeat/podman-etcd
```diff
@@ -352,13 +352,24 @@ attribute_node_ip()
...
 		rc=$?
 		ocf_log err "could not get node ip, error code: $rc"
 		return "$rc"
 	fi
 
+	# Check if the IP is IPv6 (contains a colon) and wrap it in brackets for URL usage.
+	if [[ $ip_addr == *":"* ]]; then
```
clobrano: Please, use single brackets. Also, I am not familiar with ahost, could `getent ahost` succeed, but return an empty value?

clobrano (changes requested):
I left a couple of comments

heartbeat/podman-etcd
eggfoobar: Hmm, good point, let me add a check for that, it should return with an ip address but I don't know what would happen if an IP address is not assigned. Updating

heartbeat/podman-etcd
oalbrigt: I get e.g. %2 endings when running this, so I guess you need to run `awk -F "%"` to avoid this issue.

eggfoobar: I did notice that on some of my machines, but from my understanding of it, it might be desirable to keep it as is, it's the [ScopeID](https://learn.microsoft.com/en-us/previous-versions/aa921042(v=msdn.10)#scope-id) that might cover us for dual Nics, but I'm not 100% sure given this is a bit out of my realm.

clobrano: however, as Marc noted, podman-etcd receives the node-ip-map as input. Wouldn't it be safer to parse that directly, since its entries are already in the node-name:ip format?

eggfoobar: Sounds good, updated the script to use that now, with some helper function to help create the url, give it a look when you have a chance.

heartbeat/podman-etcd
```diff
@@ -347,21 +347,41 @@ remove_container()
...
+	fi
+	echo "https://$value"
+}
+
 attribute_node_ip()
 {
 	local action="$1"
 	local attribute="node_ip"
-	local value
+	local value ip_addr name
```
oalbrigt: `value` can be removed, as it's no longer used in this function.

heartbeat/podman-etcd
```diff
@@ -414,21 +434,21 @@ prepare_env() {
 	NODEIP="$(attribute_node_ip get)"
```
oalbrigt, line 434: Just add `IPURL=$(ip_url $NODEIP)` here, and use $IPURL instead of calling `ip_url()` multiple times.

oalbrigt (changes requested):
Just a couple of improvements needed, and we should be good.

heartbeat/podman-etcd
eggfoobar: Ah, good catch, removed

heartbeat/podman-etcd
eggfoobar: Great suggestion, I went ahead and updated the code, please give it another look over, thanks!

heartbeat/podman-etcd
```diff
@@ -347,21 +347,41 @@ remove_container()
 	return $rc
 }
 
+# Correctly wraps an ipv6 in [] for url otherwise use return normal ipv4 address.
+ip_url() {
+	local ip_addr=$1
+	local value
+	if [ $ip_addr == *":"* ]; then
```
clobrano: My IDE says 🙂
> [ .. ] can't match globs. Use a case statement. [SC2081]

I did a quick test and it does seem right
```bash
IPv4="198.168.111.20"
IPv6="fd2e:6f44:5dd8:c956::16"

ip_url() {
	local ip_addr=$1
	local value
	if [ $ip_addr == *":"* ]; then
		value="[$ip_addr]"
	else
		value="$ip_addr"
	fi
	echo "https://$value"
}

ip_url "$IPv4"
ip_url "$IPv6"
```

```
$ ./main.sh
https://198.168.111.20
https://fd2e:6f44:5dd8:c956::16
```

heartbeat/podman-etcd
```diff
@@ -611,7 +633,7 @@ add_member_as_learner()
 	local member_ip=$2
 
 	ocf_log info "add $member_name ($member_ip) to the member list as learner"
-	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="https://$(attribute_node_ip get):2379" member add "$member_name" --peer-urls="https://$member_ip:2380" --learner)
+	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="$(ip_url $(attribute_node_ip get)):2379" member add "$member_name" --peer-urls="$(ip_url $member_ip):2380" --learner)
```
clobrano: For better clarity, let's declare the `endpoint_url` and `peer_url` variables first.

clobrano (changes requested):

heartbeat/podman-etcd
eggfoobar: Oh interesting, let me run shellcheck against it later and see if there's some better suggestions for it.

oalbrigt: Yeah. Also in sh it's a single `=` to make sure developers get it wrong often 😄.

I would use `if echo "$ip_addr" | grep -q ":" ; then` instead.

eggfoobar: Great idea, even better! Updated, thanks guys!

oalbrigt:
Thanks.

PR #2069, opened 2025-09-02
Filesystem: warn if mount GFS2 with SELinux but no 'context' in options

Damenly:
Even GFS2 supports xattr but SELinux attempts to read sclabel element on each file system object. It will cause a small performance drop due to lock mechanism of GFS2 inode and xattr.

It's recommended for user to add 'context=' in options so disk reads of extended attribute block can be avoided.
Here warn it if FSTYPE is GFS2 and SELinux is permissive/enforcing.


oalbrigt:
We do not want to add warnings like this to the agent, as it might cause unwanted issues if people use these these options without the knowledge of what they really do.

PR #2071, opened 2025-09-04
pgsql: Fixed an issue where tmpdir was not created when rep_mode was set to slave.

yiida-ha:
## Problem Details  
When starting a pgsql resource with rep_mode="slave", the following error occurred, causing the start operation to fail.  
The error message is as follows:  

```
Failed Resource Actions:
  * remote-site-pgsql_start_0 on dr-standby1 'error' (1): call=34, status='complete', exitreason='Can't create recovery.conf.', last-rc-change='Thu Sep  4 11:59:06 2025', queued=0ms, exec=309ms
```

## Environment  
* PostgreSQL 17  
* Resource setting: rep_mode="slave"  

## Solution  
In the current code, the tmp directory is not created when rep_mode="slave".  
The existing code creates the tmp directory only if the result of is_replication() is true within the validate_ocf_check_level_10 function.  
To resolve this issue, I modified the validate_ocf_check_level_10 function to create the tmp directory even when rep_mode="slave" is set.

### DIFF
--- a/heartbeat/pgsql
+++ b/heartbeat/pgsql
@@ -1990,6 +1990,16 @@ validate_ocf_check_level_10() {
         fi
     fi
 
+    # For slave mode in PostgreSQL 12 and later, create the tmp directory to place recovery.conf.
+    ocf_version_cmp "$version" "12"
+    rc=$?
+    if [ $rc -eq 1 ]||[ $rc -eq 2 ]; then  # PosrgreSQL 12 or later.
+        if ! mkdir -p $OCF_RESKEY_tmpdir || ! chown $OCF_RESKEY_pgdba $OCF_RESKEY_tmpdir || ! chmod 700 $OCF_RESKEY_tmpdir; then
+            ocf_exit_reason "Can't create directory $OCF_RESKEY_tmpdir or it is not readable by $OCF_RESKEY_pgdba"
+            return $OCF_ERR_PERM
+        fi
+    fi
+
     if use_replication_slot; then
         ocf_version_cmp "$version" "9.4"
         rc=$?


heartbeat/pgsql
```diff
@@ -1990,6 +1990,16 @@ validate_ocf_check_level_10() {
         fi
     fi
 
+    # For slave mode in PostgreSQL 12 and later, create the tmp directory to place recovery.conf.
+    ocf_version_cmp "$version" "12"
+    rc=$?
+    if [ $rc -eq 1 ]||[ $rc -eq 2 ]; then  # PosrgreSQL 12 or later.
+        if ! mkdir -p $OCF_RESKEY_tmpdir || ! chown $OCF_RESKEY_pgdba $OCF_RESKEY_tmpdir || ! chmod 700 $OCF_RESKEY_tmpdir; then
+            ocf_exit_reason "Can't create directory $OCF_RESKEY_tmpdir or it is not readable by $OCF_RESKEY_pgdba"
```
oalbrigt: It should say "writable", and should only be run during the start-action.

yiida-ha: Thank you for your comment.

I modified the code based on the following reasoning:
* The existing code already included a similar process for creating tmpdir within validate_ocf_check_level_10(), so I aligned with that approach.
    * https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/pgsql#L1980-L1983
* As you mentioned, if it should only run during the start action, should I move the above code to the start process as well?
* Additionally, I slightly modified the code so it only runs when rep_mode="slave" is set.

oalbrigt: You should move it to the start-action, as we dont need it for regular actions (if it's needed for monitor or similar we might want to change the logic in those actions).

You'll also have to cover "promoted" or what term they are using in the latest PostgreSQL releases.

yiida-ha: Thank you for your comment.
As requested, I've moved the tmpdir creation process to the start action.
https://github.com/ClusterLabs/resource-agents/pull/2071/commits/cd8e58fa9a7e3ffaf3a7ddcdd75ecb051a828124
Based on my research, all write operations to tmpdir within the start action were consolidated in make_recovery_conf(), so I've grouped the processing there.
How does this approach look?

heartbeat/pgsql
```diff
@@ -1609,6 +1609,25 @@ user_recovery_conf() {
...
+
+    version=`cat $OCF_RESKEY_pgdata/PG_VERSION`
+    if [ "$OCF_RESKEY_rep_mode" = "slave" ]; then
+        # For slave mode in PostgreSQL 12 and later, create the tmp directory to place recovery.conf.
+        ocf_version_cmp "$version" "12"
+        rc=$?
+        if [ $rc -eq 1 ]||[ $rc -eq 2 ]; then  # PosrgreSQL 12 or later.
+            if ! mkdir -p $OCF_RESKEY_tmpdir || ! chown $OCF_RESKEY_pgdba $OCF_RESKEY_tmpdir || ! chmod 700 $OCF_RESKEY_tmpdir; then
+                ocf_exit_reason "Can't create directory $OCF_RESKEY_tmpdir or it is not readable by $OCF_RESKEY_pgdba"
+                return $OCF_ERR_PERM
```
oalbrigt: I think you can remove the outer `if[]` here, and and use `[ "$OCF_RESKEY_rep_mode" = "slave" ] && return $OCF_ERR_PERM || return $OCF_ERR_GENERIC` here instead.

heartbeat/pgsql
```diff
@@ -1609,6 +1609,11 @@ user_recovery_conf() {
 }
 
 make_recovery_conf() {
+    if ! mkdir -p $OCF_RESKEY_tmpdir || ! chown $OCF_RESKEY_pgdba $OCF_RESKEY_tmpdir || ! chmod 700 $OCF_RESKEY_tmpdir; then
```
oalbrigt: Doesnt this fail if the directory already exists?

yiida-ha: Thanks for the comment.
The `-p` option of the `mkdir` command does not fail regardless of whether the directory exists.
Therefore, if the directory exists, `! mkdir -p XXX` is inverted by the `!` and returns exit code `1`.
Since `1` is returned, the subsequent processing proceeds via `||`.

yiida-ha:
Sorry
After test, we will publish the draft.

heartbeat/pgsql
```diff
@@ -582,6 +582,10 @@ pgsql_real_start() {
     check_stat_temp_directory
 
     if [ "$OCF_RESKEY_rep_mode" = "slave" ]; then
+        if ! mkdir -p $OCF_RESKEY_tmpdir || ! chown $OCF_RESKEY_pgdba $OCF_RESKEY_tmpdir || ! chmod 700 $OCF_RESKEY_tmpdir; then
```
oalbrigt, line 585: Based on the description in the metadata you can probably move this to the beginning of `pgsql_start()`.

Maybe the metadata needs updating as well, as it sounds like it's not optional for replication anymore.

PR #2074, opened 2025-09-11
OCPBUGS-61117: podman-etcd: Add cluster-wide force_new_cluster attribute checking

clobrano:
Implement cluster-wide validation of force_new_cluster attribute to resolve race conditions during automated cluster recovery. The enhancement ensures agents check for the cluster-wide attribute before falling back to local etcd revision comparison.

Key changes:
- Enhanced get_force_new_cluster() to query all cluster nodes
- Ensure force_new_cluster is not set in both nodes to prevent conflicting recovery attempts
- Updated startup logic to prioritize cluster-wide attribute checking

fixes OCPBUGS-61117

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -794,54 +794,72 @@ set_force_new_cluster()
 	return $rc
 }
 
+# get_force_new_cluster returns a space-separated list of nodes that have the force_new_cluster attribute set.
+# Return values:
+# - Exit code 0 with non-empty output: One or more nodes have the force_new_cluster attribute set
+# - Exit code 0 with empty output: No nodes have the force_new_cluster attribute set
+# - Exit code 1 with empty output: Error occurred while querying the cluster nodes
 get_force_new_cluster()
 {
-	crm_attribute --lifetime reboot --query --name "force_new_cluster" | awk -F"value=" '{print $2}'
-}
+	local node nodes value
+	local holders=""
 
-clear_force_new_cluster()
-{
-	local force_new_cluster_node
-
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -794,54 +794,63 @@ set_force_new_cluster()
...
+	nodes=$(crm_node -l | awk '{print $2}')
+	if [ -z "$nodes" ]; then
+		echo "ERROR"
+		return 1
+	fi
+
+	for node in $nodes; do
+		if ! value=$(crm_attribute --query --lifetime reboot --name "force_new_cluster" --node "$node" 2>/dev/null | awk -F'value=' '{print $2}' | tr -d "'"); then
+			ocf_log err "could not get force_new_cluster attribute, error code: $?"
+			echo "ERROR"
```
oalbrigt: I would just skip the `echo` here, and use `if  [ $? -eq 1 ] ` instead of `...  = "ERROR"`.

heartbeat/podman-etcd
```diff
@@ -1673,6 +1700,8 @@ podman_stop()
 {
 	local timeout=60
 	local rc
+
+	ocf_log notice "podman-etcd stop"
```
oalbrigt, line 1712: We dont usually put these messages for the actions, but for start/stop-action it isnt that bad. If you did it in the monitor-action it would get spammed to the log e.g. every 60s (or whatever the default monitor interval is).

clobrano: Got it, I know it's not the ideal solution. I was just hoping to add a visual cue to make navigating the dense logs a little easier. I'm totally open to other suggestions! :)

PR #2075, opened 2025-09-22
 OCF Resource Agents for vsftpd failure: No PID found

varkoly:
Enhance documentation: setproctitle_enable and background must be setcorrectly.

Use -f for pgrep to find the right pid.


oalbrigt:
Thanks.

PR #2076, opened 2025-09-25
IPaddr2: add support to kill dangling IP connections 

alfonso-escribano:
I have seen some cases where when you delete an ip address while there are stablished connections, the connections stay dangled.
This patch try to avoid that cases.
First it kills connections before deleting the ip address to inform the clients, and after deleting the ip address to kill connections that can initiated between first kill and ip deletion.

Other approach could be to kill connections in fuction "delete_interface", before "addr delete"

To be able to use "ss -K", you need a kernel with config option CONFIG_INET_DIAG_DESTROY set.

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -696,7 +696,7 @@ kill_ip_connections() {
 		ocf_log error "$ss_output"
 	else
 		ocf_log info "$ss_output"
-        fi
+	fi
 
 	return $OCF_SUCCESS
 }


heartbeat/IPaddr2
```diff
@@ -666,6 +676,37 @@ find_interface() {
...
 
+#
+#        Kill dangling ip connections
+#
+kill_ip_connections() {
+	local ss_output=""
+	local ss_line=""
+	local ss_out_loglevel="info"
+
+	local ipaddr="$1"
```
oalbrigt: This should be the first line of the function.

heartbeat/IPaddr2
```diff
@@ -666,6 +676,37 @@ find_interface() {
...
+
+	check_binary ss
+
+	CMD="ss -Knp src $ipaddr"
```
oalbrigt: CMD and RC should be lowercase to follow the style of the rest of the agent.

heartbeat/IPaddr2
```diff
@@ -666,6 +676,37 @@ find_interface() {
...
+
+	ocf_log info "Calling $CMD"
+
+	ss_output=`$CMD 2>&1`
+	RC=$?
+
+	if [ $RC -ne 0 ]; then
+		ocf_log warn "Command $CMD failed (rc = ${THE_RC})"
```
oalbrigt: Should be `$rc`.

heartbeat/IPaddr2
```diff
@@ -666,6 +676,37 @@ find_interface() {
...
+		ss_out_loglevel="error"
+	fi
+	[ -n "$ss_output" ] && echo "$ss_output" | while read ss_line;
+	do
+		ocf_log $ss_out_loglevel "$ss_line"
```
oalbrigt: It should be OK to just log the output in one call. This will probably end up with other lines in between the output lines.

heartbeat/IPaddr2
```diff
@@ -471,6 +473,14 @@ Consider the resource failed if the interface has status DOWN or LOWERLAYERDOWN.
 <shortdesc lang="en">Consider the resource failed if the interface has status DOWN or LOWERLAYERDOWN</shortdesc>
 <content type="string" default="${OCF_RESKEY_check_link_status_default}"/>
 </parameter>
+
+<parameter name="kill_connections_on_stop">
+<longdesc lang="en">
+Kill connections using 'ss -K src IP' before deleteing interface
```
oalbrigt: Typo in "deleting", and should say "before and after" based on the logic in the stop function.

oalbrigt (changes requested):

heartbeat/IPaddr2
```diff
@@ -666,6 +676,34 @@ find_interface() {
 	return 0
 }
 
+#
+#        Kill dangling ip connections
+#
+kill_ip_connections() {
+	local ss_output
+
+	ipaddr="$1"
```
oalbrigt: This should be the first line, as it's the function variable(s), and make it local to avoid the possibility of it clashing with global variables.

heartbeat/IPaddr2
```diff
@@ -666,6 +676,34 @@ find_interface() {
...
+
+	ocf_log info "Calling $cmd"
+
+	ss_output=`$cmd 2>&1`
+	rc=$?
+
+	if [ $rc -ne 0 ]; then
+		ocf_log warn "Command $cmd failed (rc = $rc)"
+		ocf_log error "ss_output"
+        else
```
oalbrigt: `else` and `fi` should be unindented a level.

oalbrigt (changes requested):

heartbeat/IPaddr2
alfonso-escribano: It has same syntax that "add_interface ()" function (and others functions in this resource), first local variables and then variables with arguments.
I'm changing it, but I think the other functions should be modified as well.

oalbrigt: Ah yeah. It's not always ideal.

We dont want to change it there unless we're changing doing it as part of changing that code.

oalbrigt:
What is the use-case for this?

Usually the service (being run by the cluster as well) would try to close the connections cleanly before the IPaddr2 resource is stopped, if you got them in a group or with constraints to start the service after the IP resource.

If this is because of connections in TIME_WAIT state in `ss` these are already cleanly terminated and just waiting to receive RST packet from the peer after sending RST.

alfonso-escribano:
> What is the use-case for this?
> 
> Usually the service (being run by the cluster as well) would try to close the connections cleanly before the IPaddr2 resource is stopped, if you got them in a group or with constraints to start the service after the IP resource.
> 
> If this is because of connections in TIME_WAIT state in `ss` these are already cleanly terminated and just waiting to receive FIN packet from the peer after sending FIN.

The use case is this (I think there are several cases where this patch can be very useful).

We have 2 nodes with postgresql servers, in streaming, one in host-standby attending read-only queries, and other in read/write attending read/write queries, in front of those servers, we have pgbouncer, one process in each node, same config for two nodes.
We want to change where clients connect to exec read-only queries depending on the servers load or other parameters.

When we want to change ip address for read-only queries from one server to the other, we don't want to stop pgbouncer neither postgresql, we could stop read-only pgbouncer before change ip address, but the problem appears when we have the two ip addresses in same server (and only one pgbouncer for all connections), and we want to change read-only ip address to the other server, we don't want to stop pgbouncer that holds read-write and read-only queries.

In that case, if we change the ip address, connections to the ip address that we move, get dangling in server and also in clients.
If we "kill" them with "ss -K", we avoid dangling connections in server and also in clients.

We have try to improve our linux behavior tunning tcp parameters like tcp_keepalive and othe parameters, but when we remove ip address from de server, it doesn't work as espectedm depens on the server proccess

With this patch we are sure that we close all connections to the ip address before delete it, and clients reaction to that connection reset is quickly.

Also thingking in this other way:
We want to "remove/move/down" an ip address from a cluster node, we know that it's impossible to clients maintain the same connection with that ip address, then the "quickiest" way to inform clients, is to focibly kill the connections from server side.


Other use case where this can be very usefull. 
If you want to have a floating ip address between several nodes, yo don't want to stop/restart services, only change the ip address from server.
I have seen cases where you remove the ip address from a server, but the connection remains in established state, but it's impossible to send/recieve any packets, in that cases, the client remains dangled until some timeouts expires, maybe more than 15 minutes.

I know that it's unusual to need this behavior, but I'm also sure that this option is going to be very usefull for many use cases
Also it's an optional parameter, if you don't use it, it doesn't change anything

PR #2078, opened 2025-09-30
OCPBUGS-42808: podman-etcd: add automatic learner member promotion

clobrano:
Automatically promote etcd learner members to voting members when detected.
Includes refactored member management functions and improved validation.

See https://issues.redhat.com/browse/OCPBUGS-42808
Coincidentally, it also adresses https://issues.redhat.com/browse/OCPBUGS-57467

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -712,6 +712,22 @@ attribute_node_revision_peer()
 	crm_attribute --query --type nodes --node "$nodename" --name "revision" | awk -F"value=" '{print $2}'
 }
 
+# Converts a decimal number to hexadecimal format with validation
+# Args: $1 - decimal number (test for non-negative integer too)
+# Returns: 0 on success, OCF_ERR_GENERIC on invalid input
+# Outputs: hexadecimal representation to stdout
+decimal_to_hex() {
+	local dec=$1
+
+	if ! echo "$dec" | grep -q "^[1-9][0-9]*$"; then
+		ocf_log err "Invalid member ID format: '$dec' (expected decimal number)"
+		return $OCF_ERR_GENERIC
+	fi
+
+	printf "%x" "$dec"
+	return $OCF_SUCCESS
+}
+
 attribute_node_member_id()
 {
 	local action="$1"
@@ -737,7 +753,7 @@ attribute_node_member_id()
 				return "$rc"
 			fi
 
-			local value
+			local value value_hex
 			if ! value=$(echo -n "$member_list_json" | jq -r ".header.member_id"); then
 				rc=$?
 				ocf_log err "could not get $attribute from member list JSON, error code: $rc"
@@ -745,8 +761,11 @@ attribute_node_member_id()
 			fi
 
 			# JSON member_id is decimal, while etcdctl command needs the hex version
-			value=$(printf "%x" "$value")
-			if ! crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value"; then
+			if ! value_hex=$(decimal_to_hex "$value"); then
+				ocf_log err "could not convert decimal member_id '$value' to hex, error code: $?"
+				return $OCF_ERR_GENERIC
+			fi
+			if ! crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value_hex"; then
 				rc=$?
 				ocf_log err "could not update etcd $attribute, error code: $rc"
 				return "$rc"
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -712,6 +712,22 @@ attribute_node_revision_peer()
...
+# Converts a decimal number to hexadecimal format with validation
+# Args: $1 - decimal number (must be positive integer)
+# Returns: 0 on success, OCF_ERR_GENERIC on invalid input
+# Outputs: hexadecimal representation to stdout
+decimal_to_hex() {
+	local dec=$1
+
+	# test decimal input: the variable expansion removes from the input any string
+	# with at least one non-digit, so valid decimal inputs won't result in empty strings.
+	if [ -z "${dec##*[!0-9]*}" ] 2>/dev/null; then
```
jaypoulz: I've tested this, and it works, but I can't explain why.
"##*" is a greedy prefix match.
"[!0-9]*" = A prefix ended in zero or more non decimal digits
String I thought would break this: "A12"

In my head, the prefix match would take the longest string ending in a non-decimal digit, or 'A' and leave you with 12.

It works that way if we omit the last '*', because the greedy match stops at the A.

Can you explain this?

heartbeat/podman-etcd
```diff
@@ -905,42 +924,70 @@ clear_standalone_node()
...
+	if ! learner_member_id_hex=$(decimal_to_hex "$learner_member_id"); then
+		ocf_log err "could not convert decimal member_id '$learner_member_id' to hex, error code: $?"
+		return $OCF_ERR_GENERIC
+	fi
+	if ! ocf_run podman exec "${CONTAINER}" etcdctl member promote "$learner_member_id_hex" 2>&1; then
+		# promotion is expected to fail if the peer is not yet up-to-date
+		ocf_log info "could not promote member $learner_member_id_hex, error code: $?"
+		return $OCF_SUCCESS
+	fi
+	ocf_log info "success promote member $learner_member_id_hex"
```
jaypoulz: nit: `successfully promoted member...`

heartbeat/podman-etcd
```diff
@@ -1014,42 +1061,38 @@ get_member_list_json() {
...
-	# Check peers endpoint status and locally accessible member list
-	local member_list_json
-
-	if ! container_exists; then
-		# we need a running container to execute etcdctl.
-		return $OCF_SUCCESS
+	endpoint_status_json=$(get_endpoint_status_json)
+	ocf_log info "endpoint status: $endpoint_status_json"
+
+	count_endpoints=$(printf  "%s" "$endpoint_status_json" | jq -r ".[].Endpoint" | wc -l)
```
jaypoulz: nit: extra space in the printf ?

heartbeat/podman-etcd
```diff
@@ -1014,42 +1061,38 @@ get_member_list_json() {
...
+	if [ "$count_endpoints" -eq 1 ]; then
+		ocf_log info "one endpoint only: checking status errors"
+		endpoint_status_errors=$(printf "%s" "$endpoint_status_json" | jq -r ".[0].Status.errors")
+		if echo "$endpoint_status_errors" | grep -q "no leader"; then
+			set_force_new_cluster
+			set_standalone_node
+			ocf_exit_reason "$NODENAME must force a new cluster"
+			return $OCF_ERR_GENERIC
+		fi
+		if [ "$endpoint_status_errors" != "null" ]; then
```
jaypoulz, line 1079: I am surprised jq returns null for this. I would expect an empty object. But if it's working :man_shrugging:

heartbeat/podman-etcd
clobrano: the catch is that `*[!0-9]*` matches any string that _contains_ at least one non-digit character, not just the non-digit part. So it will catch "A12" completely, returning a zero string.

clobrano: I will update the comment with something clearer

clobrano: actually, it can be much simplified with grep

oalbrigt: Maybe you could do `if ! echo "$dec" | grep -q "^[1-9][0-9]*$"; then` instead.

heartbeat/podman-etcd
clobrano: yeah, this is not new code, just moved.
It returns `null` because the `.Status.errors` field is not always defined, so if an endpoint doesn't have the field, jq returns null.

heartbeat/podman-etcd
clobrano: indeed, I'm testing something similar, but I wasn't thinking of forcing the first digit to be 1 :+1:

oalbrigt: It's looking for 1-9 as the first digit, so it wont match 0.

clobrano: yes, it's a good idea :)

clobrano: and forgot to add it. I'm on it

jaypoulz: AH gotcha - it's glob rules, not regex rules.
So * is anymatch on both sides. [!0-9] is a non-decimal digit :)

PR #2085, opened 2025-10-17
OCPEDGE-1502: feat: podman-etcd: add support for cert rotation

eggfoobar:
added a cert check function to the monitor call to force a restart of etcd when the certs have been changed

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -40,6 +40,7 @@
 # Parameter defaults
 OCF_RESKEY_image_default="default"
 OCF_RESKEY_pod_manifest_default="/etc/kubernetes/static-pod-resources/etcd-certs/configmaps/external-etcd-pod/pod.yaml"
+OCF_RESKEY_etcd_certs_dir_default="/etc/kubernetes/static-pod-resources/etcd-certs"
 OCF_RESKEY_name_default="etcd"
 OCF_RESKEY_nic_default="br-ex"
 OCF_RESKEY_authfile_default="/var/lib/kubelet/config.json"
@@ -51,6 +52,7 @@ OCF_RESKEY_backup_location_default="/var/lib/etcd"
 
 : ${OCF_RESKEY_image=${OCF_RESKEY_image_default}}
 : ${OCF_RESKEY_pod_manifest=${OCF_RESKEY_pod_manifest_default}}
+: ${OCF_RESKEY_etcd_certs_dir=${OCF_RESKEY_etcd_certs_dir_default}}
 : ${OCF_RESKEY_name=${OCF_RESKEY_name_default}}
 : ${OCF_RESKEY_nic=${OCF_RESKEY_nic_default}}
 : ${OCF_RESKEY_authfile=${OCF_RESKEY_authfile_default}}
@@ -86,6 +88,15 @@ The Pod manifest with the configuration for Etcd.
 </longdesc>
 <shortdesc lang="en">Etcd pod manifest</shortdesc>
 <content type="string" default="${OCF_RESKEY_pod_manifest_default}"/>
+</parameter>
+
+<parameter name="etcd_certs_dir" required="0" unique="0">
+<longdesc lang="en">
+The Etcd certificates directory mounted into the etcd container.
+The agent will monitor this directory for changes and restart the etcd container if the certificates have changed.
+</longdesc>
+<shortdesc lang="en">Etcd certificates directory</shortdesc>
+<content type="string" default="${OCF_RESKEY_etcd_certs_dir_default}"/>
 </parameter>
 
 <parameter name="image" required="0" unique="0">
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -51,6 +52,7 @@ OCF_RESKEY_backup_location_default="/var/lib/etcd"
 
 : ${OCF_RESKEY_image=${OCF_RESKEY_image_default}}
 : ${OCF_RESKEY_pod_manifest=${OCF_RESKEY_pod_manifest_default}}
+: ${OCF_RESKEY_etcd_certs_dir=${OCF_RESKEY_etcd_certs_dir_default}}
```
clobrano, line 55: Please, also add the parameter documentation

e.g.
`OCF_RESKEY_image` results in a `image` parameter with its own documentation below.

heartbeat/podman-etcd
```diff
@@ -289,12 +291,71 @@ Expects to have a fully populated OCF RA-compliant environment set.
...
 
+etcd_certificates_hash_manager()
+{
+	local action="$1"
+	local current_hash
+	local stored_hash
+
+	# If the certs directory doesn't exist, consider it unchanged
+	if [ ! -d "$OCF_RESKEY_etcd_certs_dir" ]; then
+		ocf_log debug "certificates directory $OCF_RESKEY_etcd_certs_dir does not exist, skipping certificate monitoring"
```
clobrano: We can use "info" level

heartbeat/podman-etcd
```diff
@@ -289,12 +291,71 @@ Expects to have a fully populated OCF RA-compliant environment set.
...
+		echo "$current_hash" > "$ETCD_CERTS_HASH_FILE"
+		ocf_log debug "created initial certificate hash: $current_hash"
+		return $OCF_SUCCESS
+	fi
+
+	case "$action" in
+		"update")
+			if ! echo "$current_hash" > "$ETCD_CERTS_HASH_FILE"; then
+				ocf_log err "failed to update certificate hash file $ETCD_CERTS_HASH_FILE"
+				return $OCF_ERR_GENERIC
```
clobrano: All the errors returned from this function should trigger a `podman-etcd` restart, right?
Even if it cannot read or update the certificate hash, the agent might still be fully functional, I don't think we should stop it

heartbeat/podman-etcd
```diff
@@ -289,12 +291,71 @@ Expects to have a fully populated OCF RA-compliant environment set.
...
+			;;
+		"check")
+			if ! stored_hash=$(cat "$ETCD_CERTS_HASH_FILE"); then
+				ocf_log err "failed to read stored certificate hash from $ETCD_CERTS_HASH_FILE"
+				return $OCF_ERR_GENERIC
+			fi
+			if [ "$current_hash" != "$stored_hash" ]; then
+				ocf_log warn "etcd certificate files have changed (stored: $stored_hash, current: $current_hash)"
+				return $OCF_ERR_GENERIC
+			fi
```
clobrano, line 346: This is the error that should cause an agent restart, we also want to tell pacemaker the exit reason.

e.g. (line 1137)
```
ocf_exit_reason "$NODENAME must force a new cluster"
```

heartbeat/podman-etcd
```diff
@@ -289,12 +291,71 @@ Expects to have a fully populated OCF RA-compliant environment set.
...
 {
 	local rc=$OCF_SUCCESS
 	local out
 
+	# Check if certificate files have changed
+	if ! etcd_certificates_hash_manager "check"; then
+		ocf_exit_reason "monitor etcd certs failed: etcd certificate files have changed, etcd needs to be restarted"
+		return $OCF_ERR_GENERIC
+	fi
+
```
clobrano: Better move this block into the main `podman_monitor` function

clobrano (changes requested):
Overall, it looks quite good, and I've added a few comments. My main concern now is that the nodes can act on certificate change **at the same time**, which result in the etcd-cluster being disrupted until both recover.
First of all, is it expected that both nodes can update certificate at the same time? If so, we might want to synchronize them, e.g. with CIB attributes. There are a few examples in the code, look for `attribute_learner_node` function for example.

heartbeat/podman-etcd
clobrano: I meant, "info" everywhere in place of "debug" which is usually disabled :smile:

oalbrigt: That will cause a lot of log spam if it's for the monitor action. "info" should mainly be used for start/stop-actions, as they're not run repeatedly.

oalbrigt: And I guess I would classify this as a warn-level. You could use the following code to do print warnings for start/stop/validate-all actions:
```
case "$__OCF_ACTION" in
    monitor) local loglevel="debug";;
    *) local loglevel="warn";;
esac
```

heartbeat/podman-etcd
oalbrigt: That depends on whether it should be run for simple_status() as well, which is used for start/stop-actions.

oalbrigt: I see it's running the function from the start-action, so this should be moved to `podman_monitor()`.

heartbeat/podman-etcd
eggfoobar: Ah, missed that, thanks! Updating

heartbeat/podman-etcd
eggfoobar: I like the suggestions here, I decided to remove the noise, so now we should log one info for when the cert has is updated during start up, and the check should be silent unless an error has happened or a warn for when the cert hash has changed

heartbeat/podman-etcd
eggfoobar: That's fair, we can remove this from the update, but note that if the hash can't be written during the `check` phase, that will mean we fail and things restart. With this change we're essentially saying that our ability to watch and write the hash of the certs is a requirement of this agent

heartbeat/podman-etcd
eggfoobar: Sounds good, moving over

heartbeat/podman-etcd
eggfoobar: Sounds good, updating

heartbeat/podman-etcd
eggfoobar: Unless we pass an empty string for `OCF_RESKEY_etcd_certs_dir` which would essentially turn off cert monitoring

oalbrigt: Setting it to an empty value when there is a default value would make it use default-value instead, so you'll have to make a value mean off, or simply `if echo "$OCF_RESKEY_etcd_certs_dir" | grep -q "^/"; then` to mean any value starting with / mean on.

eggfoobar:
@clobrano I don't know about the CIB mechanism, looking into it now so I can port over the `attribute_learner_node` functionality for syncing this restart. I'll update the code soon, but the other suggestions should be up now

heartbeat/podman-etcd
eggfoobar: Oh perfect, I don't think we want to turn it off for now, so having the default kick in would be ideal here

clobrano: > With this change we're essentially saying that our ability to watch and write the hash of the certs is a requirement of this agent

I agree we should put an initial check to ensure we can support certification rotation at boot, we can move this logic in `podman_validate`. However, once we're operating, I'm not sure an error in this feature should be a hard fail-stop. The first and most critical responsibility of the agent is running `etcd` as long as possible :smile:

oalbrigt: Sounds good. It only runs validate for the start and validate-all actions, so you can make it fail there without affecting any other actions.

eggfoobar: Okay, remove the error return so we run as long as we can, added a validation check

eggfoobar:
@clobrano I don't think we'll need the CIB, was able to validate this and the cert rotation is a rolling change, so the first node updates, then the second. I think we'll be adding extra complexity to have CIB in the mix. I haven't had a chance to look at the etcd-operator code yet, but I'll find this out so we can be sure that is the behavior. In my experimentation it was a rolling process from node to node.

clobrano:
> @clobrano I don't think we'll need the CIB, was able to validate this and the cert rotation is a rolling change, so the first node updates, then the second. I think we'll be adding extra complexity to have CIB in the mix. I haven't had a chance to look at the etcd-operator code yet, but I'll find this out so we can be sure that is the behavior. In my experimentation it was a rolling process from node to node.

If that so, I agree, there is no need to add complexity.

> the cert rotation is a rolling change

does it mean the node also reboots, or that CEO just waits for one to be completed before moving to the next?

If the first, we need no logic at all, but I remember we reproduced a scenario when the node does not reboot after a certificate change.

If the latter, we should ensure the process is _long enough_ for the agent to see (and signal) the file change, be stopped (which might require up to [80 seconds](https://github.com/ClusterLabs/resource-agents/blob/0a3c03c013096c6e1356406e1a12a4b1a1d8011a/heartbeat/podman-etcd#L1801-L1817)) and restart (which is usually much quicker), and rejoin the cluster, before the peer node is affected by the certificate change.

PR #2095, opened 2025-11-04
OCPBUGS-60588: podman-etcd: enhance etcd data backup with snapshots and retention

clobrano:
Replace basic data directory backup with proper etcd database snapshot functionality. The new implementation:
- Creates timestamped snapshot files instead of moving the entire data directory
- Stores backups in a non-volatile location (backup_location parameter) instead of the previous volatile HA_RSCTMP directory
- Validates backup file existence and size after creation
- Implements configurable retention policy via max_backup_snapshots parameter
- Automatically cleans up old snapshots to control storage usage

Default retention is set to 3 snapshots, with backups stored in /var/lib/etcd by default. This provides better backup reliability, persistence across reboots, and storage management for etcd databases.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -49,6 +49,7 @@ OCF_RESKEY_reuse_default="0"
 OCF_RESKEY_oom_default="-997"
 OCF_RESKEY_config_location_default="/var/lib/etcd"
 OCF_RESKEY_backup_location_default="/var/lib/etcd"
+OCF_RESKEY_max_backup_snapshots_default="3"
 
 : ${OCF_RESKEY_image=${OCF_RESKEY_image_default}}
 : ${OCF_RESKEY_pod_manifest=${OCF_RESKEY_pod_manifest_default}}
@@ -61,6 +62,7 @@ OCF_RESKEY_backup_location_default="/var/lib/etcd"
 : ${OCF_RESKEY_oom=${OCF_RESKEY_oom_default}}
 : ${OCF_RESKEY_config_location=${OCF_RESKEY_config_location_default}}
 : ${OCF_RESKEY_backup_location=${OCF_RESKEY_backup_location_default}}
+: ${OCF_RESKEY_max_backup_snapshots=${OCF_RESKEY_max_backup_snapshots_default}}
 
 
 #######################################################################
@@ -273,6 +275,17 @@ The directory where the resource agent stores its backups.
 </longdesc>
 <shortdesc lang="en">Resource agent backup directory</shortdesc>
 <content type="string" default="${OCF_RESKEY_backup_location_default}"/>
+</parameter>
+
+<parameter name="max_backup_snapshots" required="0" unique="0">
+<longdesc lang="en">
+Maximum number of etcd database snapshots to retain. When a new snapshot is created,
+older snapshots will be automatically removed to maintain this limit. This helps
+control storage usage while ensuring recent backups are available for recovery.
+Set max_backup_snapshots=0 to disable backups.
+</longdesc>
+<shortdesc lang="en">Maximum number of backup snapshots to retain</shortdesc>
+<content type="integer" default="${OCF_RESKEY_max_backup_snapshots_default}"/>
 </parameter>
 
 </parameters>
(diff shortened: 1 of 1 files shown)


jaypoulz (approved):
Looks good to me, but I'm wondering if we need to also save the static kuberesources in order to be compatible with the cluster-restore script.
https://docs.redhat.com/en/documentation/openshift_container_platform/4.20/html-single/backup_and_restore/index#dr-scenario-2-restoring-cluster-state_dr-restoring-cluster-state

heartbeat/podman-etcd
```diff
@@ -702,20 +714,127 @@ EOF
...
+{
+	local backup_dir="$1"
+	local max_snapshots="$OCF_RESKEY_max_backup_snapshots"
+	local backup_count
+	local backups_to_remove
+	local old_backups
+
+	# Validate max_snapshots is a positive integer
+	if ! echo "$max_snapshots" | grep -q '^[1-9][0-9]*$'; then
+		ocf_log warn "invalid max_backup_snapshots value: $max_snapshots, skipping cleanup"
```
fonta-rh: We might want to update the error message to give more info about the regex valitdation. Something like "invalid max_backup_snapshots value: $max_snapshots: it should be a number between 0 and 99. Skipping cleanup"

heartbeat/podman-etcd
```diff
@@ -702,20 +714,127 @@ EOF
...
+	# Check if the etcd database file exists
+	if [ ! -f "$etcd_db_path" ]; then
+		ocf_log warn "etcd database file not found at $etcd_db_path, skipping backup"
+		return $OCF_SUCCESS
+	fi
 
-	dest_dir_name="members-snapshot-$(date +%Y%M%d%H%M%S)"
-	if [ ! -d $data_dir ]; then
-		ocf_log info "no data dir to backup"
+	# Ensure backup directory exists
```
fonta-rh, line 814: Do we want to make sure there is enough disk space before committing to the backup? I worry about users setting a high rotation threshold and filling the disk

clobrano: Good point!
I wonder how to estimate the threshold. From one side, I can use the other backups (if any, of course) + some guard, but if 1 single new backup can fill up the disk, it is probably too late already :grin:. I wonder if we want to use a percentage of the disk available instead :thinking:

heartbeat/podman-etcd
clobrano: Actually, I wonder if we want to allow 0 snapshots, to disable the feature :thinking:

heartbeat/podman-etcd
fonta-rh: I'm torn. The simple "will we fill the disk" check is very straightforward, but as you say, might be useless and doesn't give any options to the user. 

Adding a config parameter for the threshold would be the more complete option, but it's a lot more code. Just the check function might be something like this (Claude dixit), and you still need to call it, document it, etcd

```bash
# Check if filesystem has sufficient space for backup based on usage threshold
#
# Args:
#   $1 - Directory path to check
#
# Returns:
#   0 - Sufficient space available
#   1 - Insufficient space (usage above threshold)
check_backup_disk_space()
{
	local backup_dir="$1"
	local threshold="$OCF_RESKEY_backup_disk_usage_threshold"
	local usage_percent
	local fs_path

	# Validate threshold is a number between 0-100
	if ! echo "$threshold" | grep -q '^[0-9]\{1,3\}$' || [ "$threshold" -gt 100 ]; then
		ocf_log warn "invalid backup_disk_usage_threshold value: $threshold (must be 0-100), using default"
		threshold="$OCF_RESKEY_backup_disk_usage_threshold_default"
	fi

	# Get filesystem usage percentage (without % sign)
	# df output: Filesystem Size Used Avail Use% Mounted
	usage_percent=$(df -P "$backup_dir" 2>/dev/null | awk 'NR==2 {sub(/%/, "", $5); print $5}')

	if [ -z "$usage_percent" ]; then
		ocf_log warn "could not determine disk usage for $backup_dir, skipping space check"
		return 0
	fi

	if [ "$usage_percent" -ge "$threshold" ]; then
		ocf_log warn "backup filesystem usage ${usage_percent}% exceeds threshold ${threshold}%, skipping backup"
		return 1
	fi

	ocf_log debug "backup filesystem usage ${usage_percent}% is below threshold ${threshold}%"
	return 0
}
```

heartbeat/podman-etcd
fonta-rh: As long as it's not the default and we document it, I think that's a good option to give, and easy to add at this point :)

clobrano: yeah, we must take into account that "configurability" now means we need to change CEO to configure the agent differently 😁 
In the future, there might be some real configurability and this option will be useful

heartbeat/podman-etcd
clobrano: > I can use the other backups (if any, of course) + some guard

of course I was hallucinating here... we are backing up _1 specific file_, I **know** what size it has :facepalm:

clobrano: > The simple "will we fill the disk" check is very straightforward, but as you say, might be useless and doesn't give any options to the user.

Considering that this is going to go on a kubernetes node that has better ways to detect disk pressure, I don't think we need to complicate this change further. What do you think?

fonta-rh: Yeah, what about just making we're not the ones filling the disk with something like:

```bash
# Check available disk space before creating backup
local available_kb
available_kb=$(df -P "$backup_dir" | awk 'NR==2 {print $4}')
local db_size_kb
db_size_kb=$(($(stat -c %s "$etcd_db_path") / 1024))

if [ "$available_kb" -lt "$((db_size_kb * 2))" ]; then
    ocf_log warn "insufficient disk space in $backup_dir (available: ${available_kb}KB, needed: ~$((db_size_kb * 2))KB), skipping backup"
    return $OCF_SUCCESS
fi
```

and it's out of our hands!

clobrano: I came up with a similar code :grin: :robot:

heartbeat/podman-etcd
```diff
@@ -702,20 +715,190 @@ EOF
...
+{
+	local backup_dir="$1"
+	local max_snapshots="$OCF_RESKEY_max_backup_snapshots"
+	local backup_count
+	local backups_to_remove
+	local old_backups
+
+	# Validate max_snapshots is a positive integer
+	if ! echo "$max_snapshots" | grep -q '^[1-9][0-9]*$'; then
+		ocf_log warn "invalid max_backup_snapshots value. Positive integer expected, got '$max_snapshots' instead, skipping cleanup"
```
fonta-rh, line 889: Positive integer (1-99) to make the error clearer?

heartbeat/podman-etcd
```diff
@@ -702,20 +715,190 @@ EOF
...
+	fi
+
+	# Calculate how many backups to remove
+	backups_to_remove=$((backup_count - max_snapshots))
+	ocf_log info "removing $backups_to_remove old backup(s) to maintain retention limit of $max_snapshots"
+
+	# Find oldest backups sorted by modification time
+	# -t sorts by modification time, -r reverses (oldest first)
+	# -print0 and -0 handle filenames with spaces/special characters
+	old_backups=$(find "$backup_dir" -maxdepth 1 -name "snapshot-*.db" -type f -print0 2>/dev/null | \
```
fonta-rh, line 908: Claude tells me maybe we don't need -print0 because ls -tr is already going to produce newlines. Not sure it's worth retesting, though, as it does work

heartbeat/podman-etcd
clobrano: Positive because we handle [elsewhere](https://github.com/clobrano/resource-agents/blob/fc7ddfe956f1e67230c550ba2386ba42350479f0/heartbeat/podman-etcd#L785-L788) the case the user doesn't want snapshots

fonta-rh: Ah, nevermind, I just realized that we allow for any positive integer, I thought it was only from 1 to 99 :)

heartbeat/podman-etcd
clobrano: Yeah, the code was using print0 for safety with special characters. Since we aren't meant to use any, I like the idea of simplifying this

clobrano: Oh, I see now why this `print0` was suggested

https://www.shellcheck.net/wiki/SC2038
> Use -print0/-0 or find -exec + to allow for non-alphanumeric filenames.

clobrano:
/un-hold
Ready again for other reviews/mergs

oalbrigt (approved):
LGTM.

oalbrigt:
Thanks.

PR #2097, opened 2025-11-06
storage_mon : fix handling of 4k block devices

wenningerk:
Calculation of seek position was always 512 byte aligned instead of adhering the actual block size of the device.


oalbrigt:
Thanks.

PR #2100, opened 2025-11-16
 OCPBUGS-60098: podman-etcd: prevent last active member from leaving the etcd member list

clobrano:
When stopping an etcd instance, the agent should not leave the member list if it's the last active agent in the cluster. Leaving the member list in this scenario can cause WAL corruption.

This change introduces a check for the number of active resources before attempting to leave the member list. If no other active resources are found, the agent will log a message and skip the leave operation.

NOTE: the check on `standalone_node` might not be enough if both agents stop roughly at the same time, hence none of them has enough time to set the attribute.

Fixes: OCPBUGS-60098

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1341,6 +1341,11 @@ container_health_check()
 	#   recently (i.e. a failure), or not (fresh start)
 	monitor_cmd_exec
 	rc=$?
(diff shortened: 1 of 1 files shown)


fonta-rh:
Is it possible that we want to add a random 0-1 sec delay to account for the case of the two agents leaving at the same time?

oalbrigt (approved):

PR #2102, opened 2025-11-23
ocft: fix failing tests in resource-agents v4.17.0

vvidic:
IPaddr2: exit code has changed when findif fails
nfsserver: custom init script is ignored with systemd
ocft: make sure correct log is displayed on failure


oalbrigt:
Thanks.

PR #2104, opened 2025-11-25
exportfs: fix grep error on stop

### LINKED ISSUE #2070: exportfs agent causes 'error output [ grep: Unmatched ( or \( ]' when cleanup_export_cache called
On attempting a service failover we see the following in the pacemaker logs:

Sep 01 14:55:09.092 hostname.domain.name pacemaker-execd     [25360] (log_op_output)     info: exportfs-resource_stop_0[3192827] error output [ grep: Unmatched ( or \( ]

Looking at the sources for exportfs we believe this is caused by the function 'cleanup_export_cache()' which attempts to build a regular expression matching all possible fsid's for exported file systems and then compare to the contents of /proc/net/rpc/nfsd.export/content:

The line causing the issue is this:

`fsid_re="fsid=(echo `forall get_fsid`|sed 's/ /|/g'),"`

For two exports with fsids 1 and 2 this code results in fsid_re being set to:

`fsid=(echo 1 2|sed 's/ /|/g'),`

whereas I believe the intention was to set it to:

`fsid=(1|2),`

vvidic:
Stop operation does not fail but error is recorded in the output:

```
+ fsid_re=fsid=(echo 105 105
106|sed 's/ /|/g'),
+ :
+ grep -E -q fsid=(echo 105 105
106|sed 's/ /|/g'), /proc/net/rpc/nfsd.export/content grep: Unmatched ( or \(
```

fixes #2070

### DIFF
--- a/heartbeat/exportfs
+++ b/heartbeat/exportfs
@@ -390,10 +390,10 @@ cleanup_export_cache() {
 	local contentfile=/proc/net/rpc/nfsd.export/content
 	local fsid_re
 	local i=1
-	fsid_re="fsid=(echo `forall get_fsid`|sed 's/ /|/g'),"
+	local fsid_all=`forall get_fsid`
+	fsid_re="fsid=(`echo $fsid_all | sed 's/ /|/g'`),"
 	while :; do
-		grep -E -q "$fsid_re" $contentfile ||
-			break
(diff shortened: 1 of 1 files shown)


oalbrigt:
Thanks.

PR #2106, opened 2025-12-01
Filesystem: add force_unmount=move

lge:
There are pathological cases where some unmanaged 3rd party apps keep the file system busy, continuously spawning new users of the fs, while the process that is spawning those new users is not itself using the file system, and thus will not be killed by the existing `force_unmount` methods.

stop will then run into timeout, and failed stop will likely result in a blocked cluster or a hard-reset of the node via fencing.

Educating the end-users of such denial-of-stop workloads to have these workloads "managed" has proven ... difficult.

Luckily Linux can atomically move a mount with all its sub-mounts.

`force_unmount=move` behaves just like `safe`, but attempts to `mount --move /the/mount/point /the/mount/.point` first.

Two preparation commits:
 - Filesystem: improve shell trace (set -x) output
 - Filesystem: tmpfs/overlayfs have no need for systemd_drop_in

A commit reducing the timing window for new users to pop up:
 - Filesystem: try umount immediately after signals are sent

The implementation of force_unmount=move:
 - Filesystem: new force_unmount=move option

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -260,7 +260,7 @@ Only set this to "true" if you know what you are doing!
 </parameter>
 
 <parameter name="force_unmount">
-<longdesc lang="en">
+<longdesc lang="en"><![CDATA[
 This option allows specifying how to handle processes that are
 currently accessing the mount directory.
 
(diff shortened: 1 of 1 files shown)


heartbeat/Filesystem
```diff
@@ -797,13 +846,23 @@ fs_stop() {
...
+	return 0
+	# To test really bad timing of "action timeout":
+	# test -e /tmp/fail-after-move && rm -f /tmp/fail-after-move && kill -KILL $$
+}
+
 #
 # STOP: Unmount the filesystem
 #
-Filesystem_stop()
+Filesystem_stop_()
```
oalbrigt: I think we should keep the start/stop/status in one function, or at least have some sensible names for these if we dont.

E.g. for status() you could call `Filesystem_status simple` or similar when you want to skip the this additional code.

lge: Put it all in one function: it will just increase the indentation level needlessly and produce lots of white-space-only changes. The necessary recursion or iteration within that result will make the program flow harder to follow as well.
We sure can name the functions differently, though. What do you suggest?

lge: Well. I tried your suggestion anyways, and it does not look too bad, we can just "localize" the relevant variables by adding their override assignments in front of the function call.
There is now one level of recursion in case "move" is enabled.

oalbrigt (approved):
LGTM.

I'll leave it open for a couple of days before merging in case anyone else has any questions or suggestions for improvements.

oalbrigt:
Thanks.

Issue #2099, opened 2025-11-14
Regression with recent portblock status_check=rule


lge:
344beb1 introduced status_check=rule (and set that as default).

This breaks "monitor" and "probe" of "action=block" for existing setups.

As soon as the "unblock" is "started", the rule is no longer active.
With status_check=rule (the new default), the next "monitor" (or "probe") will see an unexpected state of the "block" instance, and recover the whole dependency chain. With monitor enabled, this will repeat forever.

Not a fix, even if you think it might be, is to drop the monitor.
Because the same "unexpected" state would be detected on any probe.

Probe will happen after you go into and then out of maintenance-mode, after a DC election, for example after cluster partition / rejoin / or other membership change like a peer reboot, even a planned one. Possibly for other reasons as well.

Any such probe will find the resource in an unexpected state, and "recover" by spuriously restarting the whole dependency chain.

Workaround: explicitly set status_check=pseudo (or any other string not equal to "rule").

Possible "fix": don't set "rule" as the default.

Better fix: always treat "action=block" instances as pseudo resources.

@oalbrigt

oalbrigt:
Yeah. This is not ideal, but not monitoring the rule of a firewall agent by default is really not good from a security perspective.

When you use e.g. block, then unblock like this the solution is to set `status_check=pseudo` for the first resource, so you get the best of both worlds (the first resource thinking it's running, and the 2nd fails if someone or some firewall software changes or removes the rule).

oalbrigt:
Yeah. That makes sense to me.

PR #2108, opened 2025-12-03
portblock: monitor needs to also check state file of inverse action

lge:
portblock: derive state file from parameters instead of instance
    
The expected usage of this agent is to pair a "block" with an "unblock", and order startup and configuration of some service between these.

The established idiom is to have two separate instances with inverse actions.
To "reliably" report the status of "block" during a monitor action, it is not sufficient to check the existence of the blocking rule.

It is also insufficient to rely on the pseudo resource state file of this instance only.

To know our actual expectation, we need to check the state file of the "inverse" instance as well.

Because we don't know the OCF_RESOURCE_INSTANCE value of the other instance, we override the state file name for both instances to something derived from our parameters. 

This should give use the same "global state" view as the "promotion score" does for the promotable clone variant of this agent.

Fixes regression introduced with 344beb1 status_check=rule (and set that as default), which breaks existing setups requiring user interaction. See also #2099. Alternative to #2107.

Also: add save_tcp_connection to monnitor for promoted action=unblock instance

### DIFF
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -30,11 +30,16 @@ OCF_RESKEY_portno_default=""
 OCF_RESKEY_direction_default="in"
 OCF_RESKEY_action_default=""
 OCF_RESKEY_method_default="drop"
-OCF_RESKEY_status_check_default="rule"
 OCF_RESKEY_ip_default="0.0.0.0/0"
 OCF_RESKEY_reset_local_on_unblock_stop_default="false"
 OCF_RESKEY_tickle_dir_default=""
 OCF_RESKEY_sync_script_default=""
+
+if ocf_is_ms; then
+  OCF_RESKEY_status_check_default="rule"
+else
+  OCF_RESKEY_status_check_default="pseudo"
+fi
 
 : ${OCF_RESKEY_firewall=${OCF_RESKEY_firewall_default}}
 : ${OCF_RESKEY_protocol=${OCF_RESKEY_protocol_default}}
@@ -474,8 +479,17 @@ PortStatus() {
         fi
         ;;
       *)
-        SayInactive $*
-        rc=$OCF_NOT_RUNNING
+        if [ "$OCF_RESKEY_status_check" != "rule" ] \
+        && test -e "$state_file" && test "$inverse_state_file" -nt "$state_file"; then
+          # rule present, action=unblock, unblock statefile present,
+          # block state file more recent.
+          # apparently an unusual setup: unblock first, block later
+          SayConsideredActive $*
+          rc=$OCF_SUCCESS
+        else
+          SayInactive $*
+          rc=$OCF_NOT_RUNNING
+        fi
         ;;
     esac
   elif [ "$OCF_RESKEY_status_check" = "rule" ]; then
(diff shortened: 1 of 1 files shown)


heartbeat/portblock
```diff
@@ -487,7 +511,10 @@ PortStatus() {
   else
     case $5 in
       block)
-        if ha_pseudo_resource "${OCF_RESOURCE_INSTANCE}" status; then
+        if test -e "$state_file" && test "$inverse_state_file" -nt "$state_file"; then
```
oalbrigt, line 514: Shouldnt we check `if [ "$OCF_RESKEY_status_check" != "rule" ] \` here as well?

heartbeat/portblock
```diff
@@ -30,12 +30,17 @@ OCF_RESKEY_portno_default=""
 OCF_RESKEY_direction_default="in"
 OCF_RESKEY_action_default=""
 OCF_RESKEY_method_default="drop"
-OCF_RESKEY_status_check_default="rule"
 OCF_RESKEY_ip_default="0.0.0.0/0"
 OCF_RESKEY_reset_local_on_unblock_stop_default="false"
 OCF_RESKEY_tickle_dir_default=""
 OCF_RESKEY_sync_script_default=""
 
+if ocf_is_ms; then
```
oalbrigt, line 38: This will make the metadata suggest it's not running the rule check by default, so I think we should move this down to where you define `state_file_base` and modify `OCF_RESKEY_status_check` instead of the default.

lge:
We could even change the semantics a bit, if we wanted to: up to now, the rule was only active if "block" was started, but "unblock" was not yet started. With "promotable", I think you changed that to "rule present unless unblock promoted", so rule even present after stop.

If we change the two-instance variant to
`test $state_unblock -nt $state_block && rule should NOT be active || rule should be active`
we get the same semantics.

```bash
#!/bin/bash
    state_str=(ALLOW BLOCK);
    for state in "" b "ub" "u b" "b u" u; do
        rm -f block unblock;
        for action in $state; do
            case $action in 
                u) touch unblock;;
                b) touch block;;
                
                # "ambiguous" highly unlikely: both exist, identical timestamps
                ub | bu) touch block unblock;;
            esac;
            sleep 0.1; # so we don't get accidentally get identical timestamps
        done;
        printf "%4s: should %s\n" "$state" "${state_str[$(test unblock -nt block && echo 0 || echo 1)]}";
    done
```
gives
```
    : should BLOCK
   b: should BLOCK
  ub: should BLOCK
 u b: should BLOCK
 b u: should ALLOW
   u: should ALLOW
```

Though that may again violate expectations of existing setups (no rule present if both instances stopped).
Existing setup expectations would be `$(test block -nt unblock && echo 1 || echo 0)`,
which changes above for `"": ALLOW, ub: ALLOW` (rest is identical).
I'll think about this some more.

Pre-calculating the expectation like that could also simplify the status function.

heartbeat/portblock
oalbrigt: Let's move this logic to `OCF_RESKEY_status_check=` below, so the metadata isnt suggesting the default is pseudo.

Do you want me to pull your commits into a new PR to do it, or will you update this one?

We can do the other improvements in a separate PR later.

oalbrigt: There are some other changes needed as well, so I'll take care of it.

PR #2119, opened 2026-01-27
send_arp.linux/tickle_tcp: better alpine compatibility

statzitz:
Removed a duplicated header in send_arp.linux.c

Updated tools/tickle_tcp.c to use a common structure between musl definition of tcphdr and the glibc one

I could build this project and do the make check with my patch on both my glibc archlinux and my musl alpine linux, so I am pretty confident in the fact that it won't break anything


oalbrigt:
Thanks.

PR #2121, opened 2026-01-28
docker: improve image existence check

### LINKED ISSUE #2120: Docker 29+ incompatibility
Docker 29 changed the default `image ls` output format (ref: https://docs.docker.com/engine/release-notes/29/#new-1 https://github.com/docker/cli/discussions/6651). This causes [the script for checking if image exist](https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/docker#L515) to fail.

The script assumed a static output format without specifying any flags and then implicitly parsed the output intended as human-readable. The correct approach would be to specify explicit format, preferably a machine-parsable one (e.g. json), or - in this case - to query for the specific image (see [this SO example](https://stackoverflow.com/a/30543453)) instead of querying all images and searching the list.

mjbogusz:
Query docker explicitly for specific image instead of parsing image list - fixes #2120.

I don't have a testing environment on hand, thus the draft status of this PR - I'll mark it as ready as soon as it's verified to be working correctly.

Any stylistic suggestions or other code improvements are welcome, e.g. I've noticed inconsistent usage of return codes vs `$OCF_SUCCESS`/`$OCF_ERR_ARGS` etc in existing code but I've decided to limit the scope of changes for the time being.

### DIFF
--- a/heartbeat/docker
+++ b/heartbeat/docker
@@ -513,17 +513,9 @@ image_exists()
 	# - image
 	# - server_name/image (as extracted above)
 	# - docker.io/image (some distro will display "docker.io/" as prefix)
-	docker image ls --quiet ${IMAGE_NAME}:${IMAGE_TAG} 2>&1 >/dev/null
+	docker image ls --all --format '{{.Repository}}:{{.Tag}}' | $EGREP -q -s "^(docker.io\/|${SERVER_NAME}\/)?${IMAGE_NAME}:${IMAGE_TAG}\$"
 	if [ $? -eq 0 ]; then
 		# image found
-		return $OCF_SUCCESS
-	fi
-	docker image ls --quiet ${SERVER_NAME}/${IMAGE_NAME}:${IMAGE_TAG} 2>&1 >/dev/null
-	if [ $? -eq 0 ]; then
-		return $OCF_SUCCESS
-	fi
-	docker image ls --quiet docker.io/${IMAGE_NAME}:${IMAGE_TAG} 2>&1 >/dev/null
-	if [ $? -eq 0 ]; then
 		return $OCF_SUCCESS
 	fi
 


heartbeat/docker
```diff
@@ -495,34 +495,44 @@ image_exists()
...
 	# - image
-	# - repository:port/image
+	# - server_name/image (as extracted above)
 	# - docker.io/image (some distro will display "docker.io/" as prefix)
-	docker images | awk '{print $1 ":" $2}' | $EGREP -q -s "^(docker.io\/|${SERVER_NAME}\/)?${IMAGE_NAME}:${IMAGE_TAG}\$"
-	if [ $? -eq 0 ]; then
+	docker image ls --quiet ${IMAGE_NAME}:${IMAGE_TAG} 2>&1 >/dev/null
+	if [ $? -eq $OCF_SUCCESS ]; then
 		# image found
 		return 0
```
oalbrigt: You should use OCF-error codes for the returns and not for the ifs (unless the if was for a funciton in the agent that returns the OCF-codes).

mjbogusz: Fixed, please verify if it's ok now, also thanks for clarifying OCF-error usage

heartbeat/docker
```diff
@@ -495,36 +495,46 @@ image_exists()
...
 		IMAGE_NAME="$(echo ${IMAGE_NAME} | cut -d : -f 1-${COLON_FOUND})"
 	fi
 
-	# IMAGE_NAME might be following formats:
+	# Check for existence of image, which might be in following formats:
 	# - image
-	# - repository:port/image
+	# - server_name/image (as extracted above)
 	# - docker.io/image (some distro will display "docker.io/" as prefix)
-	docker images | awk '{print $1 ":" $2}' | $EGREP -q -s "^(docker.io\/|${SERVER_NAME}\/)?${IMAGE_NAME}:${IMAGE_TAG}\$"
```
oalbrigt, line 515: Let's use the old awk format where you only need to run it once, and use the old `docker images`  command as fallback (to avoid breaking the agent for older versions of docker).

mjbogusz: I've updated the code with a somewhat hybrid approach - it's a single command but I've specified explicit output format that not only should provide resiliency against changes to default output formats but also eliminate the need for awk-ing the output (IMO the least reliable part in the old approach).

The `format` specifier is available at least since [docker 17.06](https://docker-docs.uclv.cu/v17.06/engine/reference/commandline/image_ls/) (I haven't found any easily accessible docs for older versions online), so hopefully that takes care of that aspect as well ;)

LMK if this approach is acceptable.

oalbrigt (approved):

PR #2124, opened 2026-02-05
OCPBUGS-76278: podman-etcd: fix learner node attribute not set after etcdctl failure

clobrano:
Ensure that learner_node attribute is always set when the member list
contains one learner member.

Moreover:
* Ensure set_standalone_node is called after adding a learner member.
* Capture stderr from etcdctl for better error logging.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1082,7 +1082,7 @@ add_member_as_learner()
 	local peer_url=$(ip_url $member_ip)
 
 	ocf_log info "add $member_name ($member_ip) to the member list as learner"
-	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="$endpoint_url:2379" member add "$member_name" --peer-urls="$peer_url:2380" --learner)
+	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="$endpoint_url:2379" member add "$member_name" --peer-urls="$peer_url:2380" --learner 2>&1)
 	rc=$?
 	if [ $rc -ne 0 ]; then
 		ocf_log err "could not add $member_name as learner, error code $rc, etcdctl output: $out"
@@ -1429,10 +1429,22 @@ detect_cluster_leadership_loss()
 manage_peer_membership()
 {
 	local member_list_json="$1"
+	local peer_ip_map_entry
+	local peer_member_name
+	local peer_member_ip
+	local peer_member_id
 
-	# Example of .members[] instance fields in member list json format:
-	# NOTE that "name" is present in voting members only, while "isLearner" in learner members only
-	# and the value is always true (not a string) in that case.
+	# Get peer node name and IP
+	peer_ip_map_entry=$(echo "$OCF_RESKEY_node_ip_map" | tr ';' '\n' | grep -vF "$NODENAME")
+	if [ -z "$peer_ip_map_entry" ]; then
+		ocf_exit_reason "manage_peer_membership: could not parse node_ip_map: '$OCF_RESKEY_node_ip_map'"
+		exit $OCF_ERR_CONFIGURED
+	fi
+	peer_member_name=$(echo "$peer_ip_map_entry" | cut -d: -f1)
+	peer_member_ip=$(echo "$peer_ip_map_entry" | cut -d: -f2-)
+
+	# Parsing the member list's json output to find a "learner" member.
+	# Example of .members[] instance fields in member list json format:	
 	# {
 	#   "ID": <member ID>,
 	#   "name": "<node hostname>",
@@ -1443,26 +1455,28 @@ manage_peer_membership()
 	#       "https://<node IP>:2379"
 	#   ]
 	# }
-	for node in $(echo "$OCF_RESKEY_node_ip_map" | sed "s/\s//g;s/;/ /g"); do
-		name=$(echo "$node" | cut -d: -f1)
-		# do not check itself
-		if [ "$name" = "$NODENAME" ]; then
-			continue
-		fi
+	# NOTE that the "name" field is present in voting members only, while "isLearner"
+	# field in learner members only and the value is always true (not a string) in that case.
+	peer_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .peerURLs | map(test(\"$peer_member_ip\")) | any).ID")
+	if [ -z "$peer_member_id" ]; then
+		ocf_log info "$peer_member_name is not in the members list"
+		add_member_as_learner "$peer_member_name" "$peer_member_ip"
+		set_standalone_node
+		return
+	fi
 
-		# Check by IP instead of Name since "learner" members appear only in peerURLs, not by Name.
-		ip=$(echo "$node" | cut -d: -f2-) # Grab everything after the first : this covers ipv4/ipv6
-		peer_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .peerURLs | map(test(\"$ip\")) | any).ID")
-		if [ -z "$peer_member_id" ]; then
-			ocf_log info "$name is not in the members list"
-			add_member_as_learner "$name" "$ip"
-			set_standalone_node
-		else
-			ocf_log debug "$name is in the members list by IP: $ip"
-			# Errors from reconcile_member_state are logged internally. Ignoring them here prevents stopping a healthy voter agent; critical local failures are caught by detect_cluster_leadership_loss.
-			reconcile_member_state "$member_list_json"
-		fi
-	done
+	# Ensure learner_node attribute is always set when we have a learner member
+	local learner_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .isLearner==true ).ID")
+	local current_learner_node=$(attribute_learner_node get)
+	if [ -n "$learner_member_id" ] && [ -z "$current_learner_node" ]; then
+		ocf_log debug "$peer_member_name found as learner in member list, but learner_node attribute was not set. Updating"
+		attribute_learner_node update "$peer_member_name"
+		return
+	fi
+
+	ocf_log debug "$peer_member_name is in the members list by IP: $peer_member_ip"
+	# Errors from reconcile_member_state are logged internally. Ignoring them here prevents stopping a healthy voter agent; critical local failures are caught by detect_cluster_leadership_loss.
+	reconcile_member_state "$member_list_json"
 }
 
 check_peer()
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -1082,7 +1082,7 @@ add_member_as_learner()
 	local peer_url=$(ip_url $member_ip)
 
 	ocf_log info "add $member_name ($member_ip) to the member list as learner"
-	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="$endpoint_url:2379" member add "$member_name" --peer-urls="$peer_url:2380" --learner)
+	out=$(podman exec "${CONTAINER}" etcdctl --endpoints="$endpoint_url:2379" member add "$member_name" --peer-urls="$peer_url:2380" --learner 2>&1)
```
jaypoulz, line 1085: I think this tweak will pay dividends :D

heartbeat/podman-etcd
```diff
@@ -1443,26 +1444,39 @@ manage_peer_membership()
...
-		if [ -z "$peer_member_id" ]; then
-			ocf_log info "$name is not in the members list"
-			add_member_as_learner "$name" "$ip"
-			set_standalone_node
-		else
-			ocf_log debug "$name is in the members list by IP: $ip"
-			# Errors from reconcile_member_state are logged internally. Ignoring them here prevents stopping a healthy voter agent; critical local failures are caught by detect_cluster_leadership_loss.
-			reconcile_member_state "$member_list_json"
-		fi
+		peer_member_ip=$(echo "$node" | cut -d: -f2-) # Grab everything after the first : this covers ipv4/ipv6
```
jaypoulz: This this IP actually going to grab the peer member IP?

In the case node-1 node-2 where I am node 1:
-> I continue passed the first check and get node2

In the case node-2 node-1 where I am node 1:
-> I capture node 2 and then continue past node 1 and end up with node 2.

In the case node-1 node-2 where I am node 2:
-> I capture node 1 and then continue past node 2 and end up with node 1.

In the case node-2 node-1 where I am node 2:
-> I continue passed the first check and get node1

What about peer member name:
In the case of node-1 node-2 and I am node-1:
It's set to node one then I continue and reset it to node 2. (Correct)

In the case of node-2 node-1 and I am node-1:
It's set to node-2 then I set it to node 1 and then I continue and I'm left with node-1. :red_circle: (Wrong)

You'd have the same issue if you're node 2 and the second in the list. Skipping the loop doesn't skip the assignment.

heartbeat/podman-etcd
```diff
@@ -1443,26 +1444,39 @@ manage_peer_membership()
...
 	done
+
+	local peer_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .peerURLs | map(test(\"$peer_member_ip\")) | any).ID")
+	if [ -z "$peer_member_id" ]; then
+		ocf_log info "$peer_member_name is not in the members list"
```
jaypoulz, line 1462: Because of the issue in the loop above, the log here will be wrong.

heartbeat/podman-etcd
```diff
@@ -1443,26 +1444,39 @@ manage_peer_membership()
...
+		add_member_as_learner "$peer_member_name" "$peer_member_ip"
```
jaypoulz, line 1463: This would be wrong too, if you ended up with mismatched member_name and member_ip

heartbeat/podman-etcd
```diff
@@ -1443,26 +1444,39 @@ manage_peer_membership()
...
+		set_standalone_node
+		return
+	fi
+
+	# Ensure learner_node attribute is always set when we have a learner member
+	local learner_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .isLearner==true ).ID")
```
jaypoulz, line 1469: This assumes we will never have more than one learner. But also, if we have more than one learner, yikes. So this probably fine.

heartbeat/podman-etcd
```diff
@@ -1443,26 +1444,39 @@ manage_peer_membership()
...
+	local current_learner_node=$(attribute_learner_node get)
+	if [ -n "$learner_member_id" ] && [ -z "$current_learner_node" ]; then
+		ocf_log debug "$peer_member_name found as learner in member list, but learner_node attribute was not set. Updating"
```
jaypoulz, line 1472: We need to fix the loop above because it's used in this log and node update call in this section as well.

heartbeat/podman-etcd
```diff
@@ -2209,6 +2223,7 @@ podman_start()
 				peer_node_ip="$(attribute_node_ip_peer)"
 				if [ -n "$peer_node_name" ] && [ -n "$peer_node_ip" ]; then
 					add_member_as_learner "$peer_node_name" "$peer_node_ip"
+					set_standalone_node
```
jaypoulz, line 2226: I haven't wrapped my head around what this is doing.  I'm assuming I'm setting myself as a standalone node because my peer is supposed to clear this when it starts?

heartbeat/podman-etcd
clobrano: Yeah, I didn't think through this change enough. I'll rework it

heartbeat/podman-etcd
clobrano: This is also enforced in etcd configuration
https://github.com/ClusterLabs/resource-agents/blob/35957e291480fbf7261532dbc6cd79fe82d22dfb/heartbeat/podman-etcd#L708

heartbeat/podman-etcd
clobrano: The `standalone_node` property prevents a node from attempting to leave the member list upon stopping, which avoids a `panic: removed all voters` error during reboot. The property is then cleared by the voter node when it promotes the learner peer.
 
There is some overlapping between this property and `learner_node`, but a proper refactoring was postponed to the moment we will have a CI to test resource-agents PRs in a TNF cluster before merging the content.

heartbeat/podman-etcd
```diff
@@ -1429,10 +1429,22 @@ detect_cluster_leadership_loss()
 manage_peer_membership()
 {
 	local member_list_json="$1"
+	local peer_ip_map_entry
+	local peer_member_name
+	local peer_member_ip
+	local peer_member_id
+
+	# Get peer node name and IP
+	peer_ip_map_entry=$(echo "$OCF_RESKEY_node_ip_map" | tr ';' '\n' | grep -v "$NODENAME")
```
fonta-rh: If we're going for the literal $NODENAME string, do we want to add the -F option? If not, hostnames that have basic regex characters (. mainly) might cause trouble. 
Something like 
`grep -vxF "$NODENAME"`
maybe?

heartbeat/podman-etcd
```diff
@@ -1429,10 +1429,22 @@ detect_cluster_leadership_loss()
...
+	if [ -z "$peer_ip_map_entry" ]; then
+		ocf_exit_reason "manage_peer_membership: could not parse node_ip_map: '$OCF_RESKEY_node_ip_map'"
+		exit $OCF_ERR_CONFIGURED
```
fonta-rh, line 1441: Should this be an exit, or a return? if we use exit during a monitor operation, will that kill the agent?

heartbeat/podman-etcd
fonta-rh, line 2226: So we run set_standalone_node **always** inside this if, it doesn't matter if "add_member_as_learner" was succesful? That sounds strange to me

heartbeat/podman-etcd
```diff
@@ -1443,26 +1455,28 @@ manage_peer_membership()
...
-			# Errors from reconcile_member_state are logged internally. Ignoring them here prevents stopping a healthy voter agent; critical local failures are caught by detect_cluster_leadership_loss.
-			reconcile_member_state "$member_list_json"
-		fi
-	done
+	# Ensure learner_node attribute is always set when we have a learner member
+	local learner_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .isLearner==true ).ID")
+	local current_learner_node=$(attribute_learner_node get)
+	if [ -n "$learner_member_id" ] && [ -z "$current_learner_node" ]; then
+		ocf_log debug "$peer_member_name found as learner in member list, but learner_node attribute was not set. Updating"
+		attribute_learner_node update "$peer_member_name"
```
fonta-rh, line 1473: Can this fail?

heartbeat/podman-etcd
```diff
@@ -1443,26 +1455,28 @@ manage_peer_membership()
...
-			add_member_as_learner "$name" "$ip"
-			set_standalone_node
-		else
-			ocf_log debug "$name is in the members list by IP: $ip"
-			# Errors from reconcile_member_state are logged internally. Ignoring them here prevents stopping a healthy voter agent; critical local failures are caught by detect_cluster_leadership_loss.
-			reconcile_member_state "$member_list_json"
-		fi
-	done
+	# Ensure learner_node attribute is always set when we have a learner member
+	local learner_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .isLearner==true ).ID")
```
fonta-rh, line 1469: Claude tells me (I have never found this before but it seems to be common) that using local and the $() result like this will mask the error on the command substitution (local will always return 0). I guess we're not checking that directly, but might be worth declaring the variable separately and then populating it on the next line

heartbeat/podman-etcd
clobrano: I would exit here and kill the agent. If we cannot parse `OCF_RESKEY_node_ip_map`, which at this point we must have used already (e.g. in start), it is a great problem.

heartbeat/podman-etcd
clobrano: We are running `standalone` regardless we succeeded setting the peer as learner.

heartbeat/podman-etcd
clobrano: yes, and if so, we'll retry at the next monitor loop

heartbeat/podman-etcd
clobrano: Claude's right :grin: 

> Declare and assign separately to avoid masking return values. [SC2155]

https://www.shellcheck.net/wiki/SC2155

We are indeed ignoring this error as, to my understanding:

> This may prevent conditionals, set -e and traps from working correctly.

and we don't use `set -e` nor traps

heartbeat/podman-etcd
clobrano: I agree on `-F` as it is defined as _F, --fixed-strings Interpret PATTERNS as fixed strings, not regular expressions._, however `-x`(which seems more related to regex) breaks the logic.

PR #2130, opened 2026-03-09
OCPBUGS-76538: podman-etcd: monitor/stop hardening

vimauro:
## Summary

  Harden the podman-etcd resource agent's monitor and stop operations to
  prevent unnecessary fencing cascades in two-node clusters when etcd is
  transiently unresponsive (e.g. during high I/O).

  - **Monitor retry logic**: `monitor_cmd_exec` now retries up to 3 times
    with an 8s per-attempt timeout, preventing a single transient failure
    from triggering a monitor failure. (`podman exec` doesn't have a timeout by default, so if etcd is unresponsive, it will take the whole 25s agent timeout and be killed by pacemaker )
  - **Faster etcdctl failures in monitor**: etcdctl calls in the monitor
    path use a 3s command-timeout (down from the default 5s), keeping
    total monitor execution within the 25s budget (in cases where `podman exec` is responsive, but `etcdctl` takes longer than expected)
  - **Stop no longer hangs on unresponsive etcd**: container state check
    uses `podman inspect` instead of `podman exec`, and
    `leave_etcd_member_list` is time-boxed to 30s

Some minor cosmetic fixes based on what I've found while analyzing some failure logs:
 - uninitialized `$rc` in `attribute_node_member_id`
 - `$?` always 0 after `if !` in `leave_etcd_member_list`
 - `$revision` typo in log message

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -371,26 +371,36 @@ monitor_cmd_exec()
 {
 	local rc=$OCF_SUCCESS
 	local out
+	local attempt
+	# 3 attempts × 8s = 24s worst case, fits within the 25s monitor timeout.
+	# The health check normally completes in <1s; the 8s per-attempt timeout
+	# is a safety net for when the container's process namespace is slow.
+	local max_attempts=3
+	local attempt_timeout=8
 
-	out=$(podman exec ${CONTAINER} $OCF_RESKEY_monitor_cmd 2>&1)
-	rc=$?
-	# 125: no container with name or ID ${CONTAINER} found
-	# 126: container state improper (not running)
-	# 127: any other error
-	# 255: podman 2+: container not running
-	case "$rc" in
-		125|126|255)
-			rc=$OCF_NOT_RUNNING
-			;;
-		0)
-			ocf_log debug "monitor cmd passed: exit code = $rc"
-			;;
-		*)
-			ocf_exit_reason "monitor cmd failed (rc=$rc), output: $out"
-			rc=$OCF_ERR_GENERIC
-			;;
-	esac
+	for attempt in $(seq 1 $max_attempts); do
+		out=$(timeout $attempt_timeout podman exec ${CONTAINER} $OCF_RESKEY_monitor_cmd 2>&1)
+		rc=$?
+		# 125: no container with name or ID ${CONTAINER} found
+		# 126: container state improper (not running)
+		# 127: any other error
+		# 255: podman 2+: container not running
+		case "$rc" in
+			125|126|255)
+				rc=$OCF_NOT_RUNNING
+				;;
+			0)
+				ocf_log debug "monitor cmd passed: exit code = $rc"
+				return $OCF_SUCCESS
+				;;
+			*)
+				ocf_log warn "monitor cmd failed (rc=$rc), output: $out"
+				rc=$OCF_ERR_GENERIC
+				;;
+		esac
+	done
 
+	ocf_exit_reason "monitor cmd failed after $max_attempts attempts (last rc=$rc), output: $out"
 	return $rc
 }
 
@@ -986,7 +996,7 @@ attribute_node_revision()
 		update)
 			if ! crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value"; then
 				rc=$?
-				ocf_log err "could not update etcd $revision, error code: $rc"
+				ocf_log err "could not update etcd $attribute, error code: $rc"
 				return "$rc"
 			fi
 			;;
@@ -1038,11 +1048,11 @@ attribute_node_member_id()
 			;;
 		update)
 			local member_list_json
-			member_list_json=$(get_member_list_json)
+			member_list_json=$(get_member_list_json "$MONITOR_ETCDCTL_TIMEOUT")
 			ocf_log info "member list: $member_list_json"
 			if [ -z "$member_list_json" ] ; then
 				ocf_log err "could not get $attribute: could not get member list JSON"
-				return "$rc"
+				return $OCF_ERR_GENERIC
 			fi
 
 			local value value_hex
@@ -1383,23 +1393,25 @@ get_all_etcd_endpoints() {
 get_endpoint_status_json()
 {
 	# Get the status of all endpoints
+	local cmd_timeout="${1:-5s}"
 	local all_etcd_endpoints
 
 	all_etcd_endpoints=$(get_all_etcd_endpoints)
-	podman exec "${CONTAINER}" etcdctl endpoint status --endpoints="$all_etcd_endpoints" -w json
+	podman exec "${CONTAINER}" etcdctl endpoint status --command-timeout="$cmd_timeout" --endpoints="$all_etcd_endpoints" -w json
 }
 
 get_member_list_json() {
 	# Get the list of members visible to the current node
+	local cmd_timeout="${1:-5s}"
 	local this_node_endpoint
 
 	this_node_endpoint="$(ip_url $(attribute_node_ip get)):2379"
-	podman exec "${CONTAINER}" etcdctl member list --endpoints="$this_node_endpoint" -w json
+	podman exec "${CONTAINER}" etcdctl member list --command-timeout="$cmd_timeout" --endpoints="$this_node_endpoint" -w json
 }
 
 detect_cluster_leadership_loss()
 {
-	endpoint_status_json=$(get_endpoint_status_json)
+	endpoint_status_json=$(get_endpoint_status_json "$MONITOR_ETCDCTL_TIMEOUT")
 	ocf_log info "endpoint status: $endpoint_status_json"
 
 	count_endpoints=$(printf "%s" "$endpoint_status_json" | jq -r ".[].Endpoint" | wc -l)
@@ -1489,7 +1501,7 @@ check_peer()
 		return $OCF_SUCCESS
 	fi
 
-	if ! member_list_json=$(get_member_list_json); then
+	if ! member_list_json=$(get_member_list_json "$MONITOR_ETCDCTL_TIMEOUT"); then
 		ocf_log info "podman failed to get member list, error code: $?"
 		detect_cluster_leadership_loss
 		return $?
@@ -2288,8 +2300,9 @@ leave_etcd_member_list()
 	ocf_log info "leaving members list as member with ID $member_id"
 	local endpoint
 	endpoint="$(ip_url $(attribute_node_ip get)):2379"
-	if ! ocf_run podman exec "$CONTAINER" etcdctl member remove "$member_id" --endpoints="$endpoint"; then
-		rc=$?
+	ocf_run podman exec "$CONTAINER" etcdctl member remove "$member_id" --endpoints="$endpoint"
+	rc=$?
+	if [ $rc -ne 0 ]; then
 		ocf_log err "error leaving members list, error code: $rc"
 	fi
 }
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -371,25 +371,34 @@ monitor_cmd_exec()
...
-			;;
-	esac
+	local attempt
+	# 3 attempts × 8s = 24s worst case, fits within the 25s monitor timeout.
+	# The health check normally completes in <1s; the 8s per-attempt timeout
+	# is a safety net for when the container's process namespace is slow.
+	local max_attempts=3
+	local attempt_timeout=8
+
+	for attempt in $(seq 1 $max_attempts); do
```
clobrano, line 381: Considering that OCF_RESKEY_monitor_cmd is just `/bin/true`, I will even allow a second retry only. I think it is very likely that if we need 24 seconds for this command to succeed, the other monitor operations will not have enough time to complete before the 25s overall limit.

heartbeat/podman-etcd
```diff
@@ -986,7 +995,7 @@ attribute_node_revision()
 		update)
 			if ! crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value"; then
 				rc=$?
-				ocf_log err "could not update etcd $revision, error code: $rc"
+				ocf_log err "could not update etcd $attribute, error code: $rc"
```
clobrano, line 1006: Good catch!
Since we are modifying this, we can also fix the `rc`. I made this mistake in other parts of the code, but the `rc=$?` above catches `if !` return code, not `crm_attribute...` one. So either we remove it and return `OCF_ERR_GENERIC`, or we do something like

```sh
crm_attribute --type nodes --node "$NODENAME" --name "$attribute" --update "$value"
rc=$?
if [ "$rc" -eq 0 ]; then
    ocf_log err "could not update etcd $attribute, error code: $rc"
    return "$OCF_ERR_GENERIC"
fi
```

heartbeat/podman-etcd
```diff
@@ -1038,11 +1047,11 @@ attribute_node_member_id()
 			;;
 		update)
 			local member_list_json
-			member_list_json=$(get_member_list_json)
+			member_list_json=$(get_member_list_json "$MONITOR_ETCDCTL_TIMEOUT")
 			ocf_log info "member list: $member_list_json"
 			if [ -z "$member_list_json" ] ; then
 				ocf_log err "could not get $attribute: could not get member list JSON"
-				return "$rc"
+				return $OCF_ERR_GENERIC
```
clobrano, line 1062: Good catch

heartbeat/podman-etcd
```diff
@@ -1383,23 +1392,25 @@ get_all_etcd_endpoints() {
 get_endpoint_status_json()
 {
 	# Get the status of all endpoints
+	local cmd_timeout="${1:-5s}"
```
clobrano: I would simplify `get_endpoint_status_json` calls and embed the constant default timeout here, freeing the caller from the burden to pass it as input.

```sh
local cmd_timeout="${1:-$MONITOR_ETCDCTL_TIMEOUT}"
```

(Please double check the syntax above)

heartbeat/podman-etcd
```diff
@@ -2310,14 +2322,23 @@ podman_stop()
...
 		ocf_log info "could not leave members list: etcd container not running"
 		attribute_node_member_id clear
 		return $OCF_SUCCESS
 	fi
 
-	leave_etcd_member_list
+	# Time-box the member list removal so it cannot consume the entire stop
+	# budget when etcd is unresponsive. 30s accommodates the etcdctl call
+	# plus the optional DELAY_SECOND_NODE_LEAVE_SEC (10s) sleep.
+	timeout 30 leave_etcd_member_list || \
```
clobrano: Wouldn't it be more clear to move the timeout inside `leave_etcd_member_list`?

clobrano (changes requested):
I left a few comments

heartbeat/podman-etcd
vimauro: that's a good point actually. 
We can also increase the number of retries and reduce the timeout, or maybe we could do 3 retries with a 5s timeout which should leave us at least 10s for other operations to complete. WDYT ?

heartbeat/podman-etcd
vimauro: I wanted to time box the entire `leave_etcd_member_list` call into 30s since there are other operations being performed, but I think it makes sense to just add the timeout in `podman exec`. 
Change done!

heartbeat/podman-etcd
vimauro: Done, thanks!

heartbeat/podman-etcd
vimauro: yeah I noticed the same, initially i wanted to reduce the number of changes of this PR, but now I've updated the `$?` usage in all places, let me know if I've missed anything

heartbeat/podman-etcd
vimauro: I've now reduced it to 5s x 3 attempts

heartbeat/podman-etcd
```diff
@@ -2308,14 +2331,19 @@ podman_stop()
...
 
-	podman_simple_status
-	if [ $? -eq  $OCF_NOT_RUNNING ]; then
+	# Use podman inspect instead of podman exec (podman_simple_status) to check
+	# container state. podman exec enters the container's process namespace and
+	# hangs when etcd is unresponsive — the typical scenario that triggers a stop.
+	local container_running
+	container_running=$(podman inspect --format '{{.State.Running}}' "$CONTAINER" 2>/dev/null)
+	if [ "$container_running" != "true" ]; then
 		ocf_log info "could not leave members list: etcd container not running"
```
eggfoobar: Since we're updating this part of the code, lets take the chance to use the parameters from the command to log the info, just incase for some reason `container_running` responds with an unexpected value.

something like

```
ocf_log info "could not leave members list: $CONTAINER container not running, running state: $container_running"
```

eggfoobar (commented):
Just a small suggestion, otherwise looks good!

heartbeat/podman-etcd
vimauro: Good suggestion, I've pushed the change

heartbeat/podman-etcd
clobrano: I think that if `/bin/true` times out we are already doomed XD
I'm ok with this last iteration though :)

heartbeat/podman-etcd
clobrano: It's ok, hopefully we are not breaking anything that was working for the wrong reasons :)

clobrano (approved):
LGTM
we need to test this very carefully though

fonta-rh (approved):
Looks good to me

heartbeat/podman-etcd
```diff
@@ -1385,22 +1406,27 @@ get_endpoint_status_json()
...
 }
 
 detect_cluster_leadership_loss()
 {
 	endpoint_status_json=$(get_endpoint_status_json)
 	ocf_log info "endpoint status: $endpoint_status_json"
 
+	if [ -z "$endpoint_status_json" ] || ! printf "%s" "$endpoint_status_json" | jq -e '.' >/dev/null 2>&1; then
+		ocf_log err "failed to retrieve endpoint status (podman exec may have failed): $endpoint_status_json"
+		return $OCF_ERR_GENERIC
```
clobrano: This will make Pacemaker trying to restart the agent, but without instructing about what to do (force_new_cluster?). See lines 1435-1438.
I think it is too risky to do this only for a potential temporary error getting the member status.

vimauro: The main reason behind this change is an error I've noticed while running some disrupting tests:

```
INFO: endpoint status: null
jq: error (at <stdin>:0): Cannot iterate over null (null)
```

I think this happens when  `get_endpoint_status_json` returns null (or any other error), causing the `jq` parsing to fail here `count_endpoints=$(printf "%s" "$endpoint_status_json" | jq -r ".[].Endpoint" | wc -l)` 

In any case, let me revert this back to the original state, I agree it's too risky along with many other changes in this PR. We can address this separately if needed!

heartbeat/podman-etcd
```diff
@@ -2084,7 +2107,7 @@ podman_start()
 	run_opts="$run_opts --oom-score-adj=${OCF_RESKEY_oom}"
 
 	if ocf_is_true "$JOIN_AS_LEARNER"; then
-		local wait_timeout_sec=$((10*60))
+		local wait_timeout_sec=$((2*60))
```
fonta-rh, line 2172: Why are we reducing this by 80%? Won't this have the opposite effect of what we want? This will make timeouts more likely

fonta-rh (commented):
I only have a single comment. Maybe it has already been discussed. Is the local timeout reduction because we're now by default set on a few retries in pacemaker? In that case I think it might be good to add an inline comment explaining it, since it's a rather drastic reduction and if we change those retries in the future, we need to update this

heartbeat/podman-etcd
vimauro: This was agree offline with Jay and Carlo. There is a `migration-threshold=5` being introduced in pacemaker config, so this is a way for us to fail-fast and retry since such operation is typically fast: https://github.com/openshift/cluster-etcd-operator/blob/main/pkg/tnf/pkg/pcs/etcd.go

fonta-rh: Makes sense, as long as we document it we'll be good! :muscle:

heartbeat/podman-etcd
```diff
@@ -371,25 +371,34 @@ monitor_cmd_exec()
...
+	for attempt in $(seq 1 $max_attempts); do
+		out=$(timeout $attempt_timeout podman exec ${CONTAINER} $OCF_RESKEY_monitor_cmd 2>&1)
+		rc=$?
+		# 125: no container with name or ID ${CONTAINER} found
+		# 126: container state improper (not running)
+		# 127: any other error
+		# 255: podman 2+: container not running
+		case "$rc" in
+			125|126|255)
+				rc=$OCF_NOT_RUNNING
```
clobrano: Sorry, I didn't notice this the first time.
If the return code is in [125, 126, 255], shouldn't we return immediately? `podman` either failed to find the container or the container is not running. We should probably retry to unknown errors (the last `*` case) or if the command times out.

clobrano (changes requested):
I realize this review has grown more complex than initially expected, and I'm sorry for the back-and-forth. The changes in this PR are genuinely valuable, and I want to make sure we land them safely.

The if `! cmd; then rc=$?` → `cmd; rc=$?; if [ $rc -ne 0 ]` refactor is correct and I appreciate the fix. However, it introduces a subtle change in behavior that needs attention.

Before this PR, `return $rc` was effectively `return 0` everywhere (because `$?` after `if !` is always 0). The code was buggy — errors were silently swallowed — but the effect was that the agent always returned a valid OCF code (OCF_SUCCESS=0) to Pacemaker.

Now that `$rc` captures the real exit code, `return $rc` passes a raw command exit code up to the caller. Depending on the call site, this may or may not reach Pacemaker — but where it does, Pacemaker interprets the return value and might decide between restarting, fencing, etc. — so returning an arbitrary integer could trigger unpredictable behavior.

I believe our goal should be to fix the error code reporting without changing the current agent behavior.

In practice this means:
- Where the return code doesn't reach Pacemaker, return a proper OCF error code.
- Where it does reach Pacemaker, log the real error (which this PR already does :+1:), but explicitly return `$OCF_SUCCESS` with a proper comment. This intentionally preserves the previous behavior.

What do you think?

heartbeat/podman-etcd
vimauro: Good point, we can return earlier for [125, 126, 255] and retry only for any other error. Nice catch !

heartbeat/podman-etcd
```diff
@@ -1567,11 +1573,8 @@ check_peer()
 		return $OCF_SUCCESS
 	fi
 
-	local rc
-	member_list_json=$(get_member_list_json)
-	rc=$?
-	if [ $rc -ne 0 ]; then
-		ocf_log info "podman failed to get member list, error code: $rc"
+	if ! member_list_json=$(get_member_list_json); then
+		ocf_log info "podman failed to get member list, error code: $?"
```
clobrano: This `$?` refers to `if !`, not to `get_member_list_json`

vimauro: I've restored the original logic on this specific block for now because I was thinking to address the logic of `check_peer` and `detect_cluster_leadership_loss` separately because it'll add even more changes. Let me know if you think we should it in the current PR

clobrano: We could just remove the part `, error code: $?` for now. `get_member_list_json` logs enough to help debugging

vimauro: Fixed!

clobrano (approved):
/lgtm

PR #2141, opened 2026-03-26
sybaseASE: dont report error about missing sybaseASE env during probe-action

oalbrigt:
Fixes https://github.com/ClusterLabs/resource-agents/issues/2137.


oalbrigt:
Updated patch based on testing by @arslan9292.

PR #2150, opened 2026-04-14
oraasm: Avoid false positive start/monitor operation when critical processes are offline

arslan9292:
Currently `oraasm` uses `crsctl check has` to monitor the state. The HAS service reports "online" even when internal critical Oracle resources such as "ora.asm", "ora.cssd", "ora.orcl.db", etc. are in an OFFLINE state.

Because the resource checks only the HAS service, the `oraasm` resource remains running on the same node and does not report any failures or trigger a recovery operation.

This fix introduces a new resource attribute which avoids the false positive state in monitor and start operation:

 - `monitor_resources`: A comma separated list of Oracle local/cluster resources which needs to be monitored (defaults to `ora.asm`)

Fixes RH Jira: RHEL-137744

### DIFF
--- a/heartbeat/oraasm
+++ b/heartbeat/oraasm
@@ -20,11 +20,19 @@
 OCF_RESKEY_user_default="grid"
 OCF_RESKEY_diskgroup_default=""
 OCF_RESKEY_home_default=""
+OCF_RESKEY_monitor_resources_default="ora.asm"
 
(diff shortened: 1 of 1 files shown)


oalbrigt (approved):

oalbrigt:
Thanks.

PR #2153, opened 2026-04-22
portblock: add IPv6 support to iptables and nft backends

### LINKED ISSUE #2151: portblock: IPv6 service IPs fail on both `iptables` and `nft` backends
## Summary

`ocf:heartbeat:portblock` does not support IPv6 addresses passed via `OCF_RESKEY_ip`. Both firewall backends (`iptables` and `nft`) use IPv4-specific rule templates and no IP-family detection exists anywhere in the agent. The failure is deferred to the underlying firewall tool, which rejects the colon-bearing address.

This is not an edge case — any invocation with an IPv6 address fails deterministically on current `main`.

## Related prior art

- **#1577** (open since 2020-11-12) — original feature request to add nftables support.
- **PR #2073** (merged 2025-09-10, based on #2063) — landed the `nft` backend. IPv6 was not discussed in either the PR or #2063; the rule templates were written with IPv4-only `ip daddr` / `ip saddr` literals. This issue is not about reverting or superseding that work — it is specifically about the IPv6 gap in **both** the legacy iptables path and the newly-merged nft path.

## Reproduction (empirical, verbatim from a RHEL 9.7 test bed)

Upstream `main` `heartbeat/portblock` was downloaded unchanged to `/tmp/portblock.upstream`. `OCF_ROOT` was set to the local OCF tree; `$NFTABLES` was exported manually (see note at the bottom on `ocf-binaries`).

### `firewall=nft` + IPv6 — FAILS

```
+ nft add table inet portblock
+ nft add chain inet portblock INPUT { type filter hook input priority 0; }
+ nft add chain inet portblock OUTPUT { type filter hook output priority 0; }
+ nft list chain inet portblock INPUT
+ nft insert rule inet portblock INPUT ip daddr fd00:10:1:10::44 tcp dport 2049 \
    ct state { established, related, new } drop
Error: Could not resolve hostname: Address family for hostname not supported
insert rule inet portblock INPUT ip daddr fd00:10:1:10::44 tcp dport 2049 \
    ct state { established, related, new } drop
                                          ^^^^^^^^^^^^^^^^
exit=1
```

The parser caret points precisely at the v6 literal being fed to `ip daddr`. The `inet` family table is created successfully (v6-capable in principle), but `ip daddr` / `ip saddr` only match IPv4 addresses, even inside an `inet` table. IPv6 matching in nftables requires `ip6 daddr` / `ip6 saddr`, or `meta l3proto ipv6` dispatch, neither of which the script emits.

### `firewall=nft` + IPv4 — control, succeeds

Same script, same env, only `OCF_RESKEY_ip` changed:

```
table inet portblock {
    chain INPUT {
        type filter hook input priority filter; policy accept;
        ip daddr 10.1.10.44 tcp dport 2049 ct state { established, related, new } drop
    }
    chain OUTPUT { ... }
}
exit=0
```

### `firewall=iptables` + IPv6 — FAILS

```
iptables v1.8.10 (nf_tables): host/network 'fd00:10:1:10::44' not found
exit=2
```

`$IPTABLES` is inherited from `ocf-binaries` as the literal string `iptables` and is never swapped to `ip6tables` when `$OCF_RESKEY_ip` carries a colon-bearing address. The agent passes the address through unchanged; the backend rejects it.

## Where the code lives

On current `main` (`heartbeat/portblock`):

- `inet` table creation — line 670 (fine).
- IPv4-specific rule templates — lines 573, 575, 612, 616, 618 (`ip daddr` / `ip saddr`).
- IP-family detection — **none** (zero references to `ip6tables`, `ip6`, `inet6`, `IPv6`, `meta l3proto`, or `*:*` pattern matching on `$OCF_RESKEY_ip`).

## Suggested direction

A minimal fix detects IP family once and routes both backends accordingly:

```sh
case "$OCF_RESKEY_ip" in
    *:*) ip_family=ip6 ;;
    *)   ip_family=ip  ;;
esac
```

- **`firewall=iptables` path**: set `IPTABLES=ip6tables` when `ip_family=ip6`. Also switch the single literal `iptables -V` capability probe to `$IPTABLES -V`. All existing `$IPTABLES` call sites then work without further edits.
- **`firewall=nft` path**: parameterise the rule templates so `ip daddr` / `ip saddr` become `${ip_family} daddr` / `${ip_family} saddr`. The `inet` family table already supports both — only the match keyword needs family-awareness.

An alternative for the nft path is to keep a single rule shape using `meta l3proto $family ip${v6_suffix} daddr $ip ...`, but parameterising the keyword is the smaller change and mirrors how the iptables path will work.

Metadata is already IP-agnostic (`ip` defaults to `0.0.0.0/0`); the `longdesc` could add one line noting that IPv6 CIDRs are accepted (e.g. `fd00::/64`, `2001:db8::1/128`).

## Willingness to contribute

Happy to submit a PR covering both backends if the maintainers agree on the general approach. One open question: for the nft path, would you prefer (a) simple keyword parameterisation (`ip`/`ip6`) as above, or (b) a migration to `meta l3proto`-dispatched unified rules? Either is small; (a) keeps closer to the current shape, (b) is more idiomatic nftables but slightly bigger.

## Bonus observation (not the main issue, but may be worth a note)

On a system with `resource-agents-4.13.0.linbit-3.0.33`, the shipped `ocf-binaries` does **not** define `NFTABLES`. Dropping current upstream `portblock` on such a system fails early with `Setup problem: couldn't find command: ` (empty argument to `check_binary`). This is a distribution packaging mismatch rather than an upstream bug — flagging in case it affects documentation / backport guidance.

## Attachments

- Proof-of-concept patch for the iptables backend only (not complete — does not yet cover nft): `portblock.patch` (2 hunks, +8 / -1 lines).
- Full write-up including end-to-end LINSTOR Gateway NFS reproduction, compatibility matrix for the broader stack, and verification of the iptables-path patch: `status-analysis.pdf`.

[status-analysis.pdf](https://github.com/user-attachments/files/26911267/status-analysis.pdf)

yusufyildiz:
Closes #2151.

Previously, `portblock` rejected IPv6 service IPs on both firewall backends:

- **iptables**: `$IPTABLES` is inherited as the literal `iptables`, which rejects colon-bearing addresses.
- **nft**: rule templates use `ip daddr` / `ip saddr`, which only match IPv4 even inside an `inet` family table. IPv6 needs `ip6 daddr` / `ip6 saddr`.

This change detects the IP family from `$OCF_RESKEY_ip` once and routes both backends through a single `${ip_family}` variable:

- **iptables path**: `IPTABLES=${ip_family}tables` selects `iptables` or `ip6tables`.
- **nft path**: rule templates emit `${ip_family} daddr` / `${ip_family} saddr`; the same keyword flows into the `active_grep_pat` regex generator so `monitor` matches what `insert` emits.
- The `iptables -V` capability probe now uses `$IPTABLES`.

## Verification

Exercised on RHEL 9.7 (`iptables`/`ip6tables` v1.8.10 nf_tables, `nft` v1.0.9) across all four combinations:

| # | backend | IP | result | rule written |
|---|---|---|---|---|
| 1 | `iptables` | `10.1.10.44` | `start` exit 0 | `iptables: DROP ... 10.1.10.44 ... dports 2049` |
| 2 | `nft` | `10.1.10.44` | `start` exit 0 | `nft: ip daddr 10.1.10.44 tcp dport 2049 ... drop` |
| 3 | `iptables` | `fd00:10:1:10::44` | `start` exit 0 | `ip6tables: DROP ... fd00:10:1:10::44 ... dports 2049` |
| 4 | `nft` | `fd00:10:1:10::44` | `start` exit 0 | `nft: ip6 daddr fd00:10:1:10::44 tcp dport 2049 ... drop` |

`stop` cleans up in all four; zero leftover rules in `nft list ruleset`, `iptables -S`, or `ip6tables -S` after the run. Tests 1–2 are IPv4 regression checks.

## Scope

Deliberately out of scope for this PR (to keep the review small):

- `tickle_tcp` / `tickle_local` paths (only run when `tickle_dir` is set). The `grep -Fw $OCF_RESKEY_ip` on `ss` output may need adjustment for IPv6 addresses that appear as `[fd00:...]:port` in `ss` output. Happy to follow up.
- Documentation additions (e.g. a `longdesc` example showing `fd00::/64` alongside the existing `0.0.0.0/0` default). Also a trivial follow-up.

Context discussion on approach (keyword parameterisation vs `meta l3proto` dispatch) is in #2151.

### DIFF
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -49,6 +49,14 @@ fi
 : ${OCF_RESKEY_method=${OCF_RESKEY_method_default}}
 : ${OCF_RESKEY_status_check=${OCF_RESKEY_status_check_default}}
 : ${OCF_RESKEY_ip=${OCF_RESKEY_ip_default}}
+
+# IPv6 support: detect family from service IP and parameterise both backends.
+case "$OCF_RESKEY_ip" in
+  *:*) ip_family=ip6 ;;
+  *)   ip_family=ip  ;;
+esac
+IPTABLES=${ip_family}tables
+
 : ${OCF_RESKEY_reset_local_on_unblock_stop=${OCF_RESKEY_reset_local_on_unblock_stop_default}}
 : ${OCF_RESKEY_tickle_dir=${OCF_RESKEY_tickle_dir_default}}
 : ${OCF_RESKEY_sync_script=${OCF_RESKEY_sync_script_default}}
@@ -332,9 +340,9 @@ active_grep_pat()
     local ip
     [ "$4" = "s" ] && ip=$src || ip=$dst
     if [ "$method" = "DROP" ]; then
-      echo "^\s\+ip${w}$4addr${w}${ip}${w}$1${w}$4port${w}$2${w}ct${w}state${w}{${w}established,${w}related,${w}new${w}}${w}drop$"
+      echo "^\s\+${ip_family}${w}$4addr${w}${ip}${w}$1${w}$4port${w}$2${w}ct${w}state${w}{${w}established,${w}related,${w}new${w}}${w}drop$"
     else
-      echo "^\s\+ip${w}$4addr${w}${ip}${w}$1${w}$4port${w}$2${w}ct${w}state${w}{${w}established,${w}related,${w}new${w}}${w}reject${w}with${w}tcp${w}reset$"
+      echo "^\s\+${ip_family}${w}$4addr${w}${ip}${w}$1${w}$4port${w}$2${w}ct${w}state${w}{${w}established,${w}related,${w}new${w}}${w}reject${w}with${w}tcp${w}reset$"
     fi
   else
     if [ "$method" = "DROP" ]; then
@@ -545,7 +553,7 @@ NftDelete()
 {
         local chain=$1 proto=$2 ds=$3 ip=$(echo "$4" | sed "s#/#\\\/#") ports=$5
         # Try both single port and multi-port patterns for handle search
-        local handles=$($NFTABLES -a list chain inet $TABLE $chain 2>/dev/null | awk "/\s+ip ${ds}addr $ip $proto ${ds}port $ports/"' {printf "%d ", $NF}')
+        local handles=$($NFTABLES -a list chain inet $TABLE $chain 2>/dev/null | awk "/\s+${ip_family} ${ds}addr $ip $proto ${ds}port $ports/"' {printf "%d ", $NF}')
         for handle in $handles; do
           ocf_log debug "NftDelete: Deleting $chain rule with handle $handle"
           nft delete rule inet $TABLE $chain handle $handle || {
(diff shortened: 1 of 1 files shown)


oalbrigt:
Thanks.

PR #2154, opened 2026-04-30
ganesha-nfs: new resource agent

yusufyildiz:
Adds an OCF resource agent for NFS-Ganesha at `heartbeat/ganesha-nfs`.

## What

A small (~360 LOC POSIX shell) agent that manages a single `ganesha.nfsd` instance:

- `start`, `stop`, `monitor`, `meta-data`, `validate-all`.
- Does **not** touch `/var/lib/nfs`, the kernel NFS server, rpcbind, or statd.
- Supports **multiple co-located instances on the same host** via distinct `config_file` + `pid_file` parameters (both flagged `unique="1"`), each binding its own listening IP via `nfs_ip`.
- Relies on the operator placing `RecoveryDir` on the replicated filesystem that fails over with the resource, so NFSv4 client lock state migrates correctly. The longdesc calls this out explicitly.

## Why (vs. the existing nfsserver agent)

`ocf:heartbeat:nfsserver` manages the in-kernel Linux NFS server and owns three host-wide singletons (`/var/lib/nfs` bind-mount, `nfs-server.service` + the single `rpc.nfsd` thread pool, `/var/lib/nfs/statd/...`). That caps a cluster to one NFS resource group with one atomic failover unit.

NFS-Ganesha is userspace and per-instance independent, so per-tenant Pacemaker groups become natural with cluster-level anti-collocation. The two agents are complementary; this PR adds the userspace alternative without touching `nfsserver`.

## Implementation notes

- `is_running()` reads `/proc/$pid/comm` to defend against PID reuse (not just `kill -0 $pid`).
- `start` handles "PID alive but not yet listening" with a 10s probe before tearing the sick process down — never spawns a second daemon while a first is unhealthy. The sick-instance `stop` failure is propagated.
- `stop` honours `OCF_RESKEY_CRM_meta_timeout` for SIGTERM, leaves a 5s buffer for SIGKILL escalation.
- `port_listening()` uses `ss` when available, falls back to `/proc/net/tcp{,6}` parsing — works headless.
- `meta-data` is well-formed XML; tested on real config.

## Test plan

Tested on a 3-node AlmaLinux 9.7 cluster, Pacemaker 2.1.2.linbit-4, corosync 3.1.9, NFS-Ganesha 5.9-1, DRBD 9.31 + LINSTOR 1.33 with replicated XFS:

| Scenario | Result |
|---|---|
| `bash -n heartbeat/ganesha-nfs` | clean |
| `meta-data` action | well-formed XML, 7 parameters, 5 actions |
| `validate-all` with real ganesha.conf | `exit=0` |
| Pacemaker resource cycle (start → monitor → stop) | clean |
| Multi-instance on same host (two `ganesha.nfsd` processes, distinct config_file/pid_file/listening IP) | both run side by side |
| NFSv4.2 mount over IPv4 + IPv6 | succeeds; basic R/W ops verified |
| `pcs node standby <peer>` graceful failover | client mount survives, append + read across failover |
| Anti-collocation with a second `ganesha-nfs` resource | tenants placed on separate nodes (soft `-1000`) |

## Out of scope (deliberately)

- DBUS-based per-export reload via `dbus-send` (mentioned upstream as a future enhancement; not needed for failover correctness).
- Helper sourcing à la `nfsserver-redhat.sh` — agent is portable POSIX shell with no distro-specific paths beyond the binary search (`/usr/bin` then `/usr/sbin` then `$PATH`).
- pNFS / FSAL_VFS-vs-FSAL_GLUSTER specifics — left to `ganesha.conf` itself.

## Notes

- Agent name `ganesha-nfs` follows the project's resource-agent naming pattern (single token, hyphen-separated where multi-word). Open to renaming if `ganesha` alone or `nfs-ganesha` is preferred — happy to amend.
- License: GPLv2-or-later, same as existing agents.


heartbeat/ganesha-nfs
```diff
@@ -0,0 +1,361 @@
...
+            sleep 1
+            if port_listening; then
+                ocf_log info "Ganesha became ready"
+                return $OCF_SUCCESS
+            fi
+            j=$((j + 1))
+        done
+        # Still stuck. Don't try to spawn a second process; tear the sick one down first.
+        ocf_log warn "Existing Ganesha PID $(read_pid) not listening after 10s; stopping before retry"
+        ganesha_stop
```
oalbrigt: Pacemaker will run the stop-action when it fails, so no need for this here.

heartbeat/ganesha-nfs
```diff
@@ -0,0 +1,361 @@
...
+        if ! kill -0 "$pid" 2>/dev/null; then
+            rm -f "$OCF_RESKEY_pid_file"
+            ocf_log info "Ganesha stopped cleanly"
+            return $OCF_SUCCESS
+        fi
+        sleep 1
+        i=$((i + 1))
+    done
+
+    ocf_log warn "Ganesha did not exit after ${tmo}s of SIGTERM; escalating to SIGKILL"
```
oalbrigt, line 333: You might want to subtract from the timeout above, or it wont ever get to this part.

heartbeat/ganesha-nfs
```diff
@@ -0,0 +1,361 @@
...
+    done
+}
+
+#######################################################################
+# OCF actions
+
+ganesha_validate() {
+    if [ ! -f "$OCF_RESKEY_config_file" ]; then
+        ocf_exit_reason "Config file not found: $OCF_RESKEY_config_file"
+        return $OCF_ERR_INSTALLED
```
oalbrigt: You might want to return `$OCF_ERR_CONFIGURED` here, as this might just be "config_file" being set to an incorrect location (especially if the value is != default).

oalbrigt (changes requested):
Just a couple of and it should be good.

You also need to add the agent to `doc/man/Makefile.am`, so it generates manpages automatically.

yusufyildiz:
Thanks for the review @oalbrigt — addressed all four points in `27eb43a3`:

- **Line 218** (`ganesha_validate`): now returns `$OCF_ERR_CONFIGURED` when `config_file` does not exist, since this is more likely a misconfigured path than a missing package.
- **Line 270** (`ganesha_start`): dropped the inline `ganesha_stop` call. When the existing PID is alive but not listening after the 10s probe, the agent now just returns `$OCF_ERR_GENERIC` and lets Pacemaker invoke the stop-action.
- **Line 333** (`ganesha_stop`): increased the SIGKILL reservation from 5s to 10s and lowered the floor from 10s to 5s, so the escalation branch is reachable before Pacemaker times the action out, and a small `CRM_meta_timeout` cannot itself overrun the budget.
- **`doc/man/Makefile.am`**: added `ocf_heartbeat_ganesha-nfs.7` in the alphabetical block (between `galera` and `garbd`) for automatic manpage generation.

`bash -n` clean. Happy to re-run the full Pattern-B failover test plan on the AlmaLinux 9.7 cluster if useful before merge.

oalbrigt (approved):

oalbrigt:
CI fails due to the agent not being chmod'ed to 755.
`07:58:51  /bin/sh: line 1: /srv/workspace/resource-agents-pipeline_PR-2154/resource-agents/heartbeat/ganesha-nfs: Permission denied`

yusufyildiz:
Thanks @oalbrigt fixed in `3899b83d`. The agent file was committed with mode 100644 from a Windows working tree where git's executable bit isn't auto-tracked; `git update-index --chmod=+x heartbeat/ganesha-nfs` flips it to 100755 to match the other agents. CI should pick it up on the next run.

oalbrigt:
Thanks.

PR #2155, opened 2026-04-30
OCPBUGS-84695: podman-etcd single-node cluster handling and transitions

jaypoulz:
OCPBUGS-84695: fix(podman-etcd): single-node cluster handling and transitions

Add comprehensive single-node cluster support to podman-etcd resource agent,
including initial deployment, restart scenarios, and transitions between
single-node and two-node configurations.

This PR works in tandem with https://github.com/openshift/cluster-etcd-operator/pull/1605

Changes:
- Add is_single_node() helper function for clean single-node detection
- Handle single-node mode in peer management (clean up learners, set standalone_node)
- Set standalone_node during single-node start to mark standalone state
- Fix monitor path to set standalone_node during two-node to single-node transitions
- Clean up learner members during 2→1 node transitions in monitor path
- Fix restart failures by skipping revision comparison when no peer exists
- Ensure standalone_node attribute is properly maintained across all transitions

Fixes restart failures, 2→1 node transitions, and initial single-node deployments.

Co-Authored-By: Claude Sonnet 4.5 <noreply@anthropic.com>

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1300,6 +1300,16 @@ clear_standalone_node()
 	crm_attribute --name "standalone_node" --delete
 }
 
+is_single_node()
+{
+	# Check if this is a single-node cluster by examining node_ip_map
+	# Returns 0 (true) if only one node is configured, 1 (false) otherwise
+	local peer_count
+
+	peer_count=$(echo "$OCF_RESKEY_node_ip_map" | tr ';' '\n' | grep -vF "$NODENAME" | wc -l)
+	[ "$peer_count" -eq 0 ]
+}
+
 
 # Promotes an etcd learner member to a voting member
 # Args: $1 - learner member ID in decimal format
@@ -1517,6 +1527,38 @@ manage_peer_membership()
 	local peer_member_name
 	local peer_member_ip
 	local peer_member_id
+
+	# Check if this is a single-node cluster
+	if is_single_node; then
+		# Single-node cluster - no peer to manage
+		ocf_log info "Single-node cluster detected (node_ip_map: '$OCF_RESKEY_node_ip_map')"
+
+		# Clean up any learners from the member list (could be left over from 2→1 transition)
+		local learner_member_id
+		learner_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .isLearner==true ).ID")
+		if [ -n "$learner_member_id" ]; then
+			ocf_log warn "Single-node cluster has learner member $learner_member_id - removing to ensure clean state"
+			local learner_member_id_hex
+			if learner_member_id_hex=$(decimal_to_hex "$learner_member_id"); then
+				local endpoint_url
+				endpoint_url=$(ip_url $(attribute_node_ip get))
+				if podman exec "${CONTAINER}" etcdctl --endpoints="$endpoint_url:2379" member remove "$learner_member_id_hex" 2>&1; then
+					ocf_log info "Successfully removed learner member $learner_member_id_hex"
+				else
+					ocf_log warn "Failed to remove learner member $learner_member_id_hex"
+				fi
+			else
+				ocf_log err "Could not convert learner member ID to hex: $learner_member_id"
+			fi
+		fi
+
+		# Ensure standalone_node is set (may have been cleared during 2-node operation)
+		if ! is_standalone; then
+			ocf_log info "Setting standalone_node for single-node cluster"
+			set_standalone_node
+		fi
+		return $OCF_SUCCESS
+	fi
 
 	# Get peer node name and IP
 	peer_ip_map_entry=$(echo "$OCF_RESKEY_node_ip_map" | tr ';' '\n' | grep -vF "$NODENAME")
@@ -2110,16 +2152,26 @@ podman_start()
 				start_resources_count=$(echo "$OCF_RESKEY_CRM_meta_notify_start_resource" | wc -w)
 				ocf_log info "found '$start_resources_count' starting etcd resources (meta notify environment variable: '$OCF_RESKEY_CRM_meta_notify_start_resource')"
 
-				# we need to compare the revisions in any of the following branches
-				# so call the function only once here
-				if ! revision_compare_result=$(compare_revision); then
-					ocf_log err "could not compare revisions, error code: $?"
-					return "$OCF_ERR_GENERIC"
+				# Single-node clusters can't compare revisions with a peer
+				if is_single_node; then
+					ocf_log info "Single-node cluster: starting normally without revision comparison"
+					# No need to set force_new_cluster or compare revisions
+				else
+					# we need to compare the revisions in any of the following branches
+					# so call the function only once here
+					if ! revision_compare_result=$(compare_revision); then
+						ocf_log err "could not compare revisions, error code: $?"
+						return "$OCF_ERR_GENERIC"
+					fi
 				fi
 				case "$start_resources_count" in
 				1)
 					ocf_log debug "peer not starting: ensure we can start a new cluster"
-					if [ "$revision_compare_result" != "older" ]; then
+					if is_single_node; then
+						# Single-node cluster: start normally without revision comparison
+						ocf_log info "Single-node cluster: starting normally"
+						set_standalone_node
+					elif [ "$revision_compare_result" != "older" ]; then
 						# If our revision is the same as or newer than the peer's last saved
 						# revision, and the peer agent isn't currently starting, we can
 						# restore e-quorum by forcing a new cluster.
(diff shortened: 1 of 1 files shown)


jaypoulz:
Proof of the rabbit:
```
(default) jpoulin@fedora:~/Projects/jaypoulz/resource-agents$ coderabbit review --base 1fea31887ae44d0fda288bed1d7d225c9895318c
╔═══════════════════════════════════════════╗
║                                           ║
║   New update available! 0.4.3 -> 0.4.4    ║
║          Run: coderabbit update           ║
║                                           ║
╚═══════════════════════════════════════════╝

Starting CodeRabbit review in plain text mode...

Review directory: /home/jpoulin/Projects/jaypoulz/resource-agents

Connecting to review service
Setting up
Summarizing
Reviewing

Review completed: No findings ✔
```

jaypoulz:
Rabbit is still pleased with me
```
(default) jpoulin@fedora:~/Projects/jaypoulz/resource-agents$ coderabbit review --base 432f5018c69845bcccefad1a831207857aa78442

────────────────────────────────────────
CodeRabbit CLI 0.6.1
What's new

Light mode   : coderabbit review --light for faster local review iteration
Auth         : auth status shows active org and rate-limit context; sign-in prompts are clearer
Review       : legacy review flows now point to plain review or coderabbit review --agent
Fix          : Ctrl-C exits review/auth cleanup promptly
Git          : resolves remotes across SSH aliases, self-hosted hosts, and worktree checkouts
────────────────────────────────────────


   !   jaypoulz/resource-agents is not connected to a CodeRabbit organization
       you can access, so this review will use the free CLI allowance.

       If you expected an org plan here, check your signed-in account with
       coderabbit auth status, or install CodeRabbit for
       jaypoulz/resource-agents at https://app.coderabbit.ai

────────────────────────────────────────
CodeRabbit Review

Diff      : all local changes (committed + uncommitted)
Compare   : single-node-support → 432f5018c69845bcccefad1a831207857aa78442
Directory : resource-agents
────────────────────────────────────────

(\(\
(• .•)  Your code has 99 problems and bug fixes ain't none of 'em.

✔ Preparing review
✔ Summarizing changes

────────────────────────────────────────
Review complete
No findings ✔
────────────────────────────────────────
```

heartbeat/podman-etcd
```diff
@@ -1267,6 +1267,16 @@ clear_standalone_node()
 	crm_attribute --name "standalone_node" --delete
 }
 
+is_single_node()
+{
+	# Check if this is a single-node cluster by examining node_ip_map
+	# Returns 0 (true) if only one node is configured, 1 (false) otherwise
+	local peer_count
+
+	peer_count=$(echo "$OCF_RESKEY_node_ip_map" | tr ';' '\n' | grep -vF "$NODENAME" | wc -l)
```
jaypoulz: I chose to check on the node_ip map since it's always changed AFTER the pacemaker cluster has performed and add/remove operation successfully.

If you add nodes but haven't yet updated the node_ip map, it's safe to proceed because is single node will still be operating as true (from the IP list), so your cluster is still functioning only if the standalone node is the one starting/running.

If you remove nodes, you behave the same way you did before this patch if your pacemaker cluster was out of sync with the node IP map.

There is an argument to be made here that we should check *both* the node IP map and the cluster configuration and only claim is single node if both are true. This is potentially safer, but then we need to define behavior for one of those being true without the other. In practice, this tests well and I haven't seen cases of skew like that.

jaypoulz: If we do decide to enforce *both* properties, we need to ensure that we still allow a standalone node to start because the only entity that can reconcile that state mismatch is the update-setup job in the cluster itself.

heartbeat/podman-etcd
```diff
@@ -1267,6 +1267,16 @@ clear_standalone_node()
 	crm_attribute --name "standalone_node" --delete
 }
 
+is_single_node()
```
fonta-rh, line 1270: This function guards `manage_peer_membership` and `podman_start`, but `container_health_check` is unguarded.

When a single-node container crashes, `container_health_check` finds no `force_new_cluster` holders (no peer to set it) and returns `"failed-wait-for-peer"`. `podman_monitor` maps that to `OCF_SUCCESS` (line 1797) — Pacemaker sees a healthy resource while etcd is permanently down.

The two-node recovery model assumes a peer will call `detect_cluster_leadership_loss` and set `force_new_cluster`. With zero peers, nobody triggers recovery. `time_since_heartbeat` (line 1760) is computed but only logged, never used as a timeout.

Suggested fix — after the `fnc_holder_count` check (line 1771):
```bash
if [ "$fnc_holder_count" -eq 0 ] && is_single_node; then
    ocf_log info "Single-node cluster: self-recovering without peer"
    set_force_new_cluster
    echo "failed-restart-now"
    return
fi
```

heartbeat/podman-etcd
```diff
@@ -2144,16 +2186,31 @@ podman_start()
...
-				if ! revision_compare_result=$(compare_revision); then
-					ocf_log err "could not compare revisions, error code: $?"
-					return "$OCF_ERR_GENERIC"
+				# Standalone node doesn't need peer comparison (matches active_resources_count=1 logic)
+				# Check this before is_single_node to handle drift reconciliation where standalone_node
+				# is set but node_ip_map may already include the peer being added
+				if is_standalone; then
+					ocf_log info "standalone node: skip peer comparison during recovery"
+					# Skip peer comparison - will determine force_new_cluster or normal start below
+				elif is_single_node; then
```
fonta-rh, line 2241: When `is_standalone` is true (line 2192) or `is_single_node` is true here, `compare_revision` is skipped and `revision_compare_result` stays empty. The `start_resources_count=1` branch works by accident (empty `!= "older"` → `set_force_new_cluster`), but the `start_resources_count=2` branch has no equivalent guard.

If both nodes crash simultaneously with a stale `standalone_node` attribute, the empty variable fails both the `"newer"` (line 2227) and `"older"` (line 2229) checks, falling to the `cluster_id` comparison. If IDs match, both nodes "start normal" — neither forces a new cluster. Narrow reachability (stale attribute + double crash), but worth hardening.

Suggested fix — add guards at the top of the `2)` case:
```bash
2)
    ocf_log info "peer starting"
    if is_single_node; then
        ocf_log info "Single-node cluster: starting normally"
        set_standalone_node
    elif is_standalone; then
        ocf_log info "standalone node with peer restarting: forcing new cluster"
        set_force_new_cluster
    elif [ "$revision_compare_result" = "newer" ]; then
        ...
```

heartbeat/podman-etcd
jaypoulz: I don't love this idea. I don't think single node should ever force a new cluster. If it is a valid configuration, it can start normally. If it crashes, I'd rather have it  set standalone node (if it doesn't already have it) and restart rather than it forcing a new cluster.

If it's permacrashing because it had a peer and the history is messed up, it's far safer to force the second node to be readded so that we can compare them and make a  decision about which to force-new-cluster with.

heartbeat/podman-etcd
jaypoulz: By definition, 2 starting resources means you aren't a single-node cluster. :P
I would rather not change the existing logic for non-single node clusters in this patch.

heartbeat/podman-etcd
fonta-rh: Ok, that seems fair to me, FNC is kind of a sledgehammer, but how is the recovery actually happening? When you say "force the second node to be readded", is that manual action?

We're depending on the 2 node recovery mechanism but it won't work on 1 node mode. We would need something like this, based on your suggestion of "standalone + restart"

  if is_single_node; then
      ocf_log info "Single-node cluster: no peer to wait for, setting standalone_node for restart"
      if ! is_standalone; then
          set_standalone_node
      fi
      echo "failed-restart-now"
      return
  fi


I'll add it as an inline comment exactly there

heartbeat/podman-etcd
fonta-rh: Makes sense. I could look into this when I work on https://redhat.atlassian.net/browse/OCPBUGS-87820, since I'll be looking at double crashes anyway

heartbeat/podman-etcd
fonta-rh, line 1: Based on the discussion above: `container_health_check()` still uses the 2-node peer-driven recovery model — it waits for a peer to set `force_new_cluster`, but with zero peers, nobody triggers recovery. `podman_monitor` returns `OCF_SUCCESS` on `"failed-wait-for-peer"`, so Pacemaker never restarts etcd.

Suggested fix — add this after the `fnc_holder_count` check (after line 1776, before `echo "failed-wait-for-peer"`):

```suggestion
	if is_single_node; then
		ocf_log info "Single-node cluster: no peer to wait for, setting standalone_node for restart"
		if ! is_standalone; then
			set_standalone_node
		fi
		echo "failed-restart-now"
		return
	fi

	echo "failed-wait-for-peer"
```

This uses `set_standalone_node` (not `force_new_cluster`) per your feedback, and composes with the `is_standalone` guard already in `podman_start` (line 2192).

heartbeat/podman-etcd
fonta-rh: Well, can't add it as inline, so it's a top level comment :/  Check it out though, I still think we don't want to leave this open

jaypoulz: Yeah this design is safe I think.

jaypoulz: I ended up hardening this function. Yet to be tested. :)

heartbeat/podman-etcd
jaypoulz: Updated!

fonta-rh:
/approve
Nothing to say about this version, let's ship it!

oalbrigt:
Thanks.

jaypoulz:
@oalbrigt thanks for merging!
I need this in RHEL in the 4.22 and 5.0 releases. This means 10.2 and 9.8. :)

PR #2156, opened 2026-05-05
OCPBUGS-82190: podman-etcd: Fixed etcd learner deadlock during machine deletion

vimauro:
### Summary
When a Machine is being deleted in TNF, the `cluster-etcd-operator` (CEO) keeps the EtcdQuorumOperator preDrain hook active as long as the peer IP appears in the etcd member list - including learners. `podman-etcd` resource agent was unconditionally adding/keeping peers as learners, which prevented CEO from clearing the preDrain hook, causing a deadlock.

###  Changes

- Added `is_peer_machine_deleting() ` function - queries the Machine API via `oc` to check whether a peer node's Machine has a `deletionTimestamp` set. Fails open (returns false on API errors) to preserve existing learner-addition behavior when the API is unavailable.

- Guard learner addition in `manage_peer_membership()` - when a peer is not in the member list, skip adding it as a learner if its Machine is being deleted.

- Remove stale learners in `manage_peer_membership()` - if a learner was added before the Machine deletion started, actively remove it and clear the `learner_node` attribute so CEO can clear the preDrain hook.

- Guard learner addition in `podman_start()` - same deletion check applied during the start/bootstrap path.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1530,7 +1530,7 @@ is_peer_machine_deleting()
 	local item_count
 
 	# Try MAPI first (machine.openshift.io), fall back to CAPI (cluster.x-k8s.io)
-	out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.machine.openshift.io \
+	out=$(timeout 5 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.machine.openshift.io \
 		-n openshift-machine-api -o json 2>&1)
 	oc_rc=$?
 
@@ -1607,6 +1607,7 @@ manage_peer_membership()
 	# NOTE: voting members have a "name" field but no "isLearner" field,
 	# while learner members have "isLearner": true (boolean) but no "name" field, so we search for peerURLs matching.
 	peer_member_id=$(printf "%s" "$member_list_json" | jq -r ".members[] | select( .peerURLs | map(test(\"$peer_member_ip\")) | any).ID")
+	
 	# During Machine deletion, CEO's MachineDeletionHooksController
 	# keeps the EtcdQuorumOperator preDrain hook as long as the peer IP appears in the etcd
 	# member list (learners included). If we add or keep a learner for a peer whose Machine
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1541,9 +1541,10 @@ is_peer_machine_deleting()
 	# MAPI CRD missing, namespace absent, or no Machine objects — try CAPI
 	if [ $oc_rc -ne 0 ] || [ "${item_count:-0}" -eq 0 ]; then
 		ocf_log info "MAPI returned no machines, trying CAPI for node $node_name"
-		out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.cluster.x-k8s.io \
+		out=$(timeout 5 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.cluster.x-k8s.io \
 			-n openshift-cluster-api -o json 2>&1)
-		if [ $? -ne 0 ]; then
+		oc_rc=$?
+		if [ $oc_rc -ne 0 ]; then
 			ocf_log warn "could not query Machine API (MAPI or CAPI) for node $node_name (fail-open): $out"
 			return 1
 		fi
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -295,6 +295,7 @@ Set max_backup_snapshots=0 to disable backups.
 Path to a kubeconfig file for querying Machine API objects. Used to detect
 whether a peer node's Machine is being deleted, preventing the resource agent
 from re-adding it as an etcd learner during Machine deletion flows.
+Supports both MAPI (machine.openshift.io) and CAPI (cluster.x-k8s.io) Machine resources.
 </longdesc>
 <shortdesc lang="en">Kubeconfig for Machine API queries</shortdesc>
 <content type="string" default="${OCF_RESKEY_kubeconfig_default}"/>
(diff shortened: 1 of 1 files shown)


jaypoulz (commented):
Overall, I think this works quite cleanly.
A few concerns I have - the kubeconfig we'd be using is set by the API-server. Does it have permissions? Have we confirmed that it is mounted on the podman-etcd container?
Is there a security risk we're introducing by mounting the kube-config into the podman-etcd container? These containers are manipulated by very privileged jobs in etcd. Ideally, we would restricted this kubeconfig to read-only access to the things we're checking.

Lastly, the question is - does it work? I assume you've tested this. If it does, we should kick off some builds and our resource-agents testing jobs to very that OpenShift remains stable with these changes.

heartbeat/podman-etcd
```diff
@@ -1505,6 +1517,34 @@ detect_cluster_leadership_loss()
...
+# Checks whether the Machine object for a given node is being deleted.
+# Returns 0 (true) if the Machine has a deletionTimestamp set, 1 (false) otherwise.
+# Fails open: returns 1 on API errors to preserve current learner-addition behavior.
+is_peer_machine_deleting()
+{
+	local node_name="$1"
+	local out
+	local deletion_ts
+
+	out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines \
```
dhensel-rh: @vimauro I think this line should look like this

```suggestion
	out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.machine.openshift.io \
```

vimauro: Thanks Doug for flagging this. I have pushed a separate change that adds support for both CAPI and MAPI machine CRDs

vimauro:
> Overall, I think this works quite cleanly. A few concerns I have - the kubeconfig we'd be using is set by the API-server. Does it have permissions? Have we confirmed that it is mounted on the podman-etcd container? Is there a security risk we're introducing by mounting the kube-config into the podman-etcd container? These containers are manipulated by very privileged jobs in etcd. Ideally, we would restricted this kubeconfig to read-only access to the things we're checking.
> 
> Lastly, the question is - does it work? I assume you've tested this. If it does, we should kick off some builds and our resource-agents testing jobs to very that OpenShift remains stable with these changes.

Yes It has permissions to read the kubeconfig file as podman-etcd runs as `root`  . This has been tested, however I fully agree with your concern about this extra dependency and the security aspect. Let's discuss it offline and agree on steps forward :)

vimauro:
All fencing recovery tests succeeded https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_release/78919/rehearse-78919-periodic-ci-openshift-release-main-nightly-4.22-e2e-metal-ovn-two-node-fencing-recovery-repo-ra/2052297800137314304

dhensel-rh:
## Test Results — OCP 4.22.0-rc.2 Two-Node Fencing Cluster

Successfully verified this PR fixes the etcd learner deadlock during Machine deletion.

### Test Procedure
1. Deployed clean two-node fencing cluster (OCP 4.22.0-rc.2)
2. Set `pcmk_off_action=off` and `pcmk_reboot_action=off` on stonith resources
3. Powered off master-1 via `virsh destroy`
4. Pacemaker fenced master-1; podman-etcd RA forced single-member cluster on master-0
5. podman-etcd RA re-added 192.168.111.21 as unstarted learner
6. Deleted BMH `ostest-master-1`, then deleted Machine `ostest-7hpvd-master-1`

### Results — PR 2156 working as designed

Timeline from pacemaker logs:
- `19:55:54` — podman-etcd added master-1 as learner (before Machine deletion)
- `20:00:54` — CEO refused to clear preDrain hook: *"skip removing the deletion hook from machine ostest-7hpvd-master-1 since its member is still present"*
- `20:01:24` — **PR 2156 detected deletion**: *"peer Machine is being deleted, removing learner master-1 from member list"*
- `20:01:47` — CEO cleared the hook: *"successfully removed the deletion hook from machine ostest-7hpvd-master-1 as it doesn't have a member"*
- `20:02:01` — PR 2156 continued guarding: *"peer Machine is being deleted, skipping learner addition for master-1"*

Without this patch, CEO would never clear the preDrain hook because the learner would persist indefinitely — confirmed deadlock scenario.

fonta-rh (commented):
Nice fix — the deadlock mechanism is well-understood and the fail-open design is the right call for a Pacemaker RA. Two small items below, both one-line fixes.

heartbeat/podman-etcd
```diff
@@ -1505,6 +1518,58 @@ detect_cluster_leadership_loss()
...
+is_peer_machine_deleting()
+{
+	local node_name="$1"
+	local out
+	local deletion_ts
+	local oc_rc
+	local item_count
+
+	# Try MAPI first (machine.openshift.io), fall back to CAPI (cluster.x-k8s.io)
+	out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.machine.openshift.io \
```
fonta-rh: **Monitor timeout budget:** The OCF monitor timeout is 25s (line 309), and `monitor_cmd_exec` already documents its 15s worst-case budget (line 388). Each `timeout 10 oc` call here can consume 10s — in the MAPI-fail + CAPI-fallback path that's 20s for `is_peer_machine_deleting` alone, leaving ~5s for everything else.

During Machine deletion the kube-apiserver is often degraded (the node being deleted hosted API server pods), making a 10s timeout realistic. A Pacemaker monitor timeout is treated as resource failure and can trigger fencing — exactly the opposite of what this fix intends.

Suggestion: `timeout 5` per call (10s worst case total). Same change on line 1544.

```bash
# Budget: monitor timeout is 25s. monitor_cmd_exec uses up to 15s.
# 5s per oc call, worst case 10s if both MAPI and CAPI time out.
out=$(timeout 5 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.machine.openshift.io \
```

heartbeat/podman-etcd
```diff
@@ -1505,6 +1518,58 @@ detect_cluster_leadership_loss()
...
+	if [ $oc_rc -eq 0 ]; then
+		item_count=$(printf "%s" "$out" | jq '.items | length' 2>/dev/null)
+	fi
+
+	# MAPI CRD missing, namespace absent, or no Machine objects — try CAPI
+	if [ $oc_rc -ne 0 ] || [ "${item_count:-0}" -eq 0 ]; then
+		ocf_log info "MAPI returned no machines, trying CAPI for node $node_name"
+		out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.cluster.x-k8s.io \
+			-n openshift-cluster-api -o json 2>&1)
+		if [ $? -ne 0 ]; then
```
fonta-rh: **Fragile `$?` — asymmetric with the MAPI pattern above.** The MAPI call correctly captures `oc_rc=$?` (line 1535), but the CAPI call uses bare `$?`. If anyone adds a log line between the `out=` assignment and this `if`, `$?` silently refers to the wrong command.

One-line fix to match the MAPI pattern:

```bash
out=$(timeout 10 oc ... 2>&1)
oc_rc=$?
if [ $oc_rc -ne 0 ]; then
```

heartbeat/podman-etcd
vimauro: Good point, I've reduced it to 5 seconds :)

heartbeat/podman-etcd
vimauro: good catch, this is something I faced in the past PRs too. Thanks !

heartbeat/podman-etcd
```diff
@@ -1505,6 +1518,58 @@ detect_cluster_leadership_loss()
...
+	oc_rc=$?
+
+	if [ $oc_rc -eq 0 ]; then
+		item_count=$(printf "%s" "$out" | jq '.items | length' 2>/dev/null)
+	fi
+
+	# MAPI CRD missing, namespace absent, or no Machine objects — try CAPI
+	if [ $oc_rc -ne 0 ] || [ "${item_count:-0}" -eq 0 ]; then
+		ocf_log info "MAPI returned no machines, trying CAPI for node $node_name"
+		out=$(timeout 10 oc --kubeconfig="$OCF_RESKEY_kubeconfig" get machines.cluster.x-k8s.io \
```
fonta-rh: This still has `timeout 10` — should be `timeout 5` to match the MAPI call above (line 1533). Same timeout budget concern applies here.

heartbeat/podman-etcd
```diff
@@ -1505,6 +1518,58 @@ detect_cluster_leadership_loss()
...
+			-n openshift-cluster-api -o json 2>&1)
+		if [ $? -ne 0 ]; then
```
fonta-rh: The `$?` capture is still bare here. Should match the MAPI pattern on line 1535:

```bash
out=$(timeout 5 oc ... 2>&1)
oc_rc=$?
if [ $oc_rc -ne 0 ]; then
```

fonta-rh (commented):
Thanks for the quick update! The MAPI call looks good with `timeout 5`. Looks like the same two changes were missed on the CAPI call — see inline.

heartbeat/podman-etcd
vimauro: Ops, nice catch

heartbeat/podman-etcd
vimauro: for some reason it wasn't pushed earlier, i've did it again. Thanks for spotting this

fonta-rh (approved):
Both items addressed — CAPI call now matches the MAPI pattern with `timeout 5` and explicit `oc_rc=$?` capture. LGTM.

oalbrigt:
Thanks.

PR #2157, opened 2026-05-08
OCPBUGS-83333: podman-etcd: Add learner in notify to prevent start deadlock

vimauro:
- Fix deadlock where a joining node's `podman_start` waits indefinitely for the learner_node CIB
  attribute, which can only be set by the peer's `podman_monitor` but Pacemaker suppresses monitor
  during the active start/notify cycle
  - Add a `check_peer` call in `podman_notify()` on `pre_notify_start`, so the running peer adds the joining
  node as a learner before the start action begins

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -2519,7 +2519,27 @@ podman_validate()
 
 podman_notify()
 {
-	ocf_log info "notify: type=${OCF_RESKEY_CRM_meta_notify_type}, operation=${OCF_RESKEY_CRM_meta_notify_operation}, nodes { active=[${OCF_RESKEY_CRM_meta_notify_active_uname}], start=[${OCF_RESKEY_CRM_meta_notify_start_uname}], stop=[${OCF_RESKEY_CRM_meta_notify_stop_uname}] }, resources { active=[${OCF_RESKEY_CRM_meta_notify_active_resource}], start =[${OCF_RESKEY_CRM_meta_notify_start_resource}], stop=[${OCF_RESKEY_CRM_meta_notify_stop_resource}] }"
+	local notify_type="${OCF_RESKEY_CRM_meta_notify_type}"
+	local notify_operation="${OCF_RESKEY_CRM_meta_notify_operation}"
+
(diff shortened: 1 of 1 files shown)


vimauro:
Fix validated successfully on a cluster running 4.22

fonta-rh (commented):
Nice fix — the deadlock analysis is spot on, and reusing `check_peer` keeps it minimal.

One suggestion on the logging: the current message doesn't identify *which* node is starting, and if `check_peer` fails silently the operator has to correlate logs across two nodes to understand why the joiner timed out. Consider:

```bash
if [ "$notify_type" = "pre" ] && [ "$notify_operation" = "start" ]; then
    ocf_log info "pre_notify_start: running peer membership check for ${OCF_RESKEY_CRM_meta_notify_start_uname}"
    if ! check_peer; then
        ocf_log warn "pre_notify_start: check_peer failed; ${OCF_RESKEY_CRM_meta_notify_start_uname} may time out waiting for learner_node attribute"
    fi
fi
```

This makes the log self-contained for post-incident analysis and gives a direct breadcrumb from the notify to the eventual start timeout on the joiner.

vimauro:
> Nice fix — the deadlock analysis is spot on, and reusing `check_peer` keeps it minimal.
> 
> One suggestion on the logging: the current message doesn't identify _which_ node is starting, and if `check_peer` fails silently the operator has to correlate logs across two nodes to understand why the joiner timed out. Consider:
> 
> ```shell
> if [ "$notify_type" = "pre" ] && [ "$notify_operation" = "start" ]; then
>     ocf_log info "pre_notify_start: running peer membership check for ${OCF_RESKEY_CRM_meta_notify_start_uname}"
>     if ! check_peer; then
>         ocf_log warn "pre_notify_start: check_peer failed; ${OCF_RESKEY_CRM_meta_notify_start_uname} may time out waiting for learner_node attribute"
>     fi
> fi
> ```
> 
> This makes the log self-contained for post-incident analysis and gives a direct breadcrumb from the notify to the eventual start timeout on the joiner.

I've added the extra logging for the node name but I haven't added the new IF condition as it may requires some other code changes in the return codes of check_peer and might be a bit out of scope of this PR (by design, check_peer return OCF_SUCCESS)

fonta-rh (commented):
Thanks for picking up the logging suggestion — the starting node name is a nice improvement.

Heads up: the latest commit (`0fafab70`) appears to have truncated the last line of the file:

```diff
-exit $rc
+exc
```

This would break the agent — `exc` hits `command not found` (exit 127), which Pacemaker interprets as `OCF_ERR_INSTALLED`. Likely an editor/rebase artifact. Should be a quick fix to restore `exit $rc`.

oalbrigt:
Thanks.

PR #2162, opened 2026-05-21
Mid: nfsserver: Change the log level for probe and process checks at startup to info.

HideoYamauchi:
Hi All,

An ERROR message is being output to the Probe and Start logs for the nfsserver resource.
This fix changes the log level to info.

```
May 21 10:40:54 rh101-nfc1 <daemon.err> nfsserver(nfsserver)[142883]:ERROR: nfs-idmapd is not running
May 21 10:41:03 rh101-nfc1 <daemon.err> nfsserver(nfsserver)[143997]:ERROR: nfs-idmapd is not running
```

Best Regards,
Hideo Yamauchi.


oalbrigt:
Thanks.

PR #2165, opened 2026-05-27
nfsserver: Stop fsidd when stopping nfs-server

### LINKED ISSUE #2164: [nfsserver] Failed to unmount a bind mount when stopping
Recently upgraded to Debian 13 which provides now the `fsidd` process. This is a mandatory service started with `nfs-server.service`

`fsidd` is using a sqlite3 database located at /var/lib/nfs/

I was not able to identify when this process access the database as it's does not always keep it open.
Also it seems only to happen when parameter `nfs_shared_infodir` is provided, not 100% sure.

In the situation when it fails to unmount the bind mount, `lsof` tells me:
`fsidd      8542             root   3u   REG  254,0    16384 1310736 /var/lib/nfs/reexpdb.sqlite3`

By adding `nfs_exec stop fsidd > /dev/null 2>&1` in the resource script it seems to be mitigated in our situation.

Please find attached a debug of stopping the service.
[debug-nfsserver.txt](https://github.com/user-attachments/files/28268827/debug-nfsserver.txt)

I've also tested it with the latest version of the resource (commit 530aede)

rayzilt:
Stop fsidd.service as in some cases it can block unmouting bind mounts when using `nfs_shared_infodir`

Fixes: https://github.com/ClusterLabs/resource-agents/issues/2164

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -1010,6 +1010,18 @@ nfsserver_stop ()
 		  		return $OCF_ERR_GENERIC
 		  	fi
 		  fi
+
+          nfs_exec stop fsidd > /dev/null 2>&1
+          ocf_log info "Stop: fsidd"
+          fn=`mktemp`
+          nfs_exec status fsidd > $fn 2>&1
+          rc=$?
+          ocf_log debug "$(cat $fn)"
+          rm -f $fn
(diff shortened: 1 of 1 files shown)


arslan9292:
Tested without issues.

oalbrigt:
Thanks.

PR #2169, opened 2026-06-02
ganesha-nfs: add generated mode (render config from parameters)

yusufyildiz:
## What

Adds a **generated mode** to `ocf:heartbeat:ganesha-nfs` so the agent can realise its export(s)
from its own OCF parameters at runtime, instead of only being able to run a pre-existing
`config_file`.

## Why

As merged (#2154), the agent runs a `ganesha.conf` that must already exist. A cluster manager
that drives NFS-Ganesha (e.g. linstor-gateway) then has to write `ganesha.conf` out-of-band on
every node — there is no way to hand the export intent to the agent. This is the same gap that
`ocf:heartbeat:exportfs` avoids by taking the export as attributes and programming the kernel
table at runtime.

## What changes

Fully backward compatible:

- `config_file` points at an existing file and `export_path` is unset → behaviour unchanged.
- `export_path` is set → the agent renders an **ephemeral** `ganesha.conf` under `runtime_dir`
  (default `/run/ganesha`) from its parameters and starts ganesha with `-f`. Nothing persistent
  is written; on failover the peer re-renders from its own parameters.

New generated-mode parameters: `export_path`, `export_id`, `pseudo`, `clients`, `access_type`,
`squash`, `sectype`, `protocols`, `fsal`, `recovery_dir`, `runtime_dir`.

The export uses a **deny-default** policy — `Access_Type = None` at the export level plus a
`CLIENT {}` whitelist — so an unlisted client is refused. `clients` is therefore required in
generated mode (empty is rejected by `validate`).

### Multiple exports from one process

One `ganesha.nfsd` serves any number of `EXPORT {}` blocks. `export_path`, `export_id` and
`pseudo` accept `;`-separated lists, zipped positionally — one EXPORT block per entry behind a
single service IP and port (distinguished by NFSv4 pseudo path, not by port). All other
parameters (clients, squash, RecoveryDir, …) are shared. A single value still means a single
export. `export_id` defaults to the 1-based position, `pseudo` to the path; `validate` checks the
list lengths match. This lets a multi-volume cluster resource map to one ganesha process instead
of needing one process (and IP) per export.

## Testing

On a 3-node AlmaLinux 9.7 / NFS-Ganesha V5.9 cluster:

- `ocf-tester` passes (single export).
- A two-export resource renders two `EXPORT {}` blocks that both mount and serve from one daemon
  on one IP/port.
- Runs under drbd-reactor; `drbd-reactorctl evict` failover keeps the export available, with the
  NFSv4 RecoveryDir on the replicated filesystem so client lock state migrates.
- An allowed client mounts read/write; an unlisted client is refused.

## First commit: the IPv4-mapped listener fix

The first commit is a one-line fix so the readiness probe matches IPv4-mapped IPv6 listeners
(`::ffff:…`) — generated mode is typically used with an IPv4 service IP and needs it. It was also
posted standalone as #2167, now closed in favour of this PR so the agent has a single PR. It is
independent of generated mode and can be reviewed first if preferred.

### DIFF
--- a/heartbeat/ganesha-nfs
+++ b/heartbeat/ganesha-nfs
@@ -194,6 +194,7 @@ port_listening() {
         ss -Hln -tnp "sport = :${port}" 2>/dev/null | \
             awk -v ip="$ip" '
                 { addr=$4; sub(/:[^:]+$/,"",addr); gsub(/[\[\]]/,"",addr);
+                  sub(/^::ffff:/,"",addr);
                   if (addr==ip || addr=="0.0.0.0" || addr=="*" || addr=="::" || addr=="") { found=1 } }
                 END { exit(found?0:1) }'
         return $?
--- a/heartbeat/ganesha-nfs
+++ b/heartbeat/ganesha-nfs
@@ -79,7 +79,7 @@ OCF_RESKEY_fsal_default="VFS"
 #######################################################################
 # Mode + path helpers
 
-# Generated mode iff an export_path was supplied.
+# Generated mode if and only if an export_path was supplied.
 generated_mode() {
     [ -n "$OCF_RESKEY_export_path" ]
 }
@@ -247,7 +247,7 @@ resource fails over with, or clients lose their locks on failover.
 <parameter name="export_path" unique="0" required="0">
 <longdesc lang="en">Server-side directory to export, on the replicated FS. Setting this switches the agent into generated mode (it renders ganesha.conf from its parameters). May be a ';'-separated list to export several directories from one ganesha process.</longdesc>
 <shortdesc lang="en">Exported directory, or ';'-list (enables generated mode)</shortdesc>
-<content type="string" default="" />
+<content type="string" default="${OCF_RESKEY_export_path_default}" />
 </parameter>
 
 <parameter name="export_id" unique="0" required="0">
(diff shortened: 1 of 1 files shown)


heartbeat/ganesha-nfs
```diff
@@ -30,56 +36,283 @@
...
+: ${OCF_RESKEY_squash=${OCF_RESKEY_squash_default}}
+: ${OCF_RESKEY_sectype=${OCF_RESKEY_sectype_default}}
+: ${OCF_RESKEY_protocols=${OCF_RESKEY_protocols_default}}
+: ${OCF_RESKEY_recovery_dir=${OCF_RESKEY_recovery_dir_default}}
+: ${OCF_RESKEY_fsal=${OCF_RESKEY_fsal_default}}
+
+#######################################################################
+# Mode + path helpers
+
+# Generated mode iff an export_path was supplied.
```
oalbrigt: I guess this was supposed to say "if".

heartbeat/ganesha-nfs
```diff
@@ -30,56 +36,283 @@
...
+<parameter name="clients" unique="0" required="0">
+<longdesc lang="en">Client spec for the CLIENT whitelist (e.g. "10.1.10.0/24", a bare IPv6 like fd00:10:1:10::3, or "*"). REQUIRED in generated mode (deny-default: no clients means nobody can mount). Use "*" explicitly to allow all reachable clients.</longdesc>
+<shortdesc lang="en">Allowed clients (required in generated mode)</shortdesc>
+<content type="string" default="" />
+</parameter>
+
+<parameter name="access_type" unique="0" required="0">
+<longdesc lang="en">Access type granted to whitelisted clients: RW, RO, or None.</longdesc>
+<shortdesc lang="en">Access type</shortdesc>
+<content type="string" default="RW" />
```
oalbrigt: You should use _default values for all these, so you dont need to change it later if the default change.

oalbrigt (changes requested):
Just a couple of improvements needed, and it should be ready to merge.

heartbeat/ganesha-nfs
yusufyildiz: Good catch — it was meant as the "if and only if" shorthand, but you're right it reads as a typo. Reworded to "if and only if" in b0216062.

heartbeat/ganesha-nfs
yusufyildiz: Done in b0216062 — every `<content default=...>` now references its `OCF_RESKEY_*_default` variable (export_path, pseudo, clients, access_type, squash, sectype, protocols, fsal, recovery_dir, and nfs_ip which had the same literal; export_id already did this), so the advertised defaults can't drift from the code defaults.

oalbrigt (approved):

oalbrigt:
Thanks.

PR #2174, opened 2026-06-08
Low: nfsserver: Correction of magic numbers and log output.

HideoYamauchi:
Hi All,

There was a minor issue with the nfsserver, and I would like to request a fix.

Best Regards,
Hideo Yamauchi.


oalbrigt:
Thanks.

PR #2179, opened 2026-06-11
nfsserver: Monitor nfsdcld and nfs-mountd services to trigger recovery on failure

arslan9292:
This PR enhances the `nfsserver` resource agent by adding explicit monitoring for two critical dependent systemd services: `nfsdcld.service` and `nfs-mountd.service`.

While `nfsdcld` and `nfs-mountd` are automatically started and stopped alongside `nfs-server.service` via systemd dependencies, their underlying processes can still crash or terminate unexpectedly. Previously, the agent would not detect these individual service failures. Specifically, `rpc.mountd` is critical for handling export authentication across all NFS versions.

The Solution:
During the monitor operation, the agent now explicitly checks the state of these services:
- Monitors `nfsdcld.service` for stopped or failed states.
- Monitors `nfs-mountd.service` for failed states.

This ensures that if either service fails, the agent will successfully detect it and trigger the appropriate recovery actions.

### DIFF
--- a/heartbeat/nfsserver
+++ b/heartbeat/nfsserver
@@ -370,18 +370,18 @@ nfsserver_systemd_monitor()
 			fi
 			return $OCF_NOT_RUNNING
 		fi
+	fi
 
-		ocf_log debug "Status: nfs-mountd"
-		ps axww | grep -q "[r]pc.mountd"
-		rc=$?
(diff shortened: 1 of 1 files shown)


oalbrigt:
LGTM.

oalbrigt:
Thanks.

PR #2183, opened 2026-06-21
IPaddr2: replace clusterip with cluster module

vvidic:
clusterip module was removed from kernel in 2023, see https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=9db5d918e2c07fa09fab18bc7addf3408da0c76f

### DIFF
--- a/heartbeat/IPaddr2
+++ b/heartbeat/IPaddr2
@@ -51,7 +51,6 @@
 #	OCF_RESKEY_cidr_netmask
 #	OCF_RESKEY_iflabel
 #	OCF_RESKEY_mac
-#	OCF_RESKEY_clusterip_hash
 #	OCF_RESKEY_arp_interval
 #	OCF_RESKEY_arp_count
 #	OCF_RESKEY_arp_bg
@@ -77,7 +76,11 @@ OCF_RESKEY_iflabel_default=""
 OCF_RESKEY_lvs_support_default=false
 OCF_RESKEY_lvs_ipv6_addrlabel_default=true
 OCF_RESKEY_lvs_ipv6_addrlabel_value_default=99
-OCF_RESKEY_clusterip_hash_default="sourceip-sourceport"
+OCF_RESKEY_clusterip_firewall_default="auto"
+OCF_RESKEY_clusterip_hash_seed_default=""
+OCF_RESKEY_clusterip_table_default="mangle"
+OCF_RESKEY_clusterip_chain_default="PREROUTING"
+OCF_RESKEY_clusterip_mac_mangle_default=false
 OCF_RESKEY_mac_default=""
 OCF_RESKEY_unique_clone_address_default=false
 OCF_RESKEY_arp_interval_default=200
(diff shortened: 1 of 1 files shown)


oalbrigt:
Nice.

I guess we'll have to add nftables support as well, as iptables will get deprecated in favor of it in the future.

vvidic:
> I guess we'll have to add nftables support as well, as iptables will get deprecated in favor of it in the future.

Ok, added nft commands now, it seems to be working.

heartbeat/IPaddr2
```diff
@@ -317,13 +315,49 @@ the Cluster IP Alias. Leave empty to chose automatically.
...
+If true, use arptables to mangle ARP responses so that the cluster IP
+address resolves to the multicast MAC address. Not supported for IPv6.
 </longdesc>
-<shortdesc lang="en">Cluster IP hashing function</shortdesc>
-<content type="string" default="${OCF_RESKEY_clusterip_hash_default}"/>
+<shortdesc lang="en">Mangle ARP responses for cluster MAC</shortdesc>
+<content type="boolean" default="${OCF_RESKEY_cluster_mac_mangle_default}"/>
+</parameter>
+
+<parameter name="cluster_firewall" unique="0" required="0">
```
oalbrigt: I would move this to the top of cluster IP parameter list and set nft or detect (where it would use nft if available and fallback to iptables) as default .

heartbeat/IPaddr2
```diff
@@ -317,13 +315,49 @@ the Cluster IP Alias. Leave empty to chose automatically.
 <content type="string" default="${OCF_RESKEY_mac_default}"/>
 </parameter>
 
-<parameter name="clusterip_hash">
+<parameter name="cluster_hash_seed">
```
oalbrigt: I think we should update `cluster_` to `cluster_ip_` for the parameters, as the entire agent is cluster related.

heartbeat/IPaddr2
```diff
@@ -42,7 +42,7 @@
 # - Make this the standard IP address agent on Linux; the other
 #   platforms simply should ignore the additional parameters OR can use
 #   the legacy heartbeat resource script...
-# - Check LVS <-> clusterip incompatibilities.
+# - Check LVS <-> cluster incompatibilities.
```
oalbrigt: This should probably say cluster (IP) module or something along those lines.

heartbeat/IPaddr2
```diff
@@ -624,7 +658,16 @@ ip_init() {
 
 	if [ "$IP_INC_GLOBAL" -gt 1 ] && ! ocf_is_true "$OCF_RESKEY_unique_clone_address"; then
 		IP_CIP="yes"
-		IP_CIP_HASH="${OCF_RESKEY_clusterip_hash}"
+		if [ "$FAMILY" = "inet" ]; then
```
oalbrigt: The iptables-related stuff should be moved to the iptables-part of the `if()`'s

heartbeat/IPaddr2
```diff
@@ -1006,6 +1060,138 @@ run_send_ua() {
 	fi
 }
 
+#
+# Ignore incoming packets not matching current cluster node
+#
+cip_drop() {
+	action="$1"
+
+	if [ "$OCF_RESKEY_cluster_firewall" = "iptables" ]; then
```
oalbrigt: I would reverse these to `if [ "$OCF_RESKEY_cluster_firewall" = "nft" ]`, as nftables is what we'll use mostly in the future.

oalbrigt (changes requested):
Looks good in general.

I've added a few suggestions for improvements.

heartbeat/IPaddr2
vvidic: Moved and added auto default.

heartbeat/IPaddr2
vvidic: Renamed parameters.

heartbeat/IPaddr2
vvidic: Done.

heartbeat/IPaddr2
vvidic: Moved.

heartbeat/IPaddr2
vvidic: Changed order to prefer nft.

heartbeat/IPaddr2
```diff
@@ -1006,6 +1061,135 @@ run_send_ua() {
...
+# Setup cluster IP serving
+#
+cip_setup() {
+	action="$1"
+	failed=""
+
+	cip_drop $action || failed="$failed cip_drop"
+	cip_mangle $action || failed="$failed cip_mangle"
+
+	if [ "$action" = "insert" ]; then
```
oalbrigt, line 1140: Seems like we need a nft section here, or skip this part when using nft.

vvidic: Actually, this is calling the ip command so it is the same calls for both firewall types. Anyways I have updated the function to be a bit cleaner.

oalbrigt: Oh yeah. Seems like I read IP2UTIL as IPTABLES :)

Nice improvement at least.

oalbrigt (approved):

oalbrigt:
Thanks.

PR #2185, opened 2026-06-27
IPaddr2: fix loop variable

vvidic:
$nic variable is not defined anywhere in the file so the command
would list any interface with a matching label.


oalbrigt (approved):

oalbrigt (approved):

oalbrigt:
Thanks.

PR #2187, opened 2026-06-29
oralsnr: enable support for multiple listeners per SID

arslan9292:
Previously, configuring multiple listeners for the same SID caused validation warnings and false positive monitor operations.

One can configure several listeners per instance, one for front-end applications, another for backup etc. To properly support this architecture, this commit introduces the following changes:

* Removed the `unique` constraint on the `sid` parameter, as the `listener` parameter is the true unique identifier for the OS-level process.

* Added a new resource option `tns_svc_name` to allow cluster administrators to specify a custom TNS alias (e.g. pointing to a specific port) for the monitor operation.

* Updated `test_tnsping()` to use the new `tns_svc_name` parameter instead of hardcoded `$ORACLE_SID`. This parameter dynamically defaults to the database SID, ensuring complete backward compatibility for existing single-listener setups.

### DIFF
--- a/heartbeat/oralsnr
+++ b/heartbeat/oralsnr
@@ -25,6 +25,7 @@
 #	OCF_RESKEY_home (optional; else read it from /etc/oratab)
 #	OCF_RESKEY_user (optional; user to run the listener)
 #	OCF_RESKEY_listener (optional; defaults to LISTENER)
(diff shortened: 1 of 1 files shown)


oalbrigt (approved):

oalbrigt:
Thanks.

PR #2189, opened 2026-07-02
nvmet-*: When the nvmet kernel module is not detected, attempt to load it to avoid manual execution.

twtlpla:
When the nvmet kernel module is not detected, attempt to load it to avoid manual execution.


oalbrigt (approved):

oalbrigt:
Thanks.

PR #2191, opened 2026-07-29
mediationzone: new resource agent

ja9fuchs:
The resource agent manages pico processes of DigitalRoute's MediationZone. The agent uses the `mzsh` command to control the configured resources.

Assisted-by: Claude <noreply@anthropic.com>

### DIFF
--- a/heartbeat/mediationzone
+++ b/heartbeat/mediationzone
@@ -86,7 +86,6 @@ mediationzone_meta_data() {
 <!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
 <resource-agent name="mediationzone" version="${mediationzone_version}">
 <version>1.0</version>
-<shortdesc lang="en">Manages a MediationZone pico process.</shortdesc>
 <longdesc lang="en">
 The mediationzone resource agent manages MediationZone (by DigitalRoute)
 pico processes as active/passive cluster resources.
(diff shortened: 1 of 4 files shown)


heartbeat/mediationzone
```diff
@@ -0,0 +1,384 @@
...
+</parameter>
+<parameter name="pico_name" unique="0" required="0">
+    <shortdesc lang="en">Name of the pico to manage.</shortdesc>
+    <longdesc lang="en">Name of the MediationZone pico to manage. Known names with default ports are platform (9000) and ui (9001). Set pico_port explicitly for other pico name and port combinations.</longdesc>
+    <content type="string" default="${OCF_RESKEY_pico_name_default}" />
+</parameter>
+<parameter name="pico_port" unique="0" required="0">
+    <shortdesc lang="en">Web server port of the pico.</shortdesc>
+    <longdesc lang="en">The port the pico web server listens on. Used to clear TIME-WAIT connections before startup. Defaults to 9000 for platform and 9001 for ui. Must be set explicitly when managing a different pico type, or when the default port has been changed in the container configuration.</longdesc>
+    <content type="integer" default="" />
```
oalbrigt: I would use the _defauit variable here as well, in case it changes in the future.

heartbeat/mediationzone
```diff
@@ -0,0 +1,384 @@
...
+    echo "usage: $0 {start|stop|monitor|validate-all|meta-data|methods|usage}"
+}
+
+mediationzone_methods() {
+    echo "start stop monitor validate-all meta-data methods usage"
+}
+
+mediationzone_check() {
+    timeout "$OCF_RESKEY_mzsh_timeout" \
+        su - "$OCF_RESKEY_os_user" -c \
```
oalbrigt: Check `$SU` implementation in e.g. pgsql, so we can use runuser instead of su, when available.

oalbrigt (changes requested):

oalbrigt (approved):

PR #2192, opened 2026-07-31
ibmmq: new resource agent for IBM MQ queue managers

zypek:
Manages a single IBM MQ queue manager in an active/passive Pacemaker deployment (no MQ multi-instance mode; strmqm is never invoked with -x). No IBM MQ agent exists in resource-agents today.

Features:
- start: stale-process sweep (per-family, qmgr-anchored) then strmqm; refuses with OCF_ERR_INSTALLED when the queue manager is not defined on the node (unmounted shared storage / missing addmqinf)
- stop: three-phase escalation endmqm -i -> endmqm -p -> direct termination of the queue manager process family, converging on the execution controller (amqzxma0) exiting, with a final sweep of any family members that outlive it so nothing can pin the queue manager filesystem during failover
- monitor: locale-stable dspmq -n run-state parsing, probe-safe on nodes without MQ or without the shared filesystem mounted; OCF_CHECK_LEVEL 10 adds a runmqsc "ping qmgr" responsiveness check; OCF_CHECK_LEVEL 20 adds an end-to-end put/get probe on a dedicated test queue via dmpmqmsg, catching a queue manager that answers runmqsc but can no longer move messages (full/deadlocked log)
- optional checks: command server process, listener process, custom monitor program; multi-installation support via mq_installation_path and env_file

Behaviour validated on RHEL 8.10 / Pacemaker 2.1.7 / IBM MQ 9.3.0: return-code and probe contract (including OCF_ERR_INSTALLED vs OCF_ERR_CONFIGURED classification), stop escalation under a frozen queue manager, agent SIGKILL mid-stop reinvocation, crash recovery with 10k persistent messages intact, and controlled failover.

### DIFF
--- a/heartbeat/ibmmq
+++ b/heartbeat/ibmmq
@@ -52,6 +52,13 @@
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
 
+# Use runuser if available for SELinux.
+if [ -x /sbin/runuser ]; then
+    SU=runuser
+else
+    SU=su
+fi
+
 # Parameter defaults
 OCF_RESKEY_mquser_default="mqm"
 OCF_RESKEY_mq_installation_path_default=""
@@ -90,7 +97,7 @@ meta_data() {
 <?xml version="1.0"?>
 <!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
 <resource-agent name="ibmmq" version="1.1">
-<version>1.1</version>
+<version>1.0</version>
 
 <longdesc lang="en">
 Resource agent for an IBM MQ queue manager. Starts the queue manager
@@ -143,9 +150,10 @@ running MQ commands. Leave empty to use the primary installation.
 
 <parameter name="env_file" unique="0" required="0">
 <longdesc lang="en">
-Absolute path to a Bourne-shell compatible file sourced before every MQ
-command, e.g. to run setmqenv or export queue manager environment
-variables.
+Absolute path to a file sourced before every MQ command, e.g. to run
+setmqenv or export queue manager environment variables. The file is
+executed by the system's /bin/sh, so it must be compatible with that
+shell (POSIX sh syntax is safe on every distribution).
 </longdesc>
 <shortdesc lang="en">Environment file</shortdesc>
 <content type="string" default="${OCF_RESKEY_env_file_default}" />
@@ -259,7 +267,7 @@ mq_run() {
     if [ "$(id -un)" = "$OCF_RESKEY_mquser" ]; then
         /bin/sh -c "$cmd"
     else
-        su - "$OCF_RESKEY_mquser" -s /bin/sh -c "$cmd"
+        $SU - "$OCF_RESKEY_mquser" -s /bin/sh -c "$cmd"
     fi
 }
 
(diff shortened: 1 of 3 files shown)


heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+amqzmuf0 amqzmur0 amqzmgr0 amqzdmaa"
+
+#######################################################################
+
+meta_data() {
+    cat <<END
+<?xml version="1.0"?>
+<!DOCTYPE resource-agent SYSTEM "ra-api-1.dtd">
+<resource-agent name="ibmmq" version="1.1">
+<version>1.1</version>
```
oalbrigt: Do you need OCF 1.1? If not I would set this to 1.0, so it's compatible with older versions of Pacemaker as well.

heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+When set, PATH is prefixed with this path's bin directory before
+running MQ commands. Leave empty to use the primary installation.
+</longdesc>
+<shortdesc lang="en">MQ installation path</shortdesc>
+<content type="string" default="${OCF_RESKEY_mq_installation_path_default}" />
+</parameter>
+
+<parameter name="env_file" unique="0" required="0">
+<longdesc lang="en">
+Absolute path to a Bourne-shell compatible file sourced before every MQ
```
oalbrigt: Is it Bourne-shell, or could it depend on distro? E.g. dash on Debian-based distros.

heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+command, e.g. to run setmqenv or export queue manager environment
+variables.
+</longdesc>
+<shortdesc lang="en">Environment file</shortdesc>
+<content type="string" default="${OCF_RESKEY_env_file_default}" />
+</parameter>
+
+<parameter name="strmqm_opts" unique="0" required="0">
+<longdesc lang="en">
+Additional options passed to strmqm on start. Do not use -x (multi
```
oalbrigt, line 164: Let's add a check for -x in validate-all.

heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+mq_run() {
+    local cmd envprefix=""
+    [ -n "$OCF_RESKEY_env_file" ] && envprefix=". '$OCF_RESKEY_env_file'; "
+    [ -n "$OCF_RESKEY_mq_installation_path" ] && \
+        envprefix="${envprefix}PATH='$OCF_RESKEY_mq_installation_path/bin':\$PATH; export PATH; "
+    cmd="${envprefix}$*"
+    if [ "$(id -un)" = "$OCF_RESKEY_mquser" ]; then
+        /bin/sh -c "$cmd"
+    else
+        su - "$OCF_RESKEY_mquser" -s /bin/sh -c "$cmd"
```
oalbrigt: Let's use `$SU` like we do in the `pgsql` agent, so we use runuser when it's available.

heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+        ocf_exit_reason "env_file $OCF_RESKEY_env_file is not readable"
+        return $OCF_ERR_INSTALLED
+    fi
+
+    # dspmq must be resolvable in the MQ user's environment. During a
+    # probe the MQ filesystem may legitimately not be mounted here, so
+    # only warn in that case.
+    if ! mq_run "command -v dspmq" >/dev/null 2>&1; then
+        if ocf_is_probe; then
+            ocf_log info "ibmmq: dspmq not found during probe (MQ possibly not available on this node yet)"
```
oalbrigt: I would just change the else to `if ! ocf_is_probe`, as we dont want to spam the log files during probes.

heartbeat/ibmmq
```diff
@@ -0,0 +1,577 @@
...
+        STARTING)
+            # Grace: the queue manager is coming up; report success so
+            # start can converge (per IBM sample behaviour).
+            ocf_log info "ibmmq: queue manager $OCF_RESKEY_qmgr is starting"
+            return $OCF_SUCCESS
+            ;;
+        "RUNNING AS STANDBY")
+            # We never start with -x, so a standby here means somebody
+            # started the qmgr outside cluster control.
+            ocf_log warn "ibmmq: queue manager $OCF_RESKEY_qmgr is running as standby (unexpected in this deployment model)"
```
oalbrigt: I would use `ocf_exit_reason()` here, and for other cases where it fails, as that makes it visible via `pcs status`.

oalbrigt (changes requested):
Just a few changes, and it should be good to go.

oalbrigt (approved):
LGTM.

oalbrigt:
Thanks.

PR #2193, opened 2026-07-31
Route: update `destination` parameter to non-unique

### LINKED ISSUE #2182: Route: "default" destination should not be unique
For dual-stack clustered routers, the cluster surely can have two `ocf:heartbeat:Route` resources with `destination=default`, one with `family=ip4`, and another one with `family=ip6`. Current implementation does not allow this, as `Route` declares the `destination` parameter with `unique=1`:

https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/Route#L111

To reproduce:

```
# pcs resource create defgw6 Route destination=default device=uplink gateway=fe80::1 family=ip6
# pcs resource create defgw4 Route destination=default device=uplink gateway=10.0.0.1 family=ip4
Error: Value 'default' of option 'destination' is not unique across 'ocf:heartbeat:Route' resources. Following resources are configured with the same value of the instance attribute: 'defgw6', use --force to override
```

It is of course possible to use the `--force` option, which converts the above error to a warning, or to use `0.0.0.0/0` or `::/0` instead of one or both `default` values.

arslan9292:
Allows multiple routes to use the string "default" for
different address families (e.g. dual-stack IPv4/IPv6).

Fixes #2182

### DIFF
--- a/heartbeat/Route
+++ b/heartbeat/Route
@@ -88,7 +88,7 @@ of the Route resource agent:
 
 <parameters>
 
-<parameter name="destination" unique="1" required="1">
+<parameter name="destination" unique="0" required="1">
 <longdesc lang="en">
 The destination network (or host) to be configured for the route. 
 Specify the netmask suffix in CIDR notation (e.g. "/24").


oalbrigt:
Thanks.

PR #2194, opened 2026-08-02
OCPBUGS-100303: podman-etcd: resolve force_new_cluster race during simultaneous dual-start

Neilhamza:
## Problem

When both TNF control-plane nodes restart simultaneously (dual power loss, certificate expiry), the podman-etcd resource agent can enter a permanent deadlock. Both nodes set the force_new_cluster CIB attribute due to two race vectors:

1. **CIB replication delay**: Both nodes update their revision in CIB then call compare_revision(), which reads local disk for self but CIB for peer. If the peer's update hasn't propagated, both see stale peer revision and both determine they're "newer."

2. **Inconsistent start_resources_count**: During simultaneous boot, OCF_RESKEY_CRM_meta_notify_start_resource can disagree between nodes, causing both to enter the case 1 path which sets force_new_cluster for any revision that isn't "older" (including "equal").

Once both hold force_new_cluster, the multi-holder safety check returns OCF_ERR_CONFIGURED -- a **fatal error with no automatic retry** -- causing total etcd outage requiring manual intervention.

### Considered alternative

An atomic claim via cibadmin --create (create-fails-if-exists) would eliminate the verdict, guard, and stale-claim machinery in one stroke. However it sacrifices --lifetime reboot auto-cleanup and requires staleness machinery (boot-id, dead-claimant detection) with similar net complexity and higher migration risk for clusters already running with per-node FNC attributes.

## Solution

This PR addresses OCPBUGS-100303 (force_new_cluster race during simultaneous dual-start). Fixing the bug correctly required addressing three additional aspects beyond the race itself: the fatal abort that prevents self-healing, the stale-claim cleanup for banned peers, and the active-peer guard that prevents split-brain.

The PR contains 2 commits:

1. The entire fix: conflict verdict, safety guard, stale-claim cleanup, helpers.
2. Test hooks only (separable -- drop if maintainers prefer keeping test instrumentation downstream).

### Conflict verdict (fnc_conflict_verdict)

When both nodes hold FNC, the verdict function determines the outcome using CIB-sourced revision and cluster_id data (symmetric inputs -- both nodes evaluate the same data):

| Verdict | Condition | Action |
|---------|-----------|--------|
| Winner | Higher CIB revision, or only node with revision | Winner force-new-clusters, loser yields |
| EQUAL | Same revision AND same non-empty cluster_id | Both clear FNC, joint normal start (no learner rebuild) |
| ADMIN | Same revision but different/empty cluster_id, or no revision data | Soft-fail with actual holder names in exit reason; never guess |

Alphabetic node-name guessing was explicitly rejected -- when there is no data to distinguish the nodes, the agent reports the conflict and lets an administrator inspect the etcd data directories rather than risking a wrong choice.

### Post-set safety guard

After the start-flow decision logic but before container creation, a one-shot guard sleeps 3s (load-bearing: lets a peer's in-flight FNC write propagate through attrd/CIB) then verifies this node is the sole visible holder. On any anomaly -- two holders (race), zero holders (own write not yet visible), or sole holder is the peer -- it soft-fails with OCF_ERR_GENERIC and lets Pacemaker retry. The early multi-holder check resolves the conflict on the retry with fresh notify vars.

### Stale-claim cleanup

The data-backed winner in the early multi-holder check clears a stale peer FNC attribute when the peer's resource is neither starting, active, nor has an in-flight start in the CIB. This handles the banned-peer scenario where reboot-lifetime attributes persist until membership loss and no agent ever runs to yield. peer_has_pending_start queries the CIB status section for pending ops (accepts both exit code 105/CRM_EX_NOSUCH for Pacemaker 2.x and 6/ENXIO for legacy 1.x -- both verified empirically on Pacemaker 2.1.10; xpath positive-match against real pending start ops also confirmed). Query errors are treated as "pending" to prevent deletion during CIB instability.

### Other design decisions

- **Active-peer guard checks for voter peers only** (is_peer_active_voter) -- yielding to a learner-only peer would leave the cluster with no voter
- **All error paths return OCF_ERR_GENERIC** (soft, retriable), never OCF_ERR_CONFIGURED (fatal)
- **Dual-claim always logged at err level** -- even when auto-resolution succeeds, the evidence that a startup race occurred is preserved for post-mortem
- **fnc_yield clears standalone marker** -- prevents stale is_standalone from re-triggering FNC on next restart
- **fnc_equal_clear_and_confirm re-checks after clearing** -- handles CIB divergence where one node sees EQUAL while the other sees a winner

### New helper functions

- fnc_conflict_verdict: four-outcome verdict (winner/EQUAL/ADMIN)
- fnc_equal_clear_and_confirm: clear-and-verify for EQUAL outcome
- fnc_admin_escalate: exit reason with holder names and recovery command
- fnc_yield: loser-side cleanup (FNC + standalone marker + JOIN_AS_LEARNER)
- is_peer_active_voter: voter-only active-peer check
- peer_has_pending_start: CIB status xpath for in-flight starts (exit codes 105 + 6)
- is_numeric, get_cib_revision: POSIX-safe revision helpers

OCF_RESKEY_test_race_delay is an intentionally undeclared test hook for deterministic race reproduction (commit 2, separable). Both injection points are inert when the variable is unset.

## What this PR touches beyond the race fix

- **set_force_new_cluster()**: test hook added before the crm_attribute write
- **Early multi-holder check**: OCF_ERR_CONFIGURED (fatal) replaced with verdict dispatch + stale-claim cleanup + OCF_ERR_GENERIC (retriable)
- **Single-holder-is-me path**: active-peer voter guard added
- **Both start-path set_force_new_cluster call sites**: return code now checked; on failure, return OCF_ERR_GENERIC
- **New code between decision logic and container creation**: post-set safety guard (one-shot, +3s on FNC starts only)

## Testing

Live 4.23 TNF cluster (two-node DualReplica, OCP 4.23.0-0.nightly-2026-08-01-032511, Pacemaker 2.1.10, dev-scripts on AWS). Tests run across two cluster deployments. All tests ran without manual intervention -- the cluster self-healed in every scenario. All 35 Cluster Operators verified healthy after each test.

### Test 1: Conflict resolution resolves existing deadlock
- Pre-set FNC on both nodes, stopped both etcd, exited maintenance
- Both nodes detected dual FNC, computed same verdict: master-1 wins (revision 40221 > 40210)
- master-0 yielded (cleared FNC), joined as learner
- master-1 kept FNC through startup, cleared after etcd started successfully (normal post-start clear, not a yield)
- **Exercises**: early multi-holder verdict, fnc_yield(), if/elif/else structural change
- **Result: PASSED** -- self-healed, all COs recovered

### Test 2: Quorum-loss regression (kill peer VM)
- Destroyed master-1 VM via virsh destroy
- master-0's detect_cluster_leadership_loss() fired -> set FNC + standalone
- master-0 restarted etcd with force_new_cluster (+3s guard) -> cleared FNC -> added master-1 as learner after fencing
- master-1 promoted to voter
- **Exercises**: standalone marker lifecycle, post-set safety guard on uncontested FNC start, fencing + learner rejoin
- **Result: PASSED**

### Test 3: Normal single-node failover (fence master-0)
- Destroyed master-0 VM. Same flow as Test 2, reversed roles.
- master-1 survived, set FNC+standalone, master-0 rejoined after fencing
- **Result: PASSED**

### Test 4: Equal-revision EQUAL verdict (joint normal start)
- Both nodes had genuinely equal disk revisions and same cluster_id (natural state on healthy cluster pre-fencing)
- Pre-set FNC on both, stopped etcd, exited maintenance
- Both: "FNC conflict inputs: master-0=52806 master-1=52806" -> both cleared FNC -> "equal revision and cluster_id - joint normal start"
- Neither node rebuilt as learner -- both started normally from existing data
- Disk revisions were 52546 at pre-stop check; the cluster committed between the check and container stops, so the RA read 52806 at start. Post-fencing revision divergence prevented re-running this test on the second cluster; the code and verdict logic have not changed between deployments.
- **Exercises**: fnc_conflict_verdict EQUAL path, fnc_equal_clear_and_confirm, cluster_id comparison
- **Result: PASSED**

### Test 5: ADMIN verdict (code-trace verified)
- The ADMIN path triggers when revision data is genuinely unavailable or cluster_ids diverge after split-brain. podman_start() writes revision and cluster_id from disk before the FNC check, making ADMIN unreachable without actual data corruption.
- Code-trace verified: both no-revision branches and the different-cluster_id branch echo "ADMIN"; callers return OCF_ERR_GENERIC with fnc_admin_escalate showing actual holder names.
- **Result: VERIFIED BY CODE TRACE**

### Test 6: Race window exercised via test hook
- OCF_RESKEY_test_race_delay=8 set on both nodes, stopped both etcd, exited maintenance
- Both hooks fired simultaneously. Revision asymmetry (post-fencing) produced a single-set: master-0 decided to set FNC (higher revision), master-1 decided JOIN_AS_LEARNER (lower revision)
- master-1's both-yield guard fired: "both-yield detected: no FNC holder and no active peer - retrying" (master-0's FNC write hadn't propagated yet due to hook delay)
- On retry, master-0 was active, master-1 joined as learner
- The dual-FNC variant of this race resolves via the identical verdict logic proven live in Test 1, with the post-set guard as code-trace-verified backstop
- **Exercises**: test hooks, both-yield guard (first live firing), race-window convergence
- **Result: PASSED** -- both-yield guard caught transient state, converged in one retry

### Test 7: Stale-claim deletion (banned loser)
- Set FNC on both nodes, master-0's CIB revision set artificially low (1000 -- banned node's RA never overwrites), master-0 banned
- master-1's early check: dual FNC -> verdict master-1 wins (55074 > 1000) -> detected master-0 not starting/active/pending -> deleted master-0's stale FNC attribute
- Logs: "FNC conflict inputs: master-1=55074 master-0=1000" -> "master-1 wins FNC conflict resolution" -> "peer master-0 not starting/active - clearing its stale force_new_cluster claim"
- After unban: master-0 rejoined as learner, promoted to voter
- **Exercises**: stale-claim deletion, peer_has_pending_start negative path, winner-only deletion
- **Result: PASSED**

### Verification: cibadmin exit code and peer_has_pending_start xpath
- No-match exit code: 105 (CRM_EX_NOSUCH) confirmed on Pacemaker 2.1.10
- **Positive match**: during a real pending start, the xpath `//node_state[@uname='master-1']//lrm_rsc_op[@operation='start'][@op-status='-1']` returned exit code 0 with the pending op element. The xpath is correct.
- **Negative match**: Test 7 (banned peer, no pending start) returned exit code 105. Deletion proceeded correctly.
- peer_has_pending_start verified in both directions: match -> deletion blocked; no-match -> deletion allowed.

### Final cluster state
- Both nodes: Ready (v1.35.3)
- All 35 Cluster Operators: Available=True, Progressing=False, Degraded=False
- PCS: Both etcd Started on both nodes, no FNC or standalone attributes set
- Zero manual interventions for cluster recovery across all tests

Related: OCPBUGS-86897 (fixed by #2181), OCPBUGS-100303

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1148,6 +1148,10 @@ set_force_new_cluster()
 {
 	local rc
 
+	if [ -n "$OCF_RESKEY_test_race_delay" ]; then
+		ocf_log notice "TEST: race delay ${OCF_RESKEY_test_race_delay}s before force_new_cluster write"
+		sleep "$OCF_RESKEY_test_race_delay"
+	fi
 
 	crm_attribute --lifetime reboot --node "$NODENAME" --name "force_new_cluster" --update "$NODENAME"
 	rc=$?
@@ -2262,6 +2266,10 @@ podman_start()
 	attribute_node_cluster_id update
 	attribute_node_revision update
 
+	if [ -n "$OCF_RESKEY_test_race_delay" ]; then
+		ocf_log notice "TEST: race delay ${OCF_RESKEY_test_race_delay}s after revision update in podman_start"
+		sleep "$OCF_RESKEY_test_race_delay"
+	fi
 
 	# ensure the etcd pod is not running before starting the container
 	ocf_log info "ensure etcd pod is not running (retries: $etcd_pod_poll_retries, interval: $etcd_pod_poll_interval_sec)"
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -907,6 +907,7 @@ wait_for_etcd_ports_release() {
 	return 0
 }
 
+# NOTE: 'get' reads the local disk JSON; use get_cib_attribute for CIB reads.
 attribute_node_cluster_id()
 {
 	local action="$1"
@@ -983,9 +984,7 @@ attribute_node_revision()
 
 attribute_node_revision_peer()
 {
-	local nodename
-	nodename=$(get_peer_node_name)
-	crm_attribute --query --type nodes --node "$nodename" --name "revision" | awk -F"value=" '{print $2}'
+	get_cib_attribute "$(get_peer_node_name)" "revision"
 }
 
 # Converts a decimal number to hexadecimal format with validation
@@ -1148,6 +1147,8 @@ add_member_as_learner()
 set_force_new_cluster()
 {
 	local rc
+
+
 	crm_attribute --lifetime reboot --node "$NODENAME" --name "force_new_cluster" --update "$NODENAME"
 	rc=$?
 	if [ $rc -ne 0 ]; then
@@ -1223,6 +1224,131 @@ is_force_new_cluster()
 
 	ocf_log debug "force_new_cluster attribute is not set on $NODENAME"
 	return 1
+}
+
+# Read a node attribute from CIB (not local disk) for symmetric evaluation.
+get_cib_attribute()
+{
+	crm_attribute --query --type nodes --node "$1" --name "$2" 2>/dev/null | awk -F"value=" '{print $2}'
+}
+
+# Verdict for dual force_new_cluster claims. Echoes one of:
+#   <node name> - that node force-new-clusters, the other yields
+#   EQUAL       - same revision and cluster_id: nobody needs FNC,
+#                 both clear their claim and start normally
+#   ADMIN       - no data to decide safely: never guess, soft-fail
+#                 with recovery guidance
+# All inputs read from CIB (symmetric); each node wrote revision and
+# cluster_id before setting FNC, so both are visible wherever both
+# FNC attributes are.
+fnc_conflict_verdict()
+{
+	local local_rev peer_rev peer_node_name
+
+	peer_node_name=$(get_peer_node_name)
+	local_rev=$(get_cib_attribute "$NODENAME" "revision")
+	peer_rev=$(get_cib_attribute "$peer_node_name" "revision")
+	ocf_log info "FNC conflict inputs: $NODENAME=$local_rev $peer_node_name=$peer_rev"
+
+	if ocf_is_decimal "$local_rev" && ocf_is_decimal "$peer_rev"; then
+		if [ "$local_rev" -gt "$peer_rev" ]; then
+			echo "$NODENAME"
+		elif [ "$local_rev" -lt "$peer_rev" ]; then
+			echo "$peer_node_name"
+		else
+			local local_cid peer_cid
+			local_cid=$(get_cib_attribute "$NODENAME" "cluster_id")
+			peer_cid=$(get_cib_attribute "$peer_node_name" "cluster_id")
+			if [ -n "$local_cid" ] && [ -n "$peer_cid" ] && [ "$local_cid" = "$peer_cid" ]; then
+				echo "EQUAL"
+			else
+				echo "ADMIN"
+			fi
+		fi
+	elif ocf_is_decimal "$local_rev"; then
+		echo "$NODENAME"
+	elif ocf_is_decimal "$peer_rev"; then
+		echo "$peer_node_name"
+	else
+		echo "ADMIN"
+	fi
+}
+
+# EQUAL: clear our claim, then confirm the peer concluded the same.
+# Divergent replicas can make one node see EQUAL while the other sees
+# a winner; the re-check closes that gap.
+fnc_equal_clear_and_confirm()
+{
+	local holders peer_node_name
+
+	peer_node_name=$(get_peer_node_name)
+	if ! clear_force_new_cluster; then
+		ocf_exit_reason "fnc_equal_clear_and_confirm: failed to clear own FNC"
+		return $OCF_ERR_GENERIC
+	fi
+	sleep 2
+	if ! holders=$(get_force_new_cluster); then
+		ocf_exit_reason "fnc_equal_clear_and_confirm: CIB query failed"
+		return $OCF_ERR_GENERIC
+	fi
+	if echo "$holders" | grep -q -w "$peer_node_name"; then
+		ocf_log notice "FNC race: cleared own claim but peer still holds - joining as learner"
+		JOIN_AS_LEARNER=true
+	else
+		ocf_log notice "FNC race: equal revision and cluster_id - joint normal start"
+		JOIN_AS_LEARNER=false
+	fi
+}
+
+fnc_admin_escalate()
+{
+	ocf_exit_reason "force_new_cluster held by multiple nodes ($1) with no revision data to resolve safely; inspect both nodes' etcd data, then clear the stale claim: crm_attribute --delete --lifetime reboot --node <loser> --name force_new_cluster"
+}
+
+# Loser-side cleanup: clear all leadership-asserting state in one place.
+fnc_yield()
+{
+	ocf_log notice "$NODENAME: yielding force_new_cluster, joining as learner"
+	clear_force_new_cluster
+	if is_standalone; then
+		clear_standalone_node
+	fi
+	JOIN_AS_LEARNER=true
+}
+
+# Is the peer active as a voter (not merely our learner)?
+# Yielding to a learner-only peer would leave no voter.
+is_peer_active_voter()
+{
+	local active_count
+	active_count=$(get_truly_active_resources_count)
+
+	if [ "$active_count" -ge 1 ] \
+		&& ! is_standalone \
+		&& [ "$(attribute_learner_node get)" != "$(get_peer_node_name)" ]; then
+		return 0
+	fi
+	return 1
+}
+
+# A start dispatched in an earlier transition is invisible to this
+# transition's notify vars; the CIB status section still records it
+# as a pending op (record-pending, default since Pacemaker 1.1.16).
+# Scoped to the etcd resource to avoid false positives from co-located
+# resources (VIP, STONITH) that may have pending starts during cold boot.
+# Returns 0 (pending or unknown -- don't delete) unless cibadmin
+# confirms no pending start (exit 6 = no xpath match).
+peer_has_pending_start()
+{
+	local rc rsc_id
+	rsc_id="${OCF_RESOURCE_INSTANCE%%:*}"
+	cibadmin -Q --xpath "//node_state[@uname='$1']//lrm_resource[@id='$rsc_id']/lrm_rsc_op[@operation='start'][@op-status='-1']" >/dev/null 2>&1
+	rc=$?
+	# 105 = CRM_EX_NOSUCH (Pacemaker 2.x), 6 = legacy errno ENXIO (1.x)
+	if [ "$rc" -eq 6 ] || [ "$rc" -eq 105 ]; then
+		return 1
+	fi
+	return 0
 }
 
 is_standalone()
@@ -2136,6 +2262,7 @@ podman_start()
 	attribute_node_cluster_id update
 	attribute_node_revision update
 
+
 	# ensure the etcd pod is not running before starting the container
 	ocf_log info "ensure etcd pod is not running (retries: $etcd_pod_poll_retries, interval: $etcd_pod_poll_interval_sec)"
 	for try in $(seq $etcd_pod_poll_retries); do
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -2300,6 +2485,41 @@ podman_start()
...
+		if ! fnc_guard_holders=$(get_force_new_cluster); then
+			ocf_exit_reason "Failed to verify force_new_cluster holders"
+			return "$OCF_ERR_GENERIC"
+		fi
+		if [ "$(echo "$fnc_guard_holders" | wc -w)" -ne 1 ] \
+			|| ! echo "$fnc_guard_holders" | grep -q -w "$NODENAME"; then
+			ocf_log err "force_new_cluster holders changed after decision ($fnc_guard_holders): retrying"
+			return "$OCF_ERR_GENERIC"
+		fi
+	fi
```
fonta-rh, line 2523: Non-blocking: two requests on this guard.

1. Add a comment documenting the theoretical failure mode: if CPG attribute propagation exceeds 3s (no hard SLA from attrd), both nodes pass this guard, both force-new-cluster, and the learner wipe at line 2550 destroys the loser's data. Probability is very low (CPG is typically sub-100ms, so this is a 30x+ margin), but documenting it helps future debugging.

2. The guard is silent on the success path. Add a log after the check passes:
```bash
ocf_log notice "post-set guard: $NODENAME is sole FNC holder ($fnc_guard_holders), proceeding with force-new-cluster"
```
If this message appears on both nodes within the same 3s window, that is conclusive evidence of a double-FNC event. The current err-level log at line 2504 only fires when the guard *catches* a conflict -- we need the symmetric case logged too.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1231,128 @@ is_force_new_cluster()
...
+# Divergent replicas can make one node see EQUAL while the other sees
+# a winner; the re-check closes that gap.
+fnc_equal_clear_and_confirm()
+{
+	local holders peer_node_name
+
+	peer_node_name=$(get_peer_node_name)
+	clear_force_new_cluster
+	sleep 2
+	holders=$(get_force_new_cluster 2>/dev/null)
```
fonta-rh: Blocking: both error paths in this function are swallowed.

Line 1296: `clear_force_new_cluster` return is unchecked. If the delete fails, FNC persists, the post-set guard at line 2495 fires and proceeds with FNC start -- but the EQUAL verdict intended a normal start. Same data on both nodes so no data loss, but unnecessary learner rebuild.

Line 1298: `get_force_new_cluster 2>/dev/null` discards the return code. On CIB failure, `holders` is empty, grep does not match the peer, function concludes "joint normal start." Since FNC was already cleared at line 1296, `is_force_new_cluster` at 2495 returns false and the post-set guard is entirely bypassed -- proceeding blind.

Fix: check both return codes, return `OCF_ERR_GENERIC` on failure.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1231,128 @@ is_force_new_cluster()
...
+
+	if is_numeric "$local_rev" && is_numeric "$peer_rev"; then
+		if [ "$local_rev" -gt "$peer_rev" ]; then
+			echo "$NODENAME"
+		elif [ "$local_rev" -lt "$peer_rev" ]; then
+			echo "$peer_node_name"
+		else
+			local local_cid peer_cid
+			local_cid=$(attribute_node_cluster_id get)
+			peer_cid=$(attribute_node_cluster_id_peer)
```
fonta-rh: Blocking: the comment at line 1252 states "All inputs read from CIB (symmetric)" but `attribute_node_cluster_id get` reads from disk (`jq` on `$ETCD_REVISION_JSON`, line 915), while `attribute_node_cluster_id_peer` reads from CIB (`crm_attribute --query`). Since `attribute_node_cluster_id update` swallows CIB write failures (lines 929-934, returns `OCF_SUCCESS` on error), a silent write failure produces asymmetric verdicts: one node sees EQUAL, the other sees ADMIN.

Self-heals on retry (ADMIN returns `OCF_ERR_GENERIC`), not a data-loss path, but the comment is wrong and the asymmetry is avoidable.

Fix: add a `get_cib_cluster_id` helper analogous to `get_cib_revision` and use it for both local and peer reads here. Correct the comment.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1231,128 @@ is_force_new_cluster()
...
+
+# A start dispatched in an earlier transition is invisible to this
+# transition's notify vars; the CIB status section still records it
+# as a pending op (record-pending, default since Pacemaker 1.1.16).
+# Returns 0 (pending or unknown -- don't delete) unless cibadmin
+# confirms no pending start (exit 6 = no xpath match).
+peer_has_pending_start()
+{
+	local rc
+	cibadmin -Q --xpath "//node_state[@uname='$1']//lrm_rsc_op[@operation='start'][@op-status='-1']" >/dev/null 2>&1
```
fonta-rh: Blocking: this xpath matches pending starts on *any* resource on the peer, not just the etcd resource. The `//` descendant-axis from `node_state` hits `lrm_rsc_op` across all `lrm_resource` elements.

During cold boot (the exact scenario this PR targets), co-located resources (VIP, STONITH) routinely have pending starts. A pending VIP start causes `peer_has_pending_start` to return 0, blocking stale-claim cleanup. The post-set guard then sees 2 holders and aborts -- delaying recovery by one or more retry cycles.

Fix: scope to the etcd resource:
```bash
cibadmin -Q --xpath "//node_state[@uname='$1']//lrm_resource[@id='$OCF_RESOURCE_INSTANCE']/lrm_rsc_op[@operation='start'][@op-status='-1']" >/dev/null 2>&1
```

fonta-rh (changes requested):
Solid design -- the conflict verdict + post-set guard + stale-claim cleanup is the right approach over the old OCF_ERR_CONFIGURED deadlock. Four items below, three blocking.

heartbeat/podman-etcd
Neilhamza: Addressed. Added the residual risk comment describing the attrd propagation window and the learner data wipe consequence. Also added the success path notice log so both nodes passing within 3s is visible evidence.

heartbeat/podman-etcd
Neilhamza: Addressed. Both clear_force_new_cluster and get_force_new_cluster return codes are now checked, returning OCF_ERR_GENERIC on failure.

heartbeat/podman-etcd
Neilhamza: Good catch, this mirrors the original bug's disk vs CIB asymmetry. Added get_cib_cluster_id helper (same pattern as get_cib_revision) and both sides now read from CIB. Comment corrected.

heartbeat/podman-etcd
Neilhamza: Addressed. Scoped to the etcd resource using lrm_resource[@id] with the instance suffix stripped (${OCF_RESOURCE_INSTANCE%%:*}) for clone safety. Re-verified the xpath on a live cluster in both directions: exit 0 during a real pending etcd start, exit 105 when healthy.

heartbeat/podman-etcd
Neilhamza: Updated: the caller in the EQUAL case branch now propagates the return code (if ! fnc_equal_clear_and_confirm; then return OCF_ERR_GENERIC), matching the ADMIN branch style. Both the in-function checks and the caller propagation are in place so the full error chain is covered.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1231,144 @@ is_force_new_cluster()
 	return 1
 }
 
+is_numeric()
+{
+	echo "$1" | grep -q '^[0-9][0-9]*$'
+}
+
+# Read a node's revision from CIB (not local disk) for symmetric evaluation.
+get_cib_revision()
```
clobrano: Nitpick (it would help maintenance)
I think this function could replace `attribute_node_revision_peer`, or be used inside it to avoid multiple calls of the same line `crm_attribute ... --name "revision"`.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1231,144 @@ is_force_new_cluster()
...
+{
+	local node="$1"
+	crm_attribute --query --type nodes --node "$node" --name "revision" 2>/dev/null | awk -F"value=" '{print $2}'
+}
+
+# Read a node's cluster_id from CIB (not local disk) for symmetric evaluation.
+get_cib_cluster_id()
```
clobrano: Isn't it the same as `attribute_node_cluster_id get`?

heartbeat/podman-etcd
```diff
@@ -2175,15 +2324,65 @@ podman_start()
...
+				EQUAL)
+					if ! fnc_equal_clear_and_confirm; then
+						return "$OCF_ERR_GENERIC"
+					fi
+					;;
+				ADMIN)
+					fnc_admin_escalate "$fnc_holders"
+					return "$OCF_ERR_GENERIC"
+					;;
+				*)
```
clobrano, line 2355: A comment here might help a newcomer to understand what's happening here without reading through the full logic :)

heartbeat/podman-etcd
```diff
@@ -2300,6 +2503,46 @@ podman_start()
...
+	# we are the sole visible holder. Same-invocation races (both nodes set
+	# FNC before either write propagated) are NOT resolved here - we
+	# soft-fail and let Pacemaker retry; the early multi-holder check
+	# resolves the conflict on the retry with fresh notify vars.
+	# The sleep is load-bearing: it gives a peer's in-flight write time to
+	# become visible so we fail here rather than double-FNC.
+	# Residual risk: if attrd propagation exceeds 3s (no hard SLA, but
+	# typically sub-100ms), both nodes pass this guard and both
+	# force-new-cluster; the loser's data directory is wiped during
+	# learner rejoin. Probability is very low given the 30x+ margin.
```
clobrano: If we make `revision` and `cluster-id` lifetime reboot, and let each node write them to CIB at startup, isn't reading the data correctly from CIB (i.e., matching what we have locally) a sign that attrd have been propagated?

clobrano: If we make `revision` and `cluster-id` lifetime reboot, and let each node write them to CIB at startup, isn't reading the data correctly from CIB (i.e., matching what we have locally) a sign that the attrd have been propagated?
This should let the resource agent pass this step quicker than 3s most of the times and also deal with longer network delays in heavy network load scenarios.

clobrano (commented):
The logic is sounds to me. It makes much more sense to let both nodes read from the same source.
I only left a couple of comments and an idea, but LGTM

heartbeat/podman-etcd
Neilhamza: Good call, delegated attribute_node_revision_peer to get_cib_revision so the query lives in one place.

heartbeat/podman-etcd
Neilhamza: They look similar but read from different sources: attribute_node_cluster_id get reads local disk (jq on the revision JSON file), get_cib_cluster_id reads from the CIB. That disk vs CIB asymmetry was the same class of bug that caused the original race, which is why the helper exists. Added a NOTE comment above attribute_node_cluster_id to make this clear for future readers.

heartbeat/podman-etcd
Neilhamza: Addressed, added a comment explaining the verdict named the peer and we yield.

heartbeat/podman-etcd
Neilhamza: Took your round-trip idea and applied it to the FNC write itself so it covers the actual race window. The revision/cluster_id round-trip would prove propagation up to a point before either node claims, but the FNC write happens later, so both could still pass unaware. With the FNC attribute as the barrier, CIB total ordering guarantees the later claimant always sees the conflict. Faster than the fixed 3s on healthy clusters and robust under load. Thanks for the insight.

fonta-rh (approved):

heartbeat/podman-etcd
```diff
@@ -1225,6 +1230,144 @@ is_force_new_cluster()
 	return 1
 }
 
+is_numeric()
```
oalbrigt: You should be able to use `ocf_is_decimal()` instead.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1230,144 @@ is_force_new_cluster()
...
+{
+	echo "$1" | grep -q '^[0-9][0-9]*$'
+}
+
+# Read a node's revision from CIB (not local disk) for symmetric evaluation.
+get_cib_revision()
```
oalbrigt: Can you merge these 2 functions into `get_attribute()` or a similarly named function? Seems like the only difference is the attribute name.

heartbeat/podman-etcd
```diff
@@ -1225,6 +1230,144 @@ is_force_new_cluster()
...
+# EQUAL: clear our claim, then confirm the peer concluded the same.
+# Divergent replicas can make one node see EQUAL while the other sees
+# a winner; the re-check closes that gap.
+fnc_equal_clear_and_confirm()
+{
+	local holders peer_node_name
+
+	peer_node_name=$(get_peer_node_name)
+	if ! clear_force_new_cluster; then
+		ocf_log err "fnc_equal_clear_and_confirm: failed to clear own FNC"
```
oalbrigt: You probably want to use `ocf_exit_reason()` for these errors.

heartbeat/podman-etcd
```diff
@@ -2136,6 +2279,11 @@ podman_start()
 	attribute_node_cluster_id update
 	attribute_node_revision update
 
+	if [ -n "${OCF_RESKEY_test_race_delay:-}" ]; then
```
oalbrigt: There shouldnt be any reason to do `:-` in a `if [ -n ... ]`

heartbeat/podman-etcd
```diff
@@ -2300,6 +2504,51 @@ podman_start()
...
+	# Post-set safety guard: wait until our own FNC write is visible in
+	# the local CIB replica (write round-trip through the DC). CIB
+	# updates are totally ordered: once our write is visible, any peer
+	# claim ordered before ours is visible too, so the later claimant
+	# always detects the conflict. Adaptive: passes as soon as
+	# propagation completes instead of a fixed sleep. Residual risk is
+	# bounded by CIB ordering correctness, not timing.
+	if is_force_new_cluster; then
+		local fnc_guard_holders fnc_guard_wait=0
+		while true; do
```
oalbrigt: `while :;` is probably more effective here, as it's an sh internal, and not a separate command.

oalbrigt (changes requested):
Just a few changes, and it should be good to go.

heartbeat/podman-etcd
fracappa: What @oalbrigt is suggesting would be useful also for another [PR](https://github.com/ClusterLabs/resource-agents/pull/2197) we have in queue

heartbeat/podman-etcd
Neilhamza: Done, switched to ocf_is_decimal and removed the custom is_numeric function.

heartbeat/podman-etcd
Neilhamza: Done, merged both into get_cib_attribute(node, attr_name). Also updated attribute_node_revision_peer to delegate to it.

heartbeat/podman-etcd
Neilhamza: Done, switched both to ocf_exit_reason.

heartbeat/podman-etcd
Neilhamza: Done, removed the :- at both hook sites.

heartbeat/podman-etcd
Neilhamza: Done, switched to while :; do.

oalbrigt (approved):

fonta-rh (approved):
lgtm!

oalbrigt:
Thanks.

PR #2195, opened 2026-08-03
spec: Remove EOL Fedora compatibility blocks from spec file

arslan9292:
Fedora 33 and earlier have been End of Life (EOL) for over
4 years, making backwards compatibility for these releases
unnecessary.

This commit removes/updates fedora checks and clean up the
spec file. As an added benefit, removing these blocks
resolves a silent logic bug where non-Fedora builds (like
CentOS or RHEL) evaluating `0 < 34` as true and incorrectly
executing these Fedora-specific instructions.


oalbrigt:
Thanks.

PR #2196, opened 2026-08-04
Route: restore unique="1" for destination and warn on "default"

arslan9292:
The PR #2193 change to `unique="0"` fixed dual-stack default routes but unintentionally removed the safety guardrail against accidentally duplicating normal subnet routes.

This commit addresses the issue safely by:
- Reverting the destination parameter back to `unique="1"` to ensure strict route validation.
- Updating the metadata description to advise using explicit CIDR notation (`0.0.0.0/0` and `::/0`) for dual-stack default routes.
- Adding a runtime warning when the string `default` is used.

Refs #2182

### DIFF
--- a/heartbeat/Route
+++ b/heartbeat/Route
@@ -88,13 +88,16 @@ of the Route resource agent:
 
 <parameters>
 
-<parameter name="destination" unique="0" required="1">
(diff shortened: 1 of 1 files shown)


oalbrigt:
Thanks.

PR #2201, opened 2026-08-19
OCPBUGS-112044: podman-etcd: treat crm_attribute 'not found' exits as empty, not error

fracappa:
crm_attribute --query returns exit code 105 (CRM_EX_NOSUCH) on Pacemaker 2.1.x+ and exit code 6 (ENODATA) on older versions whenan attribute does not exist.

The previous code treated any non-zero exit as a hard failure, causing false errors in get_cib_permanent_attribute
and get_cib_transient_attribute when querying attributes that have not been set yet (e.g. force_new_cluster, restart_no_leave, revision).

All callers already handle empty output gracefully, so let these two exit codes fall through to produce an empty string instead of aborting.


oalbrigt:
Thanks.

PR #2203, opened 2026-08-20
portblock: fix incorrect parameters for NftDelete() when using reset_local_on_unblock_stop

### LINKED ISSUE #2202: portblock: `NftDelete` call in `PortBLOCK` only passes 4 of 5 arguments, so the OUTPUT reject rule is never deleted
The `portblock` resource agent has a defect that prevents firewall rules from getting removed on `unblock` stop actions, on systems that use `nft`, rather than `iptables` to manage firewall rules.
Seen when testing on AlmaLinux 10. This can prevent access to managed services in some cases.

## Affected code

`heartbeat/portblock` at `main`.

`NftDelete` takes five parameters (line 557):

```sh
#NftDelete chain proto {d|s} ip ports
NftDelete()
{
        local chain=$1 proto=$2 ds=$3 ip=$(echo "$4" | sed "s#/#\\\/#") ports=$5
```

The call in `DoPort` passes five (line 589):

```sh
          NftDelete "$chain" "$proto" "$ds" "$ip" "$ports"
```

The call in `PortBLOCK`, on the `reset_local_on_unblock_stop` path, passes four (line 633):

```sh
        rc_in=$?
        if $try_reset ; then
          NftDelete "OUTPUT" "$1" "s" "$ports"
        fi
```

`$ports` is not set in `PortBLOCK`.
The function parameters there are `$1` protocol, `$2` ports, `$3` ip.
Inside `NftDelete` this leaves `ip` empty and `ports` unset, so the handle search matches nothing and no rule is deleted.

The `iptables` branch of the same `if $try_reset` block passes the values directly and deletes its rule:

```sh
          $IPTABLES $wait -D OUTPUT -p "$1" -s "$3" -m multiport --sports "$2" -j REJECT --reject-with tcp-reset
```

Introduced in #2073.

## Reproducer

No cluster or Pacemaker needed to reproduce this.
You can invoke the agent directly.

However, the same rule was left behind when using the resource agent with Pacemaker in a 3-node cluster running AlmaLinux 10.2, with `status_check=pseudo` set on both the `action=block` and the `action=unblock` primitive.

`repro.sh`:

```sh
#!/bin/sh
AGENT=${AGENT:-/usr/lib/ocf/resource.d/heartbeat/portblock}

export OCF_ROOT=/usr/lib/ocf
export OCF_RESOURCE_INSTANCE=repro_unblock
export OCF_RESKEY_action=unblock
export OCF_RESKEY_protocol=tcp
export OCF_RESKEY_portno=12345
export OCF_RESKEY_ip=192.0.2.1
export OCF_RESKEY_tickle_dir=/tmp/repro_tickle
export OCF_RESKEY_reset_local_on_unblock_stop=1
mkdir -p "$OCF_RESKEY_tickle_dir"

echo "== start =="
"$AGENT" start; echo "rc=$?"
nft list table inet portblock

echo "== stop =="
"$AGENT" stop; echo "rc=$?"
nft list table inet portblock
```

Run:

```sh
curl -sSLo /root/portblock.main \
  https://raw.githubusercontent.com/ClusterLabs/resource-agents/main/heartbeat/portblock
chmod +x /root/portblock.main
AGENT=/root/portblock.main sh repro.sh
```

## Actual result

`stop` returns 0 and leaves the `OUTPUT` reject rule in place:

```
== start ==
rc=0
table inet portblock {
	chain INPUT {
		type filter hook input priority filter; policy accept;
	}

	chain OUTPUT {
		type filter hook output priority filter; policy accept;
	}
}
== stop ==
rc=0
table inet portblock {
	chain INPUT {
		type filter hook input priority filter; policy accept;
		ip daddr 192.0.2.1 tcp dport 12345 ct state { established, related, new } drop
	}

	chain OUTPUT {
		type filter hook output priority filter; policy accept;
		ip saddr 192.0.2.1 tcp sport 12345 ct state { established, related, new } reject with tcp reset
	}
}
```

No later invocation of the agent removes the rule, and each subsequent `stop` inserts another copy of it.
Counting the rules in each chain after a `start stop start monitor stop monitor` sequence:

```
after start (rc=0)     OUTPUT=0 INPUT=0
after stop (rc=0)      OUTPUT=1 INPUT=1
after start (rc=0)     OUTPUT=1 INPUT=0
after monitor (rc=0)   OUTPUT=1 INPUT=0
after stop (rc=0)      OUTPUT=2 INPUT=1
after monitor (rc=7)   OUTPUT=2 INPUT=1
```

`start` removes the `INPUT` drop rule and leaves the `OUTPUT` reject rule in place.
`monitor` does not change either chain.
Deleting the accumulated rules requires an operator to run `nft delete rule inet portblock OUTPUT handle <handle>` once per rule, or to delete the table.

## Expected result

The `INPUT` drop rule is added by `unblock stop`, and the `OUTPUT` reject rule inserted earlier in the same action is removed before `stop` returns:

```
== stop ==
rc=0
table inet portblock {
	chain INPUT {
		type filter hook input priority filter; policy accept;
		ip daddr 192.0.2.1 tcp dport 12345 ct state { established, related, new } drop
	}

	chain OUTPUT {
		type filter hook output priority filter; policy accept;
	}
}
```

## Suggested change

Disclaimer: This is an LLM-written patch, however it was tested to produce the "expected result" by a human using the reproducer script shown earlier.

```diff
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -630,7 +630,7 @@ PortBLOCK()
         fi
         rc_in=$?
         if $try_reset ; then
-          NftDelete "OUTPUT" "$1" "s" "$ports"
+          NftDelete "OUTPUT" "$1" "s" "$3" "$2"
         fi
       else
         if $try_reset ; then
```

Apply it to the file the reproducer downloaded earlier:

```sh
sed -i 's/NftDelete "OUTPUT" "$1" "s" "$ports"/NftDelete "OUTPUT" "$1" "s" "$3" "$2"/' \
  /root/portblock.main
```

With this change applied to the same file, the reproducer produces the expected result shown earlier.

## Environment

| Item | Value |
| --- | --- |
| Agent tested | `heartbeat/portblock` from `main`, fetched 2026-08-18 |
| Also reproduced with | `resource-agents-4.16.0-53.el10_2.9.x86_64` |
| OS | AlmaLinux 10.2 |
| Kernel | 6.12.0-211.47.1.el10_2.x86_64 |
| `nftables` | 1.1.5-5.el10_2 |
| Firewall back end selected | `nft`, by `detect_firewall_tool` |

oalbrigt:
Fixes https://github.com/ClusterLabs/resource-agents/issues/2202.

### DIFF
--- a/heartbeat/portblock
+++ b/heartbeat/portblock
@@ -630,7 +630,7 @@ PortBLOCK()
         fi
         rc_in=$?
         if $try_reset ; then
-          NftDelete "OUTPUT" "$1" "s" "$ports"
+          NftDelete "OUTPUT" "$1" "s" "$3" "$2"
         fi
       else
         if $try_reset ; then


emteelb:
Thanks for the quick work on this.

Tested by directly invoking the agent (see reproducer script in #2202 ) and also when using the `portblock` RA in a Pacemaker cluster, testing failovers. The change to the RA worked in both cases. Direct invocation showed expected results (see #2202). In a Pacemaker cluster, failovers worked, client read and writes succeeded, no firewall rules accumulated as they did before the change in this commit.

Lgtm from the testing side. Cannot speak to the failing CI check though.

If you need logs or traces rather than take me at my word, lmk.

oalbrigt:
Great.

Seems like that's due to some curl timeouts, so I'll rerun it.

PR #2204, opened 2026-08-20
podman-etcd: record departed member to fix graceful shutdown restart selection

jaypoulz:
When both nodes stop gracefully, one leaves the etcd member list. Without a durable record of who departed, the startup tie-break may select the departed node — whose WAL contains a self-removal ConfChange — over the surviving member, causing a panic on force-new-cluster replay.

Add a permanent CIB attribute (departed_epoch) to record which node left. Wire it into fnc_conflict_verdict() so the non-departed node always wins. Move the revision snapshot to the end of stop so both nodes capture their final state.

### DIFF
--- a/heartbeat/podman-etcd
+++ b/heartbeat/podman-etcd
@@ -1246,6 +1246,60 @@ clear_restart_no_leave()
 	ocf_log info "$NODENAME: restart_no_leave attribute cleared"
 }
 
+# Set the member_removal_lock CIB permanent attribute on the current node.
+# Records which node successfully left the etcd member list during a
+# graceful shutdown, so the startup path can prefer the non-departed
+# node (which has fresher data).  The attribute is permanent
+# (--type nodes) so it survives reboots and is visible to the peer
+# even after this node's Pacemaker stops.
+set_member_removal_lock()
+{
+	local epoch="$1"
+
+	if ! crm_attribute --type nodes --node "$NODENAME" --name "member_removal_lock" --update "$epoch"; then
+		ocf_log err "could not set member_removal_lock attribute to $epoch"
+		return $OCF_ERR_GENERIC
+	fi
+	ocf_log info "$NODENAME: member_removal_lock set to $epoch"
+}
+
+# Clear the member_removal_lock CIB permanent attribute on the current node.
+# Called during podman_start() after a successful startup so that a
+# previous departure record does not influence future stop/start cycles.
+# Follows the same one-shot pattern as restart_no_leave.
+clear_member_removal_lock()
+{
+	local my_lock
+
+	my_lock=$(get_cib_permanent_attribute "$NODENAME" "member_removal_lock")
+	if [ -z "$my_lock" ]; then
+		return
+	fi
+	if ! crm_attribute --type nodes --node "$NODENAME" --name "member_removal_lock" --delete; then
+		ocf_log warn "could not clear member_removal_lock attribute"
+		return
+	fi
+	ocf_log info "$NODENAME: member_removal_lock attribute cleared"
+}
+
+# Clear the fnc_panic_rescue CIB permanent attribute on the current node.
+# Called during podman_start() after a successful startup so that the
+# panic rescue safety net can trigger again in a future cycle.
+clear_fnc_panic_rescue()
+{
+	local my_marker
+
+	my_marker=$(get_cib_permanent_attribute "$NODENAME" "fnc_panic_rescue")
+	if [ -z "$my_marker" ]; then
+		return
+	fi
+	if ! crm_attribute --type nodes --node "$NODENAME" --name "fnc_panic_rescue" --delete; then
+		ocf_log warn "could not clear fnc_panic_rescue attribute"
+		return
+	fi
+	ocf_log info "$NODENAME: fnc_panic_rescue attribute cleared"
+}
+
 # Read a permanent node attribute from CIB (not local disk) for symmetric evaluation.
 get_cib_permanent_attribute()
 {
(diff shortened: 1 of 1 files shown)


heartbeat/podman-etcd
```diff
@@ -2800,13 +3049,106 @@ leave_etcd_member_list()
...
+	#
+	# The concurrent-stop delay above is kept as a first-pass guard that
+	# keeps the two claims/removes from racing at the same instant.
+	local peer_node_name
+	peer_node_name="$(get_peer_node_name)"
+
+	# (1) If the peer already holds the lock, it is leaving/left first, so we
+	#     are the surviving member — skip.
+	local peer_lock
+	peer_lock=$(get_cib_permanent_attribute "$peer_node_name" "member_removal_lock")
```
vimauro: Is there a possibility that `get_cib_permanent_attribute` can fail ? In that case, `peer_lock` may not be reliable (can be empty or not-empty with an error code)

jaypoulz: I will add return code checking so that this handling is more explicit.

heartbeat/podman-etcd
```diff
@@ -2823,8 +3206,6 @@ podman_stop()
 	fi
 
 	attribute_learner_node clear
```
fonta-rh, line 3288: Blocking: the `container_running != "true"` early return (line 3215) skips both attribute snapshots below it — worth flagging since the two need different treatment.

`cluster_id` can move back to right after `attribute_learner_node clear`, unconditional, before this check — it reads from disk rather than the container, so the later position isn't required; it looks like it moved only because it rode along with revision.

`revision` is trickier: the relocation to the end reads `revision.json` after the graceful `podman stop` flush, which only happens on the running-container path. The crashed-container branch has no flush to wait for, so it's currently getting no snapshot at all. Two options that would close it: a call inside this `if`, or an unconditional read near the top that the end-of-function read overwrites on the graceful path.

heartbeat/podman-etcd
```diff
@@ -1280,6 +1354,26 @@ fnc_conflict_verdict()
 	local local_rev peer_rev peer_node_name
 
 	peer_node_name=$(get_peer_node_name)
+
+	# Member-removal-lock verdict: if one node left the etcd member list
+	# during graceful shutdown, prefer the non-departed node — its WAL
+	# does not contain a self-removal ConfChange that would panic on
+	# FNC replay.
+	local local_lock peer_lock
+	local_lock=$(get_cib_permanent_attribute "$NODENAME" "member_removal_lock")
```
fonta-rh: Blocking: the same-day fix addressed the flagged line (`leave_etcd_member_list`'s peer-lock check, now at 3112 and 3139) and `clear_member_removal_lock` (1309) — all three now check `get_cib_permanent_attribute`'s return code and fail safe on a genuine CIB error instead of reading it as "lock not held."

Six other reads of `member_removal_lock` still don't check the return code, so the same ambiguity persists in over half the call sites: `fnc_conflict_verdict`'s `local_lock`/`peer_lock` (1363, 1364), `fnc_panic_rescue`'s `rescue_peer_lock` (2499) — the "both nodes voterless → fail loud" gate, arguably the highest-stakes of the six — `podman_start`'s departure-aware routing `my_removal_lock`/`peer_removal_lock` (2536, 2537), and the learner-wait-loop's `early_peer_lock` (2833).

`get_cib_permanent_attribute` already distinguishes the two cases — it returns non-zero only on a genuine query error, not on "attribute unset" — so closing the gap at the remaining six sites is the same pattern already applied at the first three.

jaypoulz: All 9 are consistent now :)

heartbeat/podman-etcd
jaypoulz: chose the second path for revision - check up top alongside cluster-id, then refresh right before exit

heartbeat/podman-etcd
```diff
@@ -1131,6 +1147,18 @@ add_member_as_learner()
...
 			return $OCF_SUCCESS
 		fi
+		# podman exec exits 125 ("no such container"), 126 ("container
+		# state improper") or 255 ("container not running") when the etcd
+		# container is not running. Retrying etcdctl in that state cannot
+		# succeed and only masks the etcd crash behind it, so capture the
+		# container diagnostics and fail fast to let the caller restart it.
+		case "$rc" in
+			125|126|255)
+				ocf_log err "could not add $member_name as learner: etcd container not running (podman exec rc=$rc): $out"
```
oalbrigt, line 1157: Let's use `ocf_exit_reason()` when we return errors.

heartbeat/podman-etcd
```diff
@@ -1246,6 +1275,51 @@ clear_restart_no_leave()
...
+# graceful shutdown, so the startup path can prefer the non-departed
+# node (which has fresher data).  The attribute is permanent
+# (--type nodes) so it survives reboots and is visible to the peer
+# even after this node's Pacemaker stops.
+set_member_removal_lock()
+{
+	local epoch="$1"
+
+	if ! crm_attribute --type nodes --node "$NODENAME" --name "member_removal_lock" --update "$epoch"; then
+		ocf_log err "could not set member_removal_lock attribute to $epoch"
```
oalbrigt, line 1289: Let's use `ocf_exit_reason()` here as well

oalbrigt (dismissed):
Let's use `ocf_exit_reason()` where we return ERR codes.

heartbeat/podman-etcd
```diff
@@ -2554,6 +2804,16 @@ podman_start()
 			return "$OCF_ERR_GENERIC"
 		fi
 		ocf_log notice "post-set guard: $NODENAME is sole FNC holder, proceeding with force-new-cluster"
+
+		# FNC precondition: refuse force-new-cluster when this node is a
+		# learner — its data has no voter entry for itself, so FNC replay
+		# would hit "removed all voters" and panic.  Route to learner-rejoin
+		# or fail-safe instead.
+		if is_learner; then
```
fonta-rh, line 2844: Blocking: this precondition can't fire on the `active_resources_count=0` → `start_resources_count=1` cold-start path — one of the more common routes into force-new-cluster. `attribute_learner_node clear` runs unconditionally at the top of the `0` branch (line 2692) before `set_force_new_cluster` gets called further down, so by the time this `is_learner` check runs, `learner_node` has already been wiped in the same invocation — `is_learner` reads that same attribute and can only return false here. Two options that would close it: read `is_learner` once before the `0` branch clears the attribute and carry the result forward, or move the clear to after this FNC precondition check runs.

heartbeat/podman-etcd
```diff
@@ -2731,16 +3005,30 @@ podman_start()
 						ocf_log info "peer Machine is being deleted, skipping learner addition for $peer_node_name"
 					else
 						add_member_as_learner "$peer_node_name" "$peer_node_ip"
```
fonta-rh, line 3039: Non-blocking: by this point `monitor_cmd_exec` already confirmed etcd is up as the seeded voter, and `add_member_as_learner` has already retried 3 times internally (and fails fast rather than retrying when the container isn't running). Failing the whole `start` here restarts an already-healthy container instead of just retrying the learner-add. `check_peer()` → `manage_peer_membership()` already retries an absent peer as learner on every monitor cycle without a restart — the same self-healing mechanism covers the peer-offline case elsewhere in this function. Returning success here and letting that path pick up the learner-add on the next cycle would avoid the extra restart.

heartbeat/podman-etcd
```diff
@@ -2329,7 +2465,121 @@ podman_start()
...
+		if podman inspect "${CONTAINER}" >/dev/null 2>&1; then
+			prev_container="${CONTAINER}"
+		elif podman inspect "${CONTAINER}-previous" >/dev/null 2>&1; then
+			prev_container="${CONTAINER}-previous"
+		fi
+		if [ -n "$prev_container" ]; then
+			prev_logs=$(podman logs --tail 200 "$prev_container" 2>&1 || true)
+			# Require BOTH the raw panic line AND the precursor
+			# "membership: ID removed" error to avoid false triggers.
+			if echo "$prev_logs" | grep -qF 'panic: removed all voters' \
```
fonta-rh, line 2511: Non-blocking: two related risks in this match. First, the window — this greps the last 200 lines of a non-archived container's cumulative logs, not lines since the current stop. A node that panics once, gets rescued via this path, and then runs healthy long enough to log fewer than 200 lines could have that old panic still in the window — a later, unrelated force-new-cluster attempt on this node would read as "panicked recently" and get incorrectly blocked. A timestamp-bounded check (only lines since the current stop) would close that. Second, this couples to etcd's literal, unversioned panic text (`panic: removed all voters`, `membership: ID removed`) — if a future etcd release rewords it, this stops firing with no operator-visible signal. Worth a comment noting the coupling, or a test pinned against the actual etcd binary.

heartbeat/podman-etcd
jaypoulz: This comment misses that this is intentional:
  - On live paths active_resources_count >=1, the guard works with a fresh value being supplied by monitor.
  - On cold boot (active_resources_count = 0), there's no trustworthy learner signal to have (revision.json lacks it, member list needs live etcd), so the guard shouldn't fire — and cold-start voterless-FNC is owned by the leaver-lock routing (runs first) + the panic rescue (reads ground truth).
  
I will add a comment in the code to make this clearer, but I don't think this should be changed.

heartbeat/podman-etcd
jaypoulz: I've done a little discriminating here. I agree that restarting a container for no reason is not idea, but in my experience, when learner-add fails here, it's because the etcd container crashed! This error return ensures we actually have a rescue path from that option, otherwise we're in a live-lock where the peer is waiting forever to be added by a container that "successfully started" but crashed on startup.

I've added specific checking for *why* the learner add failed.

heartbeat/podman-etcd
jaypoulz: Let me add a note to add a recovery test that verifies the rescue path. I have it written up for QE, and that would close the loop on part 2.

I might be able to get chai bot to write it and we feed two birds with one scone. I think I have a fix for the time boundary issue you flagged.

heartbeat/podman-etcd
fonta-rh: Ok, yeah, I stand corrected. The lock runs before the self-removal, we're good!!

fabbione (approved):

PR #2206, opened 2026-08-21
ZFS: wait for the pool state to settle before start returns

lhjnano:
### Problem

A healthy pool can fail its first recurring monitor right after being started,
and pacemaker then recovers a resource that is fine:

```
17:23:47  Initiating start operation ZPOOL_ostPoolA_start_0 locally on ctrla
17:23:47  ZFS(ZPOOL_ostPoolA): INFO: ostPoolA:starting import
17:23:47  ZPOOL_ostPoolA start (call 1089) exited with status 0 (execution time 692ms)
17:23:47  Result of start operation for ZPOOL_ostPoolA on ctrla: ok | rc=0
17:23:47  Initiating monitor operation ZPOOL_ostPoolA_monitor_5000 locally on ctrla
17:23:47  Result of monitor operation ... : error | call=1090 rc=1
17:23:47  fail-count-ZPOOL_ostPoolA#monitor_5000 = 1
17:23:47  Actions: Recover ZPOOL_ostPoolA
```

`/proc/spl/kstat/zfs/<pool>/state` is a lock-free snapshot of the root vdev
state, which is what makes it cheap to poll. It also exposes intermediate
states: just after an import the root vdev is still being reopened, so the file
reads `OFFLINE` (`VDEV_STATE_CLOSED`) or `TRANSITIONING` for a short while.
`zpool_monitor` maps anything that is not `ONLINE`, `DEGRADED` or `FAULTED` to
`OCF_ERR_GENERIC`, and pacemaker starts the recurring monitor the moment start
returns — exactly when that window is open.

`rc=1` can only come from the catch-all branch, so the value read was none of
the three handled states.

### Measurements

Sampling the kstat on every change while repeatedly creating and destroying
pools, over 65 minutes:

| | value |
|---|---|
| non-`ONLINE` exposure, overall | **0.12 %** |
| non-`ONLINE` exposure, first 0.5 s after import | **7.06 %** (peak 31 %) |
| contiguous non-`ONLINE`, median / max | 4 ms / **78 ms** |
| `OFFLINE` intervals falling within 5 s of an import | 52 % |

So the reading is almost always fine, but it is ~59x more likely to be wrong in
the exact window where pacemaker schedules the first recurring monitor.

At each non-`ONLINE` kstat reading I also ran `zpool list -H -o health` on the
same pool:

| kstat | `zpool list -o health` | count |
|---|---|---|
| `OFFLINE` | `ONLINE` | 33 |
| `TRANSITIONING` | `ONLINE` | 11 |
| both non-`ONLINE` | — | **0** |

The pool was healthy every time; the kstat was mid-transition.

### Reproducing

No special hardware. Create and destroy a pool in a loop while polling
`/proc/spl/kstat/zfs/<pool>/state`, logging on change. Under a pacemaker
resource with a 5s monitor, the monitor eventually lands inside the window and
records a failure — it took 6 iterations here.

### The change

Wait for the state to settle at the end of a successful import rather than
softening the monitor:

* the monitor still judges exactly as before, so a pool that really is offline
  still fails, and no `zpool` command is added to the monitor path;
* the state is read once before any sleeping, so an already-settled pool — the
  common case — costs nothing;
* the wait is bounded by `settle_timeout` (default 10s) and additionally
  clamped to `OCF_RESKEY_CRM_meta_timeout`, the same way `asterisk`, `lxc` and
  `mysql-common` derive their waits, so it cannot consume the start timeout;
* `settle_timeout=0` disables it;
* it never fails the start — on timeout it logs a warning and returns success,
  since judging the pool is the monitor's job.

Happy to drop the parameter and hardcode the bound if you would rather not grow
the agent's interface.

### Checks

* `sh -n` clean; `ci/build.sh`'s shellcheck rule passes (no `error`-level
  findings). The one new note is `SC2223` on the added `: ${OCF_RESKEY_...=}`
  line, identical to the four already there.
* `meta-data` output parses, four parameters.
* Function tested against a stubbed kstat: already settled (returns at once),
  settles after 2s, never settles (warns, returns success), `settle_timeout=0`,
  no state file (older zfs), non-numeric `settle_timeout`, and the clamp with
  both a large `settle_timeout` and a very short operation timeout.

### DIFF
--- a/heartbeat/ZFS
+++ b/heartbeat/ZFS
@@ -25,10 +25,12 @@
 OCF_RESKEY_pool_default=""
 OCF_RESKEY_importargs_default=""
 OCF_RESKEY_importforce_default=true
+OCF_RESKEY_settle_timeout_default="10"
 
 : ${OCF_RESKEY_pool=${OCF_RESKEY_pool_default}}
 : ${OCF_RESKEY_importargs=${OCF_RESKEY_importargs_default}}
 : ${OCF_RESKEY_importforce=${OCF_RESKEY_importforce_default}}
+: ${OCF_RESKEY_settle_timeout=${OCF_RESKEY_settle_timeout_default}}
 
 USAGE="usage: $0 {start|stop|status|monitor|validate-all|meta-data}";
 
@@ -68,6 +70,24 @@ zpool import is given the -f option.
 <shortdesc lang="en">Import is forced</shortdesc>
 <content type="boolean" default="${OCF_RESKEY_importforce_default}" />
 </parameter>
+<parameter name="settle_timeout" unique="0" required="0">
+<longdesc lang="en">
+Seconds to wait after a successful import for the pool state to settle before
+reporting the pool as started.
+
+/proc/spl/kstat/zfs/&lt;pool&gt;/state is a lock-free snapshot of the root vdev
+state, so shortly after an import it can still read OFFLINE or TRANSITIONING
+while vdevs are being reopened. Pacemaker starts the recurring monitor as soon
+as start returns, and the monitor treats anything other than ONLINE, DEGRADED
+or FAULTED as a hard error, so a reading taken inside that window fails a
+healthy pool and triggers a needless recovery.
+
+The pool state is read once before any waiting, so a pool that is already
+settled - the common case - costs nothing. Set to 0 to disable the wait.
+</longdesc>
+<shortdesc lang="en">Seconds to wait for the pool state to settle after import</shortdesc>
+<content type="integer" default="${OCF_RESKEY_settle_timeout_default}" />
+</parameter>
 </parameters>
 
 <actions>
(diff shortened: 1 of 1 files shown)


heartbeat/ZFS
```diff
@@ -96,6 +116,57 @@ zpool_is_imported () {
...
+# state, so right after an import it can still read OFFLINE (the root vdev is
+# briefly VDEV_STATE_CLOSED while vdevs are reopened) or TRANSITIONING.
+# Pacemaker starts the recurring monitor as soon as start returns, which is
+# precisely when that window is open, and zpool_monitor maps anything that is
+# not ONLINE, DEGRADED or FAULTED to OCF_ERR_GENERIC.
+#
+# This never fails the start. Judging the pool is the monitor's job, and the
+# wait must not become a new way for start to fail.
+zpool_settle () {
+	[ -f "/proc/spl/kstat/zfs/$OCF_RESKEY_pool/state" ] || return 0
```
oalbrigt, line 143: You should move these checks to the validate_all function.

lhjnano: Moved it. `zpool_validate` now rejects a non-numeric `settle_timeout` with `ocf_exit_reason` and `OCF_ERR_CONFIGURED`:

```sh
if ! ocf_is_decimal "$OCF_RESKEY_settle_timeout"; then
	ocf_exit_reason "Invalid settle_timeout='${OCF_RESKEY_settle_timeout}': expected a non-negative integer"
	return $OCF_ERR_CONFIGURED
fi
```

`zpool_settle` no longer validates. It keeps one defensive line that falls back to the default rather than checking, so a bad value cannot break start on a system where `validate-all` was never run:

```sh
ocf_is_decimal "$OCF_RESKEY_settle_timeout" \
	|| OCF_RESKEY_settle_timeout="$OCF_RESKEY_settle_timeout_default"
[ "$OCF_RESKEY_settle_timeout" -gt 0 ] || return 0
```

Dropping that line entirely would leave `[ "$x" -gt 0 ]` to fail on a non-numeric value, which is noisier than falling back. Happy to remove it if you would rather the check live in exactly one place.

The `-f` test on the kstat path stays in `zpool_settle` - it is a runtime condition, not configuration: older ZFS has no such file, and the same pool has one only once imported.

oalbrigt: You should also move the other parameter checks to validate_all().

And you can add `zpool_validate || exit $?` for `start)` in the action case at the end of the agent.

lhjnano: Moved the `settle_timeout` check to `zpool_validate`, which also rejects an empty `pool` now. `start` calls `zpool_validate || exit $?` before importing.

oalbrigt:
LGTM.

lhjnano:
Thanks for the review and the pointers — the parameter checks now live in `zpool_validate()` (rejects a non-numeric `settle_timeout` and an empty `pool` with `OCF_ERR_CONFIGURED`), and `start)` runs `zpool_validate || exit $?` before importing.

If there is nothing else to address, could you take another look and approve when you get a chance? Happy to make further changes if anything else comes up. (The CI run may also need an admin authorization.)

oalbrigt:
Thanks.

lhjnano:
**Follow-up: the settle wait can still lose the race — reproduced on a live cluster with this patch applied**

We backported this patch into our gluesys fork of the ZFS agent (same settle
logic, `settle_timeout` default 10) and ran it through our HA regression:
3 identical dual-controller clusters, preset teardown/rebuild cycles that
repeatedly stop and re-import ZFS pools, recurring monitor interval 5s.

One of the three clusters hit exactly the false positive this PR targets,
with the patch active:

```
Aug 27 18:26:46  Initiating start operation ZPOOL_mngP_start_0 locally on ctrla
Aug 27 18:26:46  ZFS(ZPOOL_mngP)[313368]: INFO: mngP:starting import
Aug 27 18:26:47  Result of start operation for ZPOOL_mngP on ctrla: ok
Aug 27 18:26:47  Result of monitor operation for ZPOOL_mngP on ctrla: error (exec=36ms)
Aug 27 18:26:47  fail-count-ZPOOL_mngP#monitor_5000 = 1
Aug 27 18:26:47  Actions: Recover ZPOOL_mngP
```

`pcs status` afterwards:

```
* ZPOOL_mngP_monitor_5000 on ctrla 'error' (1): call=361, status='complete', exec=36ms
```

## Why the wait doesn't cover this case

`zpool_settle()` polls the state kstat at 1-second granularity and returns on
the **first** ONLINE read. But the post-import window is not a monotonic
"not-yet-ONLINE → ONLINE" transition — the kstat oscillates between ONLINE
and OFFLINE while the root vdev is reopened, and the OFFLINE bursts are short:

- 65-minute sampling on this hardware (1,210 transitions): OFFLINE bursts
  median 4ms / max 78ms, concentrated in the first ~0.5s after import
- OFFLINE exposure during that 0.5s window: ~7%

So the first settle sample lands on an ONLINE instant in ~93% of imports,
`zpool_settle()` returns immediately without ever sleeping, and the recurring
monitor — which pacemaker starts the moment start returns — can still hit an
OFFLINE burst tens of milliseconds later (here: monitor exec 36ms).

The wait only protects the ~7% of imports where the first sample happens to
catch OFFLINE.

## What would close it

Require stability instead of a single sample, e.g.:

- poll every ~100ms and require 2–3 consecutive ONLINE reads (covers the
  observed 78ms max burst), or
- enforce a minimum post-import dwell (300–500ms) before start returns, or
- both

Happy to share the sampler data, and to test a revised patch on the same
setup — our regression churns enough pool imports to verify a fix quickly.

PR #2207, opened 2026-08-23
OCPBUGS-100304: podman-etcd: pre-check force_new_cluster holders before writing

Neilhamza:
### Problem and fix

`set_force_new_cluster()` wrote the `force_new_cluster` (FNC) CIB attribute without first checking whether the peer already held it. Since #2194 the post-set guard and the multi-holder tiebreaker catch a lost race, but only after a full failed-start retry cycle, and the monitor path (`detect_cluster_leadership_loss`) had no holder awareness at all — in a symmetric "no leader" partition both nodes wrote FNC and asserted standalone. This renames the setter to `try_set_force_new_cluster()` with a holder pre-check (rc `0`=set, `1`=CIB error, `2`=another node already holds FNC) and handles rc=2 at every call site, so contention is resolved before the write instead of after a retry.

### Call sites (4)

- Start, `start_resources_count=1` and `=2` — on rc=2 set `JOIN_AS_LEARNER=true` and continue; the FNC holder is already driving recovery.
- Monitor (`detect_cluster_leadership_loss`) — on rc=2 skip `set_standalone_node`, log the deferral, still return `OCF_ERR_GENERIC`; the restart lands in `podman_start` and joins as learner.
- Seed site (from afcec12f, "peer departed via `member_removal_lock`, we are the surviving voter") — rc=2 fails closed with an `ocf_exit_reason` naming the contradiction and the exact `crm_attribute --delete` remedy: a peer that departed gracefully must not also hold FNC, so its claim is stale and would fail every retry identically until cleared.

### Rebase

Rebased onto the departure-handshake work (afcec12f, ec3d5a83), which added a deterministic `member_removal_lock` routing block at the top of `podman_start`. The two revision-comparison start arms now run only when the lock recorded no decision (the ungraceful / symmetric-no-leader case). The seed call site it added is the fourth above.

### Validation

Validated on a 4.22 TNF cluster (RA md5 caf66f10, both nodes):
- Monitor-path defer: master-1 saw master-0's claim, deferred, never set standalone, rejoined as learner.
- Stock RA control: same setup, stock writes a second FNC claim and hits the tiebreaker.
- Ungraceful node loss: unchanged — survivor forces new cluster, peer rejoins as learner.
- Organic restarts x3: unchanged, no contention.
- Graceful departure seed (afcec12f path): survivor seeds normally, no exit reason.
- Seed with stale peer claim: start fails with the exit reason and remedy command; the command clears it and the survivor seeds.

### Known limitation

If the peer's FNC claim is stale because it never starts, the learner wait burns its timeout and retries through the existing single-peer-holder path — a pre-existing exposure this change converges to rather than introduces.

Fixes OCPBUGS-100304. Builds on #2194.


Neilhamza:
@jaypoulz @clobrano — ready for review. This builds on #2194 (the FNC race resolution you reviewed): adds a holder pre-check to `set_force_new_cluster()` and handles the new rc=2 at all three call sites. The PR body has the full validation report with before/after journal excerpts (patched vs stock RA on the same monitor-path scenario). CI is waiting for admin authorization on the knet-jenkins runs.

fracappa (approved):
This is a robust pre-check for force_new_cluster scenario.

LGTM, thanks @Neilhamza !

heartbeat/podman-etcd
```diff
@@ -1144,21 +1144,42 @@ add_member_as_learner()
...
 
+# Best-effort pre-check: reduces churn but cannot close the race.
+# The check-then-set is inherently TOCTOU; the post-set guard in
+# podman_start and the multi-holder tiebreaker (fnc_conflict_verdict)
+# are the authoritative resolution for any write that slips through.
+# Return codes:
+#   0 - attribute set (holders were empty or self-only)
+#   1 - crm/CIB error (caller: soft-fail)
+#   2 - another node already holds force_new_cluster (caller: defer/yield)
 set_force_new_cluster()
```
clobrano: Given the changes, this function might be renamed including `try` in the new name (or following `git` example, `set_force_new_cluster_with_lease` :) ).

heartbeat/podman-etcd
```diff
@@ -2494,9 +2544,26 @@ podman_start()
...
+						2)
+							# First-claimant-wins in this residual window:
+							# for the peer to claim first despite our
+							# "newer" verdict, its own comparison must have
+							# used stale/asymmetric saved attributes —
+							# the same data fnc_conflict_verdict would
+							# read, so the tiebreaker is no more
+							# authoritative here. With accurate attributes
+							# the older node computes "older" and goes
+							# learner without ever claiming.
```
clobrano: This comment is a bit too complex to understand

heartbeat/podman-etcd
```diff
@@ -1144,21 +1144,42 @@ add_member_as_learner()
 	return $OCF_ERR_GENERIC
 }
 
+# Best-effort pre-check: reduces churn but cannot close the race.
+# The check-then-set is inherently TOCTOU; the post-set guard in
+# podman_start and the multi-holder tiebreaker (fnc_conflict_verdict)
+# are the authoritative resolution for any write that slips through.
```
clobrano: This docstring seems more an explanation of the current fix, than the description of what the function does

clobrano (changes requested):
I left a couple of comments

clobrano (approved):

heartbeat/podman-etcd
```diff
@@ -2481,9 +2527,17 @@ podman_start()
...
-						if ! set_force_new_cluster; then
+						try_set_force_new_cluster
+						case $? in
+						0) ;;
+						2)
+							ocf_log notice "peer claimed force_new_cluster first: joining as learner"
+							JOIN_AS_LEARNER=true
+							;;
+						*)
 							return "$OCF_ERR_GENERIC"
```
oalbrigt, line 2538: Let's use `ocf_exit_reason()` here, so users can see why it failed.

Neilhamza: Done in 76f3e37 — both arms now set an exit reason, same as the monitor path. Also rebased onto afcec12f and gave its new seed call site the same treatment.

Neilhamza:
Live validation on a 5.1 two-node-fencing cluster (5.1.0-0.nightly-2026-09-13-222843, RHEL CoreOS 10.2).

Installed this PR as an RPM (resource-agents-4.18.0-1.99.76f3e.el10, built from 76f3e379) via rpm-ostree override on both control-plane nodes, so the running podman-etcd is the patched one (md5 caf66f10, `try_set_force_new_cluster` present). The override was re-checked on both nodes every 2 minutes for the whole run and never reverted across any fence, reboot or cold boot.

Results
- openshift/etcd/certrotation: 4/4 pass.
- openshift/two-node, node-replacement test excluded (needs the RPM delivered via MachineConfig to survive reprovisioning): 30 pass, 1 fail, 11 skip in 4h08m. Skips are topology-gated (6 arbiter-only tests, degraded-mode sig-apps). All force-new-cluster, cold-boot and member re-add recovery tests passed.
- The one failure (PacemakerHealthCheck maintenance-mode test) missed a PacemakerHealthy *event* inside its 2-minute window while the cluster itself had fully recovered (both nodes Ready, 2 voting members, pacemaker Healthy=True in the test's own diagnostics). Passed on a single isolated re-run (2m48s). Both runs are in the attached junit.

i can attach: junit for both suites, per-node override proof (rpm-ostree status, rpm -q, md5), the md5 watch log, and the run logs. Cluster-specific identifiers are redacted.

cc @jaypoulz

jaypoulz (commented):
Before we merge this - I have a different idea I'd like us to discuss/rule out.
Preface: cluster-etcd-operator **sets force_new_cluster during the update-setup job**, so whatever we decide needs a companion PR in CEO.

The whole point of making this atomic was to prevent a situation where both nodes claim force_new_cluster. The issue with this strategy is - even if we check first, we haven't actually made it atomic,  and then we need to have this tiebreak code that relies on the CIB reflection of the last know CEO-written revision number - which is know to be unreliable.

Would it be better - and potentially simpler (i.e. can we possible delete tie-break code) if we make the force_new_cluster flag a timestamp like how the member_removal_lock works? Then you have an easy tiebreak - if they're both set, the older timestamp wins. If only one is set, you know you have the force_new_cluster lock?

I feel like this is a better design than this approach and would likely result in us deleting a lot of the tie-break specific code.

PR #2211, opened 2026-08-26
SAPInstance - fix-issue-2209 - list-unit-files creates high load on systemd

fmherschel:
* list-unit-files (without limiting parameter) creates high load on systemd
* list-unit-files SAP<SID>_<INO>.service is 50 to 100 times faster
* RA could directly use systemctl return code and does not search for SAP<SID>_<INO>.service in the list

### DIFF
--- a/heartbeat/SAPInstance
+++ b/heartbeat/SAPInstance
@@ -457,8 +457,8 @@ check_systemd_integration() {
     local rc=1
 
     if which "$SYSTEMCTL" 1>/dev/null 2>/dev/null; then
-        if $SYSTEMCTL list-unit-files | \
-            awk '$1 == service { found=1 } END { if (! found) {exit 1}}' service="${systemd_unit_name}.service"; then
+        # use list-unit-files with expected unit-file to limit systemd load; this also allows to use the return-code of systemctl
+        if $SYSTEMCTL list-unit-files "${systemd_unit_name}.service" >/dev/null ; then
             rc=0
         else
             rc=1


oalbrigt:
LGTM.

fmherschel:
@oalbrigt - I do not know how the project is processing the PRs. Do I need to do something to trigger the next steps after your approval? Does it need a second approval?

oalbrigt:
Thanks.

oalbrigt:
> @oalbrigt - I do not know how the project is processing the PRs. Do I need to do something to trigger the next steps after your approval? Does it need a second approval?

No. I just tend to keep tickets open for a couple of days to let others comment if they have any suggestions or questions.

fmherschel:
@oalbrigt Thanks for the explanation.

fdanapfel:
@fmherschel @oalbrigt As far as I remember a long time ago we agreed that patches to the SAPInstance and SAPDatabase resource agents should be reviewed and approved by both the "SAP engineering" teams from Red Hat and SUSE before they are commited, to make sure they don't contain anything that might cause issues for customers using those agents in production (for example if a patch adds a distribution specific feature which would cause the RA to no longer work on the other distribution).

This is less critical if a patch is submitted by a member of the "SAP engineering" team from either company, but if somebody from outside tries to submit a patch we should make sure it gets reviewed and approved before it is commited.

The patch looks good to me, it matches the suggestions made in https://unix.stackexchange.com/questions/539147/systemctl-check-if-a-unit-service-or-target-exists on how to avoid issues with the "list-unit-files" command of systemctl.

Issue #2212, opened 2026-08-27
WIP: About monitor_10 for the Filesystem RA.


HideoYamauchi:
Hi Oyvind,
Hi All,

The `monitor_10` action of the Filesystem RA skips execution when a UUID or LABEL is specified instead of a block device path.
However, even when a UUID or LABEL is used, shouldn't it be possible to perform the `monitor_10` check (which uses the `dd` command) if the underlying block device can be identified via `blkid`?

```
Filesystem_monitor_10()
{
        RDEVICE=$DEVICE
        if [ "$blockdevice" = "no" ] ; then
                RDEVICE=`blkid | grep ${DEVICE} | cut -d ':' -f1`
                ocf_log info "### monitor_#### DEVICE=${DEVICE}"
                if [ ! -z "$RDEVICE" ]; then
                        ocf_log info "### monitor_#### RDEVICE=${RDEVICE}"
                else
                        ocf_log warn "$RDEVICE is not a block device, monitor 10 is noop"
                        return $OCF_SUCCESS
                fi
        fi
        dd_opts="iflag=direct bs=4k count=1"
        err_output=$(dd if="$RDEVICE" $dd_opts 2>&1 >/dev/null)
        if [ $? -ne 0 ]; then
                ocf_exit_reason "Failed to read device $RDEVICE"
                ocf_log err "dd said: $err_output"
                return $OCF_ERR_GENERIC
        fi
        return $OCF_SUCCESS
}
```

That said, one thing I haven't verified is the point mentioned regarding `monitor_20`: that direct access is not possible for cifs/smbfs.
```
Filesystem_monitor_20(){
        if [ "$blockdevice" = "no" ] ; then
                # O_DIRECT not supported on cifs/smbfs 
                dd_opts="oflag=sync bs=4k conv=fsync,sync"
        else
                # Writing to the device in O_DIRECT mode is imperative
                # to bypass caches.
                dd_opts="oflag=direct,sync bs=4k conv=fsync,sync"
        fi
(snip)
```

If direct access is indeed impossible for cifs/smbfs, then I assume the `monitor_10` check cannot be performed in those cases either, even when a UUID or LABEL is specified.

What do you all think? I would appreciate your feedback.

Best Reards,
Hideo Yamauchi

oalbrigt:
Sounds like we just didnt think of those cases when we implemented it, so it would make sense to add support for it.

[...]

PR #2216, opened 2026-09-02
Mid: Filesystem: For block devices, allow monitor_10 even when specified by UID or LABEL.

HideoYamauchi:
Hi Oyvind,
Hi All,

This modification is based on the discussion in the following issue.
 - https://github.com/ClusterLabs/resource-agents/issues/2212
 
It allows the `monitor_10` processing to proceed when a match with `blkid` is found, even if UUID or LABEL is specified.

Best Reards,
Hideo Yamauchi.

### DIFF
--- a/heartbeat/Filesystem
+++ b/heartbeat/Filesystem
@@ -1063,12 +1063,25 @@ Filesystem_status()
 #
 Filesystem_monitor_10()
 {
+	RDEVICE=$DEVICE
+	monitor_10_err_config="no"
 	if [ "$blockdevice" = "no" ] ; then
-		ocf_log warn "$DEVICE is not a block device, monitor 10 is noop"
-		return $OCF_SUCCESS
+		if [ ! -z "$use_check_blkid_opt" ] ; then
+			RDEVICE=`blkid ${use_check_blkid_opt} ${DEVICE}`
+			if [ -z "$RDEVICE" ]; then
+				monitor_10_err_config="yes"
+			fi
+		else
+			monitor_10_err_config="yes"
+		fi
+
+		if [ "$monitor_10_err_config" = "yes" ] ; then
+			ocf_exit_reason "$DEVICE is not a block device, monitor 10 cannot be executed. Please check OCF_RESKEY_device parameter and OCF_CHECK_LEVEL attribute for the monitor operation."
+			return $OCF_ERR_CONFIGURED
+		fi
 	fi
 	dd_opts="iflag=direct bs=4k count=1"
-	err_output=$(dd if="$DEVICE" $dd_opts 2>&1 >/dev/null)
+	err_output=$(dd if="$RDEVICE" $dd_opts 2>&1 >/dev/null)
 	if [ $? -ne 0 ]; then
 		ocf_exit_reason "Failed to read device $DEVICE"
 		ocf_log err "dd said: $err_output"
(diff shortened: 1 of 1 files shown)


heartbeat/Filesystem
```diff
@@ -1063,12 +1063,21 @@ Filesystem_status()
...
 Filesystem_monitor_10()
 {
+	RDEVICE=$DEVICE
 	if [ "$blockdevice" = "no" ] ; then
-		ocf_log warn "$DEVICE is not a block device, monitor 10 is noop"
-		return $OCF_SUCCESS
+		if [ ! -z "$use_check_blkid_opt" ] ; then
+			RDEVICE=`blkid ${use_check_blkid_opt} ${DEVICE}`
+			if [ -z "$RDEVICE" ]; then
+				ocf_log warn "$DEVICE is not a block device, monitor 10 is noop"
```
oalbrigt: Let's change these to `ocf_exit_reason()` and `return $OCF_ERR_CONFIGURED` instead, to let users know that their setup is not working the way they expected.

HideoYamauchi: @oalbrigt 
I’ve made the change.

I left the existing logs as they were for the sake of compatibility.
I switched to the slightly longer `ocf_exit_reason`.
It gets truncated in the `crm_mon` display, but the full text appears in the logs.

What do you think?

Best Regards,
Hideo Yamauchi.

heartbeat/Filesystem
```diff
@@ -1063,12 +1063,21 @@ Filesystem_status()
...
-		ocf_log warn "$DEVICE is not a block device, monitor 10 is noop"
-		return $OCF_SUCCESS
+		if [ ! -z "$use_check_blkid_opt" ] ; then
+			RDEVICE=`blkid ${use_check_blkid_opt} ${DEVICE}`
+			if [ -z "$RDEVICE" ]; then
+				ocf_exit_reason "$DEVICE is not a block device, monitor 10 cannot be executed. Please check OCF_RESKEY_device parameter and OCF_CHECK_LEVEL attribute for the monitor operation."
+				return $OCF_ERR_CONFIGURED
+			fi
+		else
+			ocf_log warn "$DEVICE is not a block device, monitor 10 is noop"
```
oalbrigt: We also have to do the exit reason/CONFIGURED return here.

HideoYamauchi: @oalbrigt 

I’ve also modified the legacy processing.

If there are no issues, I’ll clean up the PR and resubmit it.
What do you think?

Many thanks,
Hideo Yamauchi

HideoYamauchi: @oalbrigt 
It looks like there is an error occurring in Jenkins.
The fix itself doesn't seem to be the problem, though...

heartbeat/Filesystem
oalbrigt: That is good. Just giving them some info instead of simply "Failed with error code" is the important thing, and also makes it easier for them to search their logs.

heartbeat/Filesystem
oalbrigt: You dont have to resubmit it.

You can run `git rebase -i HEAD~4` and change `pick` to `s` on the lines you want to merge into the first commit. Then you'll get mix and match any parts of the commit messages for the commit.

And finally you run `git push -f` and we should get the new commit instead of the 4 we have now.

oalbrigt: Yeah. Jenkins is or was under maintenance today, so dont mind that failing when it's not related to the PR.

HideoYamauchi: @oalbrigt 
I tried rebasing. 

How does it look?

HideoYamauchi:
@oalbrigt 

It looks like there a points that need correction.
I will check it again.

----
I added a check because duplicate values ​​can be set for the device LABEL.
However, I believe this check should be performed at an earlier stage.
--> For example, checking right before the mount operation at startup.

What do you think?


Best Regards,
Hideo Yamauchi.

PR #2217, opened 2026-09-08
Filesystem: fix search for bind mounts to avoid substring matching

sim-:
When looking for bind mounts, we need to be more careful with matching of $mount_disk in the list_mounts output to avoid cases such as "/dev/drbd1" finding a match with "/dev/drbd10", resulting in attempts to unmount unrelated mountpoints.

Fixes: 7b0ad80a04ee2a6bd30e4e7ca0552f96e979d5e5


sim-:
Hrm, this is wrong, since -w is present... Digging further...

Issue #2218, opened 2026-09-11
aws-vpc-move-ip fragile handling of credentials output from AssumeRole call


thimslugga:
The `aws-vpc-move-ip` resource agent looks to have a bug due to the fragile handling output when making a `aws sts assume-role` call.

- https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/aws-vpc-move-ip#L392-L409
- https://github.com/ClusterLabs/resource-agents/blob/main/heartbeat/aws-vpc-move-ip#L238-L255

The way this is being done in the resource agent script, is extremely fragile:

```bash
...

execute_cmd_as_role(){
    cmd=$1
    role=$2
    output="$($AWSCLI_CMD sts assume-role --role-arn $role --role-session-name AWSCLI-RouteTableUpdate --output=text)"
    export AWS_ACCESS_KEY_ID="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $5}')"
    export AWS_SECRET_ACCESS_KEY="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $7}')"
    export AWS_SESSION_TOKEN="$(echo $output | awk -F" " '$4=="CREDENTIALS" {print $8}')"

    #Execute command
    ocf_log debug "Assumed Role ${role}"
    ocf_log debug "$($OCF_RESKEY_awscli sts get-caller-identity)"
    ocf_log debug "executing command: $cmd"
    response="$($cmd)"
    unset output AWS_ACCESS_KEY_ID AWS_SECRET_ACCESS_KEY AWS_SESSION_TOKEN
    echo $response
}

...
```

when ideally it would do it via `--query`:

```bash
...

output="$($AWSCLI_CMD sts assume-role \
  --role-arn "$role" \
  --role-session-name AWSCLI-RouteTableUpdate \
  --query 'Credentials.[AccessKeyId,SecretAccessKey,SessionToken]' \
  --output text)"
rc=$?

...
```

```
diff --git a/heartbeat/aws-vpc-move-ip b/heartbeat/aws-vpc-move-ip
index 1348c59..233668c 100755
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -238,18 +238,39 @@ END
 execute_cmd_as_role(){
    cmd=$1
    role=$2
[...]

PR #2219, opened 2026-09-11
aws-vpc-move-ip: fix fragile handling of response from AssumeRole

thimslugga:
Fix to address https://github.com/ClusterLabs/resource-agents/issues/2218

### DIFF
--- a/heartbeat/aws-vpc-move-ip
+++ b/heartbeat/aws-vpc-move-ip
@@ -1,5 +1,6 @@
 #!/bin/sh
-#
+# shellcheck shell=sh
+# vim: set sw=4 sts=4 et tw=80 :
 #
 # OCF resource agent to move an IP address within a VPC in the AWS
 #
@@ -25,11 +26,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write the Free Software Foundation,
 # Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
-#
-
 
 #######################################################################
-# Initialization:
+# Initialization
 
 : ${OCF_FUNCTIONS_DIR=${OCF_ROOT}/lib/heartbeat}
 . ${OCF_FUNCTIONS_DIR}/ocf-shellfuncs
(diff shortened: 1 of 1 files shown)


heartbeat/aws-vpc-move-ip
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
-#
+# shellcheck shell=sh
```
arslan9292: Is there a specific reason of this directive here? `ShellCheck` should pick up the POSIX `sh` rules automatically from the `#!/bin/sh` shebang, so we could probably drop this line to keep things clean.

heartbeat/aws-vpc-move-ip
```diff
@@ -1,5 +1,6 @@
...
+# vim: set sw=4 sts=4 et tw=80 :
```
arslan9292: Let's move this line to the end of the file to keep the header clean.

heartbeat/aws-vpc-move-ip
```diff
@@ -25,11 +26,9 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write the Free Software Foundation,
 # Inc., 59 Temple Place - Suite 330, Boston MA 02111-1307, USA.
-#
-
 
 #######################################################################
-# Initialization:
+# Initialization
```
arslan9292: Minor cosmetic thing: let's also keep the colon here so it matches the rest of `resource-agents` script files (best to stay consistent).

arslan9292 (commented):
Please review and update the highlighted sections.

heartbeat/aws-vpc-move-ip
thimslugga: addressed in commit

heartbeat/aws-vpc-move-ip
thimslugga: addressed in commit

heartbeat/aws-vpc-move-ip
thimslugga: addressed in commit

thimslugga:
This is related to RH case 04537910 and SUSE case 02081228

- https://access.redhat.com/solutions/7148183
- https://redhat.atlassian.net/browse/RHEL-259558

olvnerys:
Can this be actioned please. There is no clear workaround and this issue will prevent failover for customers that use assumed role and a shared VPC. It is important to have the fix available.

